Issue #7, opened 2024-07-14
[DOCS] Make code one click copyable in readme


darigovresearch:
Thought it may be helpful to have any commands one click copyable in the readme

domdomegg:
Closed by #8

PR #8, opened 2024-07-14
docs: Updates Developer setup instructions to be one click copyable

darigovresearch:
Relevant to https://github.com/bluedotimpact/bluedot/issues/7


domdomegg (approved):
LGTM. I think it might be worth simplifying this further by just having one script that installs everything for a certain platform (ideally as a script in the repo, that can just be run including in CI... one day).

PR #16, opened 2024-09-26
Add airtable-cache prototype

domdomegg:


apps/airtable-cache/README.md
```diff
@@ -0,0 +1,35 @@
+# airtable-cache
+
+A caching proxy to Airtable. Consumers can generally ignore the details, and just pretend this is Airtable itself for standard [Airtable API](https://airtable.com/developers/web/api/) routes. For example, with [`airtable-ts`](https://github.com/domdomegg/airtable-ts):
```
domdomegg, line 3: Need to explain why this exists, i.e.:

Airtable has low rate limits
Airtable is slow (sometimes)

apps/airtable-cache/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+No special actions needed, just follow [the general developer setup instructions](../../README.md#developer-setup-instructions)
+
+## Deployment
+
+This app is deployed onto the K8s cluster as a docker container.
+
+To deploy a new version, simply commit to the master branch. GitHub Actions automatically handles CD.
+
+## TODO
```
domdomegg, line 33: Delete this TODO - has now been implemented

apps/airtable-cache/src/app.ts
```diff
@@ -0,0 +1,226 @@
...
+  await instance.register(async (i) => {
+    i.get('/', async (req, reply) => {
+      reply.redirect('https://github.com/bluedotimpact/bluedot/tree/master/apps/airtable-cache#readme');
+    });
+
+    i.get('/v0/meta/bases/:baseId/tables', cachedGetHandler);
+    i.get('/v0/:baseId/:tableId', cachedGetHandler);
+    i.get('/v0/:baseId/:tableId/:recordId', cachedGetHandler);
+
+    // TODO: clear out base schema
```
domdomegg, line 25: These TODOs are done

apps/airtable-cache/src/lib/cache.ts
```diff
@@ -0,0 +1,74 @@
+import z from 'zod';
+import { sql } from 'kysely';
+import { db } from '../db/client';
```
domdomegg, line 3: I reckon we should probably scrap the entire database, and just use an in-memory cache.

Postgres uses up a bunch of memory in practice (~200MB) and we probably don't need all its functionality.

Maybe we want to keep Postgres if we want to set up webhooks, so we don't set up webhooks then forget about them.

apps/airtable-cache/src/app.ts
```diff
@@ -0,0 +1,226 @@
+import { fastify, FastifyReply, FastifyRequest } from 'fastify';
```
domdomegg, line 1: File is a little messy and probably too long: consider splitting into multiple files

apps/frontend-example/src/lib/api/db/index.ts
```diff
@@ -3,4 +3,5 @@ import env from '../env';
 
 export default new AirtableTs({
   apiKey: env.AIRTABLE_PERSONAL_ACCESS_TOKEN,
+  endpointUrl: 'http://localhost:8001',
```
domdomegg, line 6: This should be dynamic, and also needs to work in prod

apps/airtable-cache/src/app.ts
```diff
@@ -0,0 +1,226 @@
...
+        const airtableResponse = await proxyToAirtable(req);
+        return airtableToFastifyReply(airtableResponse, reply);
+      },
+    });
+  });
+
+  return instance;
+};
+
+const cachedGetHandler = async (req: FastifyRequest, reply: FastifyReply) => {
```
domdomegg, line 57: Maybe consider setting up webhooks on Airtable so that if the records change through other routes e.g. different API changes, or manual edits in Airtable, the cache isn't stale.

Or at least make it very easy to clear the cache - maybe a web UI or something.

apps/airtable-cache/src/app.ts
```diff
@@ -0,0 +1,226 @@
...
+  CacheValue, clearValues, clearValuesContainingBody, getValue, setValue,
+} from './lib/cache';
+
+export const getInstance = async () => {
+  const instance = fastify();
+
+  await instance.register(errorHandlerPlugin);
+  await instance.register(fastifyCors);
+
+  await instance.register(async (i) => {
```
domdomegg, line 16: At the moment, this is a free proxy for any requests to Airtable, and will cache anybody's requests there.

We should instead validate the token is from some allowlisted set.

PR #27, opened 2024-11-10
meet: don't let user enter meet if session doesn't start for 1 hour

### LINKED ISSUE #20: meet: Warn users when they're joining at the wrong time
## Context

We run educational courses on AI safety and biosecurity, and in any week have hundreds of cohort sessions (a cohort is a group of 5-8 people following our course, and a cohort session is a single meeting of these people i.e. a calendar event with an associated meeting - usually once a week).

To join a cohort session people open the URL added to the calendar event, which looks like `https://meet.bluedot.org/?cohortId=recrTJlMSGX14wkkB`.

We use the `cohortId` (which is the underlying primary key for our cohorts table in Airtable, which we use as a database) to fetch the cohort, and its corresponding cohort sessions. Of those cohort sessions, we find the one starting closest to the current time and show its details. The user can then register their attendance by clicking their name, and get redirected to the appropriate Zoom room.

(In some places you'll see 'cohort class' mentioned. This is the same as a cohort session, but we haven’t got around to renaming everywhere we’ve used this terminology yet.)

## Problem

Occasionally, people on our courses get confused as to when they’re supposed to join their session, and end up trying to join at a different time (this happens especially often when there are changes in daylight savings etc.).

If each cohort had its own Zoom room they’d just be in an empty Zoom room by themselves. However, because we reuse Zoom rooms this can result in them crashing into an existing session - causing disruption to that session, and greater confusion for the participant. (We reuse Zoom rooms because we have to pay for a licence for each room we create - this is the difference between £140/month and £3000/month. We also can’t issue new licences via the API, so we’d manually have to be managing hundreds of Zoom accounts.)

As well as a poor experience for participants, we often get people then emailing or Slacking us confused about their session time, which takes our time away from making and delivering great courses.

## Proposed solution

You could change our “meet” application to block or warn people trying to join their cohort session at the wrong time. For example, a warning might look like showing a dismissible alert banner if it’s not an appropriate time to try joining their session.

There are multiple ways to approach almost any problem. You can implement a different solution if you think it would be better - and if you consider another solution but reject it, you can still include this in your accompanying documentation. (But you are not expected to come up with a different solution! We have not been able to think of a much better ‘obvious’ solution ourselves)

## Get started

To get started:

- Change directory to the subfolder `apps/meet`
- Set the in `AIRTABLE_PERSONAL_ACCESS_TOKEN` in `.env.local`
- Run `npm start`, and navigate to http://localhost:8000/?cohortId=recABCD, replacing `recABCD` with a cohort id.

If all is well, this should show you something like this:

Untitled

To implement the proposed solution you’ll probably want to start by looking in:
- `src/components/SelectPersonView.tsx`
- `src/pages/api/public/meeting-participants.ts`

osc245:
fixes #20 

Demo:
https://github.com/user-attachments/assets/932acd09-f4c2-4a94-af90-009d0c4164ac

Please let me know if there's anything you want me to change


osc245:
I just had a thought: these changes only check if someone showed up >1 hour early, but do you also have issues with people joining sessions that have already been? Perhaps it'd be better we compared the absolute difference between current and start time

domdomegg (approved):

PR #32, opened 2025-01-17
Fix lint

domdomegg:

PR #34, opened 2025-01-17
Add new website-25 project config for CI/CD

tarinrickett:
Relies on https://github.com/bluedotimpact/bluedot/pull/32


domdomegg (approved):
LGTM!

You might not need the Airtable token/alerts slack channels, but probably fine to include for now.

Issue #60, opened 2025-01-21
Testing broken for /pages/ dir


tarinrickett:

tarinrickett:
See https://github.com/bluedotimpact/bluedot/pull/57

domdomegg:
I think you'll want to exclude the test files from the Next.js build - I'm not sure how to do that off the top of my head but I'd guess there's some way?

Maybe editing the shared BlueDot next config to exclude files with a `.test.tsx?` suffix or similar.

janraasch:
Hi @tarinrickett & @domdomegg 👋,

I had a look into Next.js docs. Here's their [`with-vitest`-example](https://github.com/vercel/next.js/tree/canary/examples/with-vitest) repo.

Their recommended solution seems to be to have `pages`-tests in `__tests__` directory.

On https://github.com/bluedotimpact/bluedot/pull/74 I checked that this works for `careers` & `about`.

If we were to migrate these pages to the `app`-router, we could keep the colocation of test & source files.

**conclusion (short term)**

To get coverage for a complete page (e.g. because components have not been moved into their separate files or there is some complex `page`-specific logic going on.) *quickly* their solution of placing those tests inside `__tests__` isn't so bad.

What do you think?

**further discussion (not so short term)**

Still, colocating these tests next to their source files would be nice. Is there a particular reason why did not go with the new `app`-router for this project? Full disclosure: I have not used the new `app`-router myself (yet?) 😇. 

**💡 idea** If it makes sense, I'm happy to try to migrate to the `app`-router. Then we can see how that goes. However that sounds like a completely new task with a possibly different priority as we do not want to slow down `MVP`-development.

**💁‍♀️ why did I pick this task?** It seemed "small" enough & it had a high priority assigned to it / top of the list.

PS: @tarinrickett: I can't seem to be able to assign this task to me and change its status. Can you help? 🙏

janraasch:
Fixed with https://github.com/bluedotimpact/bluedot/pull/74.

I'll create a separate card to look into the `app`-router.

PR #57, opened 2025-01-20
Add test coverage for About page

tarinrickett:
# Summary

Add test coverage for About page

## Testing
```
$ npm run test
 RUN  v1.6.0 /Users/tarin-bluedot/bluedot/apps/website-25

 ✓ src/components/ErrorPage.test.tsx (2)
 ✓ src/pages/about.test.tsx (1)

 Test Files  2 passed (2)
      Tests  3 passed (3)
   Start at  17:22:26
   Duration  412ms (transform 46ms, setup 0ms, collect 419ms, tests 16ms, environment 119ms, prepare 52ms)
```

PR #74, opened 2025-01-22
Add test coverage for `about`- & `careers`-page

janraasch:
# Summary

Working version of #57 😇☺️.

# Issue 

See #60.

### What
Placed the two `pages` tests into a `__tests__` directory.

### Why
Looking at the [`with-vitest`-example](https://github.com/vercel/next.js/tree/canary/examples/with-vitest) referenced by NextJS docs this seems to be the way to go.

### Other options

For colocating tests next to their source files (even for `pages`) we could migrate to the [`app`-router](https://nextjs.org/docs/app), see #60 for more information.

## Testing
```
$ npm run test:update
Cached:    0 cached, 7 total
  Time:    3.51s 
```

### DIFF
--- a/apps/website-25/src/__tests__/pages/__snapshots__/about.test.tsx.snap
+++ b/apps/website-25/src/__tests__/pages/__snapshots__/about.test.tsx.snap
@@ -1,0 +1,368 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`AboutPage > should render the error message correctly 1`] = `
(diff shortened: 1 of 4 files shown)


tarinrickett (approved):
thank you for the investigation, Jan! i think you'll need a rebase and snapshot update but otherwise LGTM.

i prefer colocating tests whenever possible, but im comfortable keeping full page tests in this directory for now. we can circle back on the app router question and see if it is worth the additional effort.

PR #82, opened 2025-01-22
Create Card and CTALinkOrButton components

### LINKED ISSUE #77: Create Card component
There are several instances of highly similar but not exact cards. let's create a generic Card component in libraries/ui. Let's use the Team cards as a source of truth for now (e.g., header and padding).

@tarinrickett will follow up with the UX designer to confirm what the intent is between these multiple stylings

Image

### Other cards – we will align these to use the generic card styles from above:
Image
Image

orhanna14:
# Summary

Create `Card` component and `CTALinkOrButton` component. Deprecate legacy `Card` component

## Issue

Fixes #77

## Description

Create Card and CTALinkandButton components & tests

## Developer checklist
- [x] Used [BEM naming convention](https://getbem.com/naming/)
- [X] Added new unit tests where applicable

## Screenshot

| 📸 |  |
|---------|---|
| 🖥️ | 
Screenshot 2025-01-22 at 3 16 43 PM
Screenshot 2025-01-23 at 2 12 40 PM
Screenshot 2025-01-23 at 2 12 28 PM
Screenshot 2025-01-23 at 2 12 18 PM
Screenshot 2025-01-23 at 2 12 07 PM

 |
| 📱  |  |

## Testing
```
$ npm run test:update
@bluedot/website-25:test:  Test Files  10 passed (10)
@bluedot/website-25:test:       Tests  11 passed (11)
@bluedot/website-25:test:    Start at  14:10:12
@bluedot/website-25:test:    Duration  2.29s (transform 801ms, setup 1ms, collect 8.46s, tests 135ms, environment 5.13s, prepare 925ms)
```


orhanna14:
The example image above is from a `<TeamSection />` I was working on - let me know if you'd like me to continue here, @tarinrickett !

tarinrickett:
> The example image above is from a `<TeamSection />` I was working on - let me know if you'd like me to continue here, @tarinrickett !

Awesome, @orhanna14 ! Yes, would you like to replace the current temporary set-up in the existing [TeamSection](https://github.com/bluedotimpact/bluedot/blob/b9b5ac01a7a67bde4f9af838fe17fdf82de95ed6/apps/website-25/src/components/about/TeamSection.tsx#L3) website component with your new cards? We can do this in a follow-up PR to keep this one focused.

libraries/ui/src/CardGeneric.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { CTALink } from './CTALink';
+
+interface CardProps {
+  imageSrc: string;
+  name: string;
+  role: string;
+  linkedInUrl: string;
+}
+
+export const CardGeneric: React.FC<CardProps> = ({
```
tarinrickett, line 11: nit: let's rename this simply to `Card`

libraries/ui/src/CardGeneric.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { CTALink } from './CTALink';
+
+interface CardProps {
+  imageSrc: string;
+  name: string;
+  role: string;
+  linkedInUrl: string;
```
tarinrickett, line 8: let's name these props to be more generic. i'd recommend `title`, `subtitle`, `ctaUrl`

libraries/ui/src/CardGeneric.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <p className="card__role text-base text-black">{role}</p>
+        </div>
+
+        <CTALink
+          href={linkedInUrl}
+          variant="secondary"
+          withChevron={false}
+        >
+          LinkedIn
+        </CTALink>
```
tarinrickett, line 39: CTA should be optional

libraries/ui/src/CardGeneric.tsx
```diff
@@ -0,0 +1,45 @@
...
+          className="card__image w-full h-full object-cover rounded-2xl"
+          src={imageSrc}
+          alt={`${name}`}
+        />
+      </div>
+
+      <div className="card__content flex flex-col gap-8">
+        <div className="card__text space-y-1">
+          <h2 className="card__name text-2xl font-semibold text-bluedot-darker">{name}</h2>
+          <p className="card__role text-base text-black">{role}</p>
```
tarinrickett, line 30: i think `subtitle` can be optional. let's make just image and title required for now

libraries/ui/src/CTALink.tsx
```diff
@@ -0,0 +1,47 @@
+import clsx from 'clsx';
+import React from 'react';
+
+// Create a styled link component using CTAButton styling
+export const CTALink: React.FC<{
```
tarinrickett, line 5: this is great, thanks for adding this as well!!

now that i'm looking at it, can we update `CTAButton` to include handling for links _or_ buttons? we can rename it to a more generically named `CTA` component. the styles are identical, so colocating should make it easier for us to maintain moving forward.

libraries/ui/src/CTALink.tsx
```diff
@@ -0,0 +1,47 @@
...
+  href,
+  className,
+  variant = 'primary',
+  withChevron = false,
+  children,
+}) => {
+  return (
+    <a
+      href={href}
+      target="_blank"
```
tarinrickett, line 21: this makes sense for the current use case of LinkedIn urls, but when we move it to CTA please make this optional so we can support onsite/offsite link behavior

tarinrickett (commented):
thanks Olivia! added some notes but overall looking really clean. great call adding support for CtaLink now too!

libraries/ui/src/CardGeneric.tsx
orhanna14: 💯 agree. I was hesitant with the legacy `Card` component, but see that that isn't being used anywhere so I will deprecate and update this

libraries/ui/src/CTALink.tsx
orhanna14: Updated for a new `CTALinkOrButton` component - I notice an existing `LinkOrButton` component that may be able to be consolidated here in a future iteration

orhanna14:
> > The example image above is from a `<TeamSection />` I was working on - let me know if you'd like me to continue here, @tarinrickett !
> 
> Awesome, @orhanna14 ! Yes, would you like to replace the current temporary set-up in the existing [TeamSection](https://github.com/bluedotimpact/bluedot/blob/b9b5ac01a7a67bde4f9af838fe17fdf82de95ed6/apps/website-25/src/components/about/TeamSection.tsx#L3) website component with your new cards? We can do this in a follow-up PR to keep this one focused.

Yep! Sounds good

libraries/ui/src/index.ts
```diff
@@ -23,8 +23,7 @@ export type { BoxProps } from './legacy/Box';
 export { Button } from './legacy/Button';
 export type { ButtonProps } from './legacy/Button';
 
-export { Card, CardButton } from './legacy/Card';
-export type { CardProps } from './legacy/Card';
+export { CardButton } from './legacy/CardButton';
```
tarinrickett, line 26: ah, is there a clash on Card naming exports between legacy and updated?

let's keep the legacy export the same – that may be used elsewhere. for now could we export as DefaultCard or GenericCard, as you had suggested? sorry for introducing this clash inadvertently !

orhanna14: I didn't see the legacy being `Card` used anywhere else, so this should be safe FWIW

libraries/ui/src/CTALinkOrButton.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    expect(chevron).toBeTruthy();
+    expect(chevron.getAttribute('src')).toBe('/icons/chevron_white.svg');
+  });
+
+  test('renders as a link when href is provided', () => {
+    render(<CTALinkOrButton url="https://example.com" variant="primary">Click me</CTALinkOrButton>);
+    const link = screen.getByTestId('cta-link');
+    expect(link).toBeTruthy();
+    expect(link.tagName).toBe('A');
+    expect(link.getAttribute('href')).toBe('https://example.com');
```
tarinrickett, line 28: 💯

libraries/ui/src/CTALinkOrButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  className,
+  variant = 'primary',
+  withChevron = false,
+  children,
+  url,
+  ...rest
+}) => {
+  const commonClassNames = clsx(
+    CTA_BASE_STYLES,
+    CTA_VARIANT_STYLES[variant],
```
tarinrickett, line 29: loveee great idea for making this really clean!

libraries/ui/src/index.ts
tarinrickett: awesome thanks for checking! LGTM then.

this is making me realize we don't have a process to document backwards incompatible changes right now. do you think it is worth it for major shared library like `libraries/ui` to maintain an internal changelog? +@janraasch for vis too!

tarinrickett (approved):

libraries/ui/src/index.ts
janraasch: Hi @tarinrickett, adding a changelog might come in handy in the future. 

We'll need this for sure, if we'd like integrating apps to be able to update at their own pace.

Status quo: At the moment every `app` uses `"@bluedot/ui": "*"` so any change to the `ui`-library has **immediate** impact on every integrating app.

Issue #76, opened 2025-01-22
Update TeamMember to user Card component


tarinrickett:
Requires #77

PR #98, opened 2025-01-24
Update TeamSection to use Card component

orhanna14:
# Summary

Update TeamSection to use Card component

## Issue

Context #76 & #77 

## Description

Updates the existing TeamSection to use the Card component

## Developer checklist
- [X] Used [BEM naming convention](https://getbem.com/naming/)
- [ ] Added new unit tests where applicable

## Screenshot

| 📸 |  |
|---------|---|
| 🖥️ | 
Screenshot 2025-01-24 at 12 33 51 PM
 |
| 📱  |  |

## Testing
```
$ npm run test:update
Test Files  10 passed (10)
@bluedot/website-25:test:       Tests  11 passed (11)
@bluedot/website-25:test:    Start at  12:47:29
@bluedot/website-25:test:    Duration  1.72s (transform 328ms, setup 0ms, collect 6.18s, tests 177ms, environment 1.59s, prepare 654ms)

```


apps/website-25/src/components/about/TeamSection.tsx
```diff
@@ -1,41 +1,68 @@
...
+      <div className="team__grid flex flex-row flex-wrap mt-16 gap-4 max-w-[1400px] mx-auto">
+        {teamMembers.map((member) => (
+          <Card
+            key={member.name}
+            imageSrc={member.imageSrc}
+            title={member.name}
+            subtitle={member.role}
+            ctaUrl={member.linkedInUrl}
+            className="team__card w-[323px]"
+            imageClassName="team__card-image h-[223px] object-cover"
```
tarinrickett, line 60: `object-cover` is already included in OOTB styling, so we dont need to pas it here

`h-full` is also set on the image, which will be the same specificity level as `h-[223px]`. this could introduce conflicts depending on how TailwindCSS builds out the class orders. can we set these heights (`h-[223px]`) and widths (`w-[323px]` above) directly in `Card`? i think it would be good to standardize the sizes we want across the ecosystem regardless.

Issue #112, opened 2025-01-28
Add website-25 to storybook


janraasch:
See https://github.com/bluedotimpact/bluedot/pull/108#discussion_r1932072113.

This is a follow-up to #59.

**issue**

When adding `stories` for `website-25` components to `storybook` the ui broke

<img width="1562" alt="Image" src="https://github.com/user-attachments/assets/802c86b7-5bb1-4e68-8d55-18a7f5bef949" />

**💡 possible solution(s)**

* Add `import React from 'react';` to `website-25` components. This seems to work, but then `eslint` fails with `'React' is defined but never used.eslint[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)`.
* Try using the ~~`storybook-nextjs`-framework~~ `@storybook/experimental-nextjs-vite`-framework, see https://storybook.js.org/docs/get-started/frameworks/nextjs?renderer=react#with-vite

PR #121, opened 2025-01-29
Add `website-25` to `storybook`

janraasch:
# Summary

Add `website-25` to `storybook`, see `IntroSection.stories.tsx`.

## Issue

#112

## Description

✅ Extending `viteConfig` with `plugins: [react(), ...(config?.plugins ?? [])]` takes care of having to import `React` in the respective component.

🚧 Having to import `globals.css` on the `story`-level 
seems like a little bit of a crutch - same for the `wrapper`. Still, this sets us up for using `storybook` on `website-25`, so might be 🆗 for now. **What do you think, @tarinrickett?**

## Screenshot

Screenshot 2025-01-29 at 11 34 07

## Testing
```
$ npm run test:update
Cached:    6 cached, 7 total
  Time:    2.45s 
```


apps/website-25/src/components/about/IntroSection.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import type { Meta, StoryObj, StoryFn } from '@storybook/react';
+
+import IntroSection from './IntroSection';
+
+// Include globals
+import '../../globals.css';
```
tarinrickett, line 6: are we able to import this from storybook preview wrapper?

https://storybook.js.org/docs/configure/styling-and-css

apps/website-25/src/components/about/IntroSection.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+// Wrap component
+const withWebsiteWrapper = (Story: StoryFn) => (
+  <main className="bluedot-base">
+    <div><Story /></div>
+  </main>
+);
```
tarinrickett, line 13: can we set this up as a custom decorator for all Stories from storybook/preview ?

https://storybook.js.org/docs/writing-stories/decorators

apps/website-25/src/components/about/IntroSection.stories.tsx
janraasch: Yes we could. 

However, then we have this on `libraries/ui`-stories as well.

apps/website-25/src/components/about/IntroSection.stories.tsx
janraasch: Yes we could.

However, then we'd wrap this around `libraries/ui-stories` as well.

janraasch:
Hi @tarinrickett, I moved the `css`-import & the `bluedot-base`-wrapper to a `lib`-helper. Then it's all nicely wrapped up in one place, while `libraries/ui`-stories remain untouched by `website base.css`. Let me know what you think.

janraasch:
Never mind!!! 😇. Hi @tarinrickett, sorry, I was under the impression that those `import globals` for isolated to the `stories.tsx`-file... Since that's not the case we might as well import this on `preview` as you suggested.

tarinrickett (approved):

tarinrickett:
> Never mind!!! 😇. Hi @tarinrickett, sorry, I was under the impression that those `import globals` were _isolated_ to the respective `stories.tsx`-file... Since that's not the case we might as well import this on `preview` as you suggested.

thanks Jan! yeah, i think we should be safe to set this up for both. I would like to move website-25's CSS to libraries/ui once we have a stable design system and bandwidth to audit the impact on the other apps 💯

janraasch:
Awesome!

>  I would like to move website-25's CSS to libraries/ui once we have a stable design system and bandwidth to audit the impact on the other apps 💯

That's a great idea 👍.

PR #139, opened 2025-01-30
Enable `tailwindcss/no-unnecessary-arbitrary-value` `eslint` rule

### LINKED ISSUE #101: Linting for TailwindCSS arbitrary values that exist in the design system
E.g. `px-[12px]` vs. `px-4`

This was something that came up on my last team as well, but we never prioritized the investigation. I think it would be a huge lift in baking in checks for this distributed team.

I'd be surprised if there isn't a VSCode extension or some lint rule we can borrow. Need to do some digging into what's out there 👍

janraasch:
# Summary

[Avoid unjustified arbitrary classnames](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/no-unnecessary-arbitrary-value.md).

## Issue

Fixes #101.

## Description

* Add `eslint-plugin-tailwindcss`.
* Add `lint:fix`-task.
* Enable `no-unnecessary-arbitrary-value`-rule.

While at it, I also enabled

* [no-contradicting-classname](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/no-contradicting-classname.md) and
* [enforces-shorthand](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/enforces-shorthand.md)

I think those are useful as well, but I am open to discussion here.

ℹ️ There are more interesting rules (e.g. [classnames-order](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/classnames-order.md), [enforces-negative-arbitrary-values](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/enforces-negative-arbitrary-values.md)), but I wanted to keep the changeset *smallish* :)

Check out [more supported rules here](https://github.com/francoismassart/eslint-plugin-tailwindcss?tab=readme-ov-file#supported-rules).

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📕 |  |
| 🖥️ |  |
| 📱  |  |

## Testing
```
$ npm run test:update

Cached:    0 cached, 7 total
  Time:    4.166s 
```

### DIFF
--- a/apps/availability/package.json
+++ b/apps/availability/package.json
@@ -7,6 +7,7 @@
     "start": "next dev -p 8000",
     "build": "next build",
     "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
     "test": "vitest --run",
     "test:watch": "vitest",
     "deploy:prod": "tools/deployDocker.sh"
--- a/apps/availability/src/components/SpinnerIcon.tsx
+++ b/apps/availability/src/components/SpinnerIcon.tsx
@@ -1,7 +1,7 @@
 export const SpinnerIcon: React.FC = () => {
   return (
     <svg
-      className="animate-spin h-5 w-5 text-gray-700"
+      className="animate-spin size-5 text-gray-700"
       xmlns="http://www.w3.org/2000/svg"
       fill="none"
       viewBox="0 0 24 24"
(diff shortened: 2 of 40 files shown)


tarinrickett (approved):
🙌 so glad to have this fix! thanks for the quick solutioning, Jan. agree with your additions as well.

we might want to run lint:fix as a git hook as well. we can follow up on this as a separate task

Issue #132, opened 2025-01-30
Add default typography to website-25 global.css


tarinrickett:
Default typography should be set on the element level. (Element specificity is lower specificity than classes etc., which allow us to provide default states while still allowing for custom overrides for independent use cases)

<img width="818" alt="Image" src="https://github.com/user-attachments/assets/34c2e351-2ab1-4adb-a38b-70a74f996e08" />

PR #190, opened 2025-02-04
Update default typography on `<h1>` & `<h2>`

janraasch:
**🚧 work in progress 🚧**

# Summary

Update typography on `<h1>` & `<h2>` to match [Figma Typography](https://www.figma.com/design/s4dNR4ELGKPbja6GkHLVJy/Website-Laura's-Working-File?node-id=7760-3924&m=dev).

## Issue

Part of #132.

## To-dos & Questions

* [ ] Do we currently use `.subtitle` & `.title` classes? If not, remove.
* [ ] Check `leading`-values: Currently those are not used on `Card` & `CourseCard` bc. they overwrite this using `text-2xl` which carries it's own `line-height`.

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📕 |  |
| 🖥️ |  |
| 📱  |  |

## Testing
```
$ npm run test:update
```

### DIFF
--- a/apps/website-25/src/globals.css
+++ b/apps/website-25/src/globals.css
@@ -5,9 +5,12 @@
   h1 {
     @apply text-bluedot-darker font-sans text-5xl font-bold leading-tight;
   }
+  /* TODO: Where is the title class used? */
+  /* TODO: Introduce font-demibold for font-[650] */
   h2, .title {
     @apply text-bluedot-darker font-sans text-2xl font-[650] leading-normal;
   }
+  /* TODO: Where is the subtitle class used? */
   h3, .subtitle {
     @apply text-bluedot-darker font-sans text-2xl font-[650] leading-[1.5];
   }
--- a/apps/website-25/src/globals.css
+++ b/apps/website-25/src/globals.css
@@ -6,7 +6,7 @@
     @apply text-bluedot-darker font-sans text-5xl font-bold leading-tight;
   }
   h2, .title {
-    @apply text-bluedot-darker font-sans text-5xl font-[700] leading-[1.5];
+    @apply text-bluedot-darker font-sans text-2xl font-[650] leading-normal;
   }
   h3, .subtitle {
     @apply text-bluedot-darker font-sans text-2xl font-[650] leading-[1.5];
--- a/apps/website-25/src/globals.css
+++ b/apps/website-25/src/globals.css
@@ -3,7 +3,7 @@
 
 @layer base {
   h1 {
-    @apply text-bluedot-darker font-sans text-5xl font-[700] leading-[1.25];
+    @apply text-bluedot-darker font-sans text-5xl font-bold leading-tight;
   }
   h2, .title {
     @apply text-bluedot-darker font-sans text-5xl font-[700] leading-[1.5];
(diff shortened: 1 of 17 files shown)


libraries/ui/src/Card.tsx
```diff
@@ -55,7 +55,7 @@ export const Card: React.FC<CardProps> = ({
 
       <div className="card__content flex flex-col gap-6 w-full flex-1 justify-between">
         <div className="card__text">
-          <h2 className="card__title text-2xl font-[650] text-bluedot-darker mb-1">{title}</h2>
+          <h2 className="card__title text-2xl text-bluedot-darker mb-1">{title}</h2>
```
tarinrickett, line 58: if we update this to the semantically correct h3, we should not need any custom typography styling:
```suggestion
          <h3 className="card__title mb-1">{title}</h3>
```

libraries/ui/src/CourseCard.tsx
```diff
@@ -61,7 +61,7 @@ const FeaturedCourseCard: React.FC<CourseCardProps & { footerRow: React.ReactNod
           <p className="course-card__featured-label uppercase font-[650] text-xs mb-3">
             Featured course
           </p>
-          <h2 className="course-card__title text-2xl font-semibold text-bluedot-darker mb-6">
+          <h2 className="course-card__title text-2xl text-bluedot-darker mb-6">
```
tarinrickett, line 64:
```suggestion
          <h2 className="course-card__title mb-6">
```

libraries/ui/src/Section.tsx
```diff
@@ -18,15 +18,15 @@ export const Section: React.FC<SectionProps> = ({
       <div className="section__title-container flex justify-between items-center gap-4">
         <div className="section__content flex-1">
           {title && (
-            <h2 className={clsx(
+            <h1 className={clsx(
```
tarinrickett, line 21: best practice is having only one `<h1>` tag per page. in our case, that is already defined in [HeroSection](https://github.com/bluedotimpact/bluedot/blob/86e791f7cafe9189c2c04bac8730f09d514973ed/libraries/ui/src/HeroSection.tsx#L19), which is why i would prefer Section to use `<h2>` and subtitles to use `<h3>`

libraries/ui/src/Card.tsx
tarinrickett: https://github.com/user-attachments/assets/4fcd311e-c9ae-4d1c-82d2-e3f83e68593f

apps/website-25/src/globals.css
```diff
@@ -3,11 +3,14 @@
 
 @layer base {
   h1 {
-    @apply text-bluedot-darker font-sans text-5xl font-[700] leading-[1.25];
+    @apply text-bluedot-darker font-sans text-5xl font-bold leading-tight;
   }
+  /* TODO: Where is the title class used? */
```
tarinrickett, line 8: good point - i dont think `.title` or `.subtitle` are used today. 

i introduced these as available aliases in case an override was needed somewhere (e.g., in the Card example below, if we couldn't use the semantic h3, we could apply `.subtitle` to any semantic element to get the same styles)

do you think it is worth it to have these aliases available for future use, to encourage folks to align to these definitions and avoid custom definitions or code dupe?

janraasch:
Hi @tarinrickett, thank you for your comments!!! I'll split this up a little bit. Let's start with a quick clean up: #207.

PR #239, opened 2025-02-10
SlideList refactor

### LINKED ISSUE #268: SlideList Refactor
#239

Will-Howard:
# Summary

- Update SlideList to collapse based on the container size rather than overall screen size
- Various changes to the styling of SlideList itself, and refactoring to reduce duplication between SlideList and Section
- Use SlideList everywhere it should be used

**Important note**: I wasn't able to easily update the tests to meaningfully check the desktop and mobile versions render as expected, because the logic in SlideList now depends a lot on CSS, and this doesn't appear to be applied properly in tests. Given the short deadline for the MVP I decided to go ahead without the tests, but I would like to come back and update them later. Also cc @janraasch for applying CSS in tests, I couldn't get it to work in the ~30 mins I spent trying

Before: 

https://github.com/user-attachments/assets/a18203dc-3550-4bf1-a4b3-33af78d0c5b2

After: 

https://github.com/user-attachments/assets/cba6de0a-5e87-4834-b54b-35865abf56e7

## Issue

Fixes #268, #315, #345

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | Screenshot 2025-02-13 at 11 16 35 |
| | Screenshot 2025-02-13 at 11 16 27 |
| 🖥️ | Screenshot 2025-02-13 at 10 58 16 |
| | Screenshot 2025-02-13 at 10 58 32 |
| | Screenshot 2025-02-13 at 10 58 42 |
| | Screenshot 2025-02-13 at 10 58 49 |
| | Screenshot 2025-02-13 at 10 59 03 |
| | Screenshot 2025-02-13 at 10 59 12 |
| | Screenshot 2025-02-13 at 10 59 21 |
| 📱  | Screenshot 2025-02-13 at 11 08 59 |
| | Screenshot 2025-02-13 at 11 09 04 |
| | Screenshot 2025-02-13 at 11 09 10 |
| | Screenshot 2025-02-13 at 11 09 19 |
| | Screenshot 2025-02-13 at 11 09 21 |
| | Screenshot 2025-02-13 at 11 09 29 |

## Testing
```
$ npm run test:update
Cached:    5 cached, 7 total
  Time:    4.621s
```


Will-Howard:
@tarinrickett I won't have time to fix the tests on this until this evening (it's quite fiddly because the CSS doesn't appear to be applied in the tests currently). Would you prefer I merged it earlier without the tests?

tarinrickett:
hey Will! thanks for calling out this bug! can you document the "before" experience with a screen recording in the PR for reference?

i'm wondering if we can revisit the functionality to reduce the complexity of this component... can we pause on the work while i sync with our PM?

how can we simplify this work? is the progress bar complicating things? if we removed that progress bar requirement, could we just translate the content by a passed in "element size" for now? 🤔

Will-Howard:
@tarinrickett 
> can you document the "before" experience with a screen recording in the PR for reference?

Yep, I've added a quick before/after in the PR description, otherwise I'll pause working on this for now

> how can we simplify this work?

I think there are two things we can do:
1. Separate out much of the layout logic that is in this component, and make it just handle the scrolling list itself. Currently it duplicates a lot of what is in `Section`, and also has a space for the featured card. I assume the reason we decided to combine these things in the first place is that the buttons overlap vertically with the section heading (so putting these buttons below a standard section heading would create too much space), but maybe the design could be changed to simplify this
2. Fixing the size of the items, and making the `slide-list__container` a regular (horizontally) scrollable element, rather than trying to make the items exactly fit. We could keep the left and right buttons as a convenience and just have them do a `scrollTo` to the next item

> is the progress bar complicating things? if we removed that progress bar requirement, could we just translate the content by a passed in "element size" for now?

Fixing the element size would simplify things (separate from the progress bar), I think if we did that then keeping the progress bar wouldn't make too much difference to the complexity

libraries/ui/src/Section.tsx
```diff
@@ -1,55 +1,57 @@
 import React from 'react';
 import clsx from 'clsx';
-import { CTALinkOrButton } from './CTALinkOrButton';
```
Will-Howard, line 3: Changes to this file (you will probably want to view it side-by-side):
- Remove the `CTALinkOrButton` right hand node (which is no longer used), and replace it with a generic `rightNode` (which is now used by `SlideList`)
- Factor out the heading element into `SectionHeading`, so it can be reused with consistent styling by `SlideList`
- Factor out the styling applied to the outer wrapper into the `section-body` class, so other components (`SlideList`) can act as sections just by adding this class
- Support different heading levels for the subtitle as well

So `Section` is now a convenience wrapper for `SectionHeading` and the `section-body` class, and these are intended to be reusable elsewhere

apps/website-25/src/components/homepage/CommunitySection/TestimonialSection.test.tsx
```diff
@@ -8,18 +8,6 @@ describe('TestimonialSection', () => {
...
-    render(<TestimonialSection />);
-
-    // Check for testimonial content
-    const testimonials = screen.getAllByRole('blockquote');
-    expect(testimonials).toHaveLength(3);
-
-    // Verify slide navigation does not exist
-    expect(screen.queryByLabelText('Previous slide')).toBeNull();
-    expect(screen.queryByLabelText('Next slide')).toBeNull();
-  });
```
Will-Howard, line 21: This is one of the tests that I punted on fixing (#347)

apps/website-25/src/components/homepage/CommunitySection/index.tsx
```diff
@@ -6,7 +6,7 @@ import CommunityValuesSection from './CommunityValuesSection';
 
 const CommunitySection = () => {
   return (
-    <Section title="Our community" className="community-section">
+    <Section className="community-section">
```
Will-Howard, line 9: Title moved into `CommunityValuesSection` to allow the buttons to live in the header

libraries/ui/src/CourseCard.tsx
```diff
@@ -136,7 +136,7 @@ export const CourseCard: React.FC<CourseCardProps> = ({
         isEntireCardClickable
         className={clsx(
           'course-card course-card--regular container-lined p-5',
-          'flex flex-col w-[323px] h-[466px] hover:container-elevated',
+          'flex flex-col max-w-full h-[520px] hover:container-elevated',
```
Will-Howard, line 139: Width is now set by the parent, it would still be better not to hardcode the height

libraries/ui/src/SlideList.tsx
```diff
@@ -1,230 +1,231 @@
...
-  itemsPerSlide = 1,
-  slideClassName,
-  containerClassName,
-  slidesWrapperWidth = { mobile: '100%', desktop: '800px' },
+  maxItemsPerSlide = 1,
+  minItemWidth = 260,
+  title,
+  titleLevel,
+  subtitle,
+  subtitleLevel,
```
Will-Howard, line 24: This component has ended up very complex unfortunately

libraries/ui/src/Section.tsx
```diff
@@ -1,55 +1,57 @@
...
+  subtitleLevel?: 'p' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  rightNode?: React.ReactNode;
+};
 
-export const Section: React.FC<SectionProps> = ({
-  className, title, titleLevel = 'h2', subtitle, ctaText, ctaUrl, children,
+export type SectionHeadingProps = BaseProps;
+export type SectionProps = React.PropsWithChildren<BaseProps>;
+
+export const SectionHeading: React.FC<BaseProps> = ({
```
domdomegg, line 16: should: Section probably shouldn't have all the title / subtitle props, given these could be composable.

libraries/ui/src/SlideList.tsx
domdomegg: I think this could do with a simplification refactor, but it's already a big improvement

domdomegg (approved):
LGTM: this clearly looks and works a lot better. Code seems fine.

Can probably still make improvements, but keen to get this in the main branch so we can build on top of it.

apps/website-25/src/components/homepage/CourseSection.test.tsx
```diff
@@ -1,33 +1,20 @@
...
 
     // Verify featured course is present
     expect(screen.getByText('AI Safety: Intro to Transformative AI')).toBeDefined();
 
-    // Verify navigation is in header on desktop
-    const header = container.querySelector('.slide-list__header');
+    // Verify navigation is present in header
+    // TODO Make this check the desktop vs mobile behaviour
+    const header = container.querySelector('.slide-list__nav--header');
     expect(header?.querySelector('[aria-label="Previous slide"]')).toBeDefined();
```
tarinrickett, line 18: nit: see above - i dont think we need to have tests on SlideList behavior across consumer use cases. it just adds unnecessary overhead imo

libraries/ui/src/Section.tsx
tarinrickett: thanks for all the clarification, Will! really appreciate the documentation

libraries/ui/src/SlideList.tsx
```diff
@@ -1,230 +1,231 @@
...
+
   const childrenArray = React.Children.toArray(children) as React.ReactElement[];
-  const totalSlides = Math.ceil(childrenArray.length / effectiveItemsPerSlide);
+  const allChildrenFit = childrenArray.length <= itemsPerSlide;
 
-  const handlePrevious = () => {
-    setCurrentSlide((prev) => Math.max(0, prev - 1));
-  };
+  // If there are multiple slides, show 15% of the next card to signal that the section is scrollable
+  const peekAdjustment = allChildrenFit ? '1' : `(1 - ${0.15 / itemsPerSlide})`;
```
tarinrickett, line 108: 🙌 💯 

also "peekAdjustment" is such a cute name

tarinrickett (commented):
this is awesome Will. thanks for taking the initiative on making _so_ many fixes – i appreciate your quick work ahead of MVP 🌟 

i agree there are a number of ways we could simplify this logic moving forward, based on revisits from designs. i feel strongly about decoupling this component from the section headers, for instance. i've raised this with our PM who is directing the designers to revisit it as well. added tracking here: https://github.com/bluedotimpact/bluedot/issues/344

again - loving the ownership and initiative here. Thank you !!!

Issue #314, opened 2025-02-12
Mobile > About > Our team: People's names and LinkedIn buttons overlap the thumbnails


vioricagh:
Tested on iPhone 13 Pro Max, Safari, iOS V 18.3 

Mobile > About > Our team: People's names and LinkedIn buttons overlap the thumbnails, please see ss attached. 

![Image](https://github.com/user-attachments/assets/880ecba7-aa4c-4f32-b5b3-2d894d00a36f)

PR #324, opened 2025-02-12
Fix card renders on TeamSection

tarinrickett:
# Summary

Fix card renders on TeamSection

## Description

- When we decentralized width/heights from Card, it looks like we inadvertently introduced unrestrained card size renders on the Team page 😅 
- This is a **quick fix** ahead of further bug bashing early tomorrow AM
- We should revisit as a team to consider how we can prevent this from occurring in the future: 

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | image |
| 🖥️ | https://github.com/user-attachments/assets/82da8738-0c82-44ab-bbaa-74a97b742692 |
| 📱  | image |

## Testing
```
$ npm run test:update
 Tasks:    1 successful, 1 total
Cached:    0 cached, 1 total
  Time:    3.161s 
```


domdomegg:
Does this also fix #314? I don't have an iOS device to hand to test this.

Issue #299, opened 2025-02-12
[feedback] website: on mobile, whitespace above testimonials section looks odd


domdomegg:
![Image](https://github.com/user-attachments/assets/e84bdd11-8703-49a3-961c-c1955995c128)

tarinrickett:
Moving this under #344 as we need larger scale design support for this improvements

Issue #295, opened 2025-02-12
website: new website links to old bluedot.org


domdomegg:
To deploy the new website on the bluedot.org domain, we'll probably need to remove the old website (or move it somewhere).

Things that need migrating?
- blog
- job postings

domdomegg:
Also:
* projects

Issue #290, opened 2025-02-12
[feedback] website: privacy policy bullets are too close to text and each other


domdomegg:
![Image](https://github.com/user-attachments/assets/e9632f36-64eb-4b4d-91eb-e50916bc17de)

PR #348, opened 2025-02-13
website: Add contact page

domdomegg:
# Summary

Add a contact page, given we link to this a bunch of places and is one of the things people try to find the most on our site. This is part of wider work under #295 to migrate pages that are currently only on old website.

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📕 |  |
| 🖥️ | image image |
| 📱  | image |

Mobile looks a little cramped, but this is because we use the same components as elsewhere. When these are fixed (#299, #290, etc.) presumably this will be too.

### DIFF
--- a/apps/website-25/src/__tests__/pages/__snapshots__/about.test.tsx.snap
+++ b/apps/website-25/src/__tests__/pages/__snapshots__/about.test.tsx.snap
@@ -802,40 +802,6 @@ exports[`AboutPage > should render the error message correctly 1`] = `
         </a>
       </div>
     </div>
-    <div
-      class="section section-normal py-spacing-y border-b border-divider overflow-hidden"
-    >
-      <div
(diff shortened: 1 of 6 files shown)


tarinrickett (changes requested):
hi adam, we do not want to add a contact us page in the new site architecture. we will redirect /contact to /about

domdomegg:
We link to the contact page in a lot of places, and redirecting to the about page instead I think will be confusing in most of these cases - e.g. a lot of links like 'get in touch'.

I think we should reconsider not having a contact page, or make it clearer to people landing here how to contact us - currently this user experience would be signfiicantly worse.

domdomegg:
Converting to draft: we want to change the image and flip some of the text before merging, and only want to do so after the web launch

domdomegg:
Superceded by #504

Issue #359, opened 2025-02-13
Be able to run proxy servers locally more easily


domdomegg:
Currently it requires faffing with docker/installing nginx

Desired state: all the *-proxy apps can be run with `npm start` without special dependencies

PR #360, opened 2025-02-13
website-proxy: Add proxy for web hosting

domdomegg:
# Summary

Add a proxy for routing traffic between old and new websites, so we can have both simultaneously on bluedot.org.

## Issue

Prepares for #295 - just need to switch the domain after testing this does in fact, work.

## Developer checklist

NB: this is not very well tested because nginx is a pain to run locally atm - will fix this later (#359)

### DIFF
--- a/apps/website-proxy/src/nginx.template.conf
+++ b/apps/website-proxy/src/nginx.template.conf
@@ -1,0 +1,26 @@
+http {
+    server {
+        listen 80;
+        add_header X-BlueDot-Version '$VERSION_TAG';
+
+        location /blog/ {
+            proxy_ssl_server_name on;
+            proxy_set_header Host bluedot.org;
+            proxy_pass http://45.76.132.116;
+        }
+
+        # Some static resources e.g. images for blog posts
+        location /u/ {
+            proxy_ssl_server_name on;
+            proxy_set_header Host bluedot.org;
+            proxy_pass http://45.76.132.116;
+        }
+
+        location / {
+            proxy_ssl_server_name on;
+            proxy_pass https://website-25.k8s.bluedot.org;
+        }
+    }
+}
+
+events {}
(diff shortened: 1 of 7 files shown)


apps/website-proxy/src/nginx.template.conf
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        # Some static resources e.g. images for blog posts
+        location /u/ {
+            proxy_ssl_server_name on;
+            proxy_set_header Host bluedot.org;
+            proxy_pass http://45.76.132.116;
+        }
+
+        location / {
```
tarinrickett, line 19: synced offline: for initial config test, we will update this just to target /about. post that, we will inverse the config approach here and keep just the following targeting new site:
1. location /about
2. location = /
3. location /careers
4. location /privacy-policy

libraries/backend-contract/package.json
```diff
@@ -16,7 +16,7 @@
     "devDependencies": {
         "@bluedot/eslint-config": "*",
         "@bluedot/typescript-config": "*",
-        "@types/node": "^20.12.12",
+        "@types/node": "^22.13.2",
```
tarinrickett, line 19: curious why did bumping `@types/node` come up in this change?

domdomegg: Deps were being a pain because package-lock.json got out of sync with package.json at some point, I think related to this merge commit: https://github.com/bluedotimpact/bluedot/pull/360/commits/24db0b2c957422f90ff7e22a9b1e9996822266c2

Additionally, we had actually locked in later versions in the package-lock.json that weren't recorded in package.json.

As I was fiddling with the other deps including the lock of @types/node, I spotted it needed updating anyway given we're on Node 22.

PR #361, opened 2025-02-13
Add `utm` tracking for `CourseSection`

### LINKED ISSUE #276: Add CTR Tracking for Course Cards
See [Next.js Docs - sending events](https://nextjs.org/docs/app/building-your-application/optimizing/third-party-libraries#sending-events-1).

Sync on `event` & name and params with @tarinrickett.

janraasch:
# Summary

Enable CTR tracking for clicks on course cards.

## Issue

Fixes #276.

## Description

* Add `utmParams` helper
* Use on `CourseCards`

## Developer checklist

## Testing
```
$ npm run test:update
Cached:    0 cached, 7 total
  Time:    5.33s 
```

### DIFF
--- a/apps/website-25/src/lib/utmParams.ts
+++ b/apps/website-25/src/lib/utmParams.ts
@@ -1,20 +1,20 @@
 // TODO: Possibly move to future @bluedot/analytics package
-export type SOURCE = 'website'
+export type SOURCE = 'website';
 
-export type CONTENT = 'course_section'
+export type CONTENT = 'course_section';
 
-export type CAMPAIGN = 'relaunch'
+export type CAMPAIGN = 'relaunch';
 
 export interface UTMParams {
-    source: SOURCE;
-    content: CONTENT;
-    campaign: CAMPAIGN;
+  source: SOURCE;
+  content: CONTENT;
+  campaign: CAMPAIGN;
 }
 
 export const addUtmParams = (url: string, params: UTMParams): string => {
-    const urlObj = new URL(url);
-    urlObj.searchParams.set('utm_source', params.source);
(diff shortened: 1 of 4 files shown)


apps/website-25/src/lib/utmParams.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type CAMPAIGN = 'relaunch';
+
+export interface UTMParams {
+  source: SOURCE;
+  content: CONTENT;
+  campaign: CAMPAIGN;
+}
+
+export const addUtmParams = (url: string, params: UTMParams): string => {
```
domdomegg, line 14: I think to maintain parity with the previous version, we'd want to inherit params from our URL, and only override them if the user doesn't have any params.

This enables tracking through from e.g. an ad click onto registering for the course

janraasch: The objective here was to enable **ctr** tracking for the **course cards** *specifically*. If we keep existing `utm` params here, we would not be able to count those on the **course card ctr** report.

tarinrickett: + @vioricagh

tarinrickett: hi @janraasch ! i was able to sync up with the team offline. 

✨ **we are good to move forward with utm approach!** 🚀 

we will need to revisit our event tracking in the future, which you and I have already discussed, to make this more robust – but in the meanwhile, there are no ads being run that would be pointing to bluedot.org that this utm approach might interfere with. so all OK 🎉

tarinrickett (approved):

PR #383, opened 2025-02-14
infra: Avoiding having to manually add apps to cd_cd config

### LINKED ISSUE #382: ci/cd: Avoid having to add each app manually to the ci/cd configuration
At the moment, the process for adding a new app requires adding a block in the ci_cd.yaml for GitHub Actions.

We want to be able to easily create new apps to quickly deploy experiments etc, so the easier the process is the better. This part is also particularly fiddly so it'd be good to cut it out completely.

I'm imagining that somehow the cd part loops over all the apps that need deploying, to deploy them. And similarly for filter paths. Maybe something like this is the answer?: https://stackoverflow.com/questions/71679906/how-to-github-action-parallelize-tests-dynamically-by-folder

domdomegg:
# Summary

Update GitHub Actions configuration to discover apps, rather than having to specify them explicitly

## Description

At the moment, adding a new app requires adding a block in the ci_cd.yaml for GitHub Actions.

We want to be able to easily create new apps to quickly deploy experiments etc, so the easier the process is the better. This part is also particularly fiddly so it'd be good to cut it out completely.

I've updatd the ci_cd.yaml to loops over all the apps that need deploying, and then deploys the ones that need deploying. I've also taken the opportunity to properly get the dependencies for the apps, e.g. so a change in libraries/ui results in website deploying, as one might expect (Currently apps will deploy only if you edit files within their app folder, and not their dependencies - although they will lint/build/test if dependencies are changed due to turbo. After this PR they will also deploy if dependencies are changed).

NB: I think these changes should work. But because [it's a pain to test GitHub actions locally](https://stackoverflow.com/questions/59241249/how-can-i-run-github-actions-workflows-locally), there may be some follow-ups for things that weren't immediately obvious.

## Issue

Fixes #382

### DIFF
--- a/.github/workflows/ci_cd.yaml
+++ b/.github/workflows/ci_cd.yaml
@@ -40,34 +40,36 @@ jobs:
       - name: Test
         run: npm run test --if-present
 
-      # Get paths that apps are dependent on, e.g. availability depends on apps/availability/** and libaries/ui/**
-      - id: get_paths_filter
+      - name: Get paths that apps are dependent on
+        # e.g. availability depends on apps/availability/** and libaries/ui/**
+        id: get_paths_filter
         run: |
           node .github/workflows/create-paths-filter.js > paths-filter.json
           FILTER=$(cat paths-filter.json)
           echo "paths_filter=$FILTER" >> $GITHUB_OUTPUT
 
-      # Figure out what apps have changed files in their paths
-      - id: modified_paths_check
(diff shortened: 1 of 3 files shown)


domdomegg:
FYI: vultr has an outage which is causing pushing containers to container registry to fail intermittently. I'm going to wait until that's resolved before merging.

PR #453, opened 2025-02-18
Fix SlideList snapping to next item on click

### LINKED ISSUE #443: Clicking on the second course card in a list moves the list forward instead of navigating to the course

Will-Howard:
# Summary

I've replaced a custom `'focusin'` handler for tabbing through the list with native CSS [scroll-snap behaviour](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type)

Before (scrolls on click rather than following link when clicking on the second item):

https://github.com/user-attachments/assets/2285ce67-52ec-4434-8f58-446c6071b02c

After (follows links, and always snaps to the start of the current item):

https://github.com/user-attachments/assets/fdacda86-1769-4219-bf95-12fe34fcdcfc

## Issue

Fixes #443 

## Developer checklist
- [x] [N/A] Used [BEM naming convention](https://getbem.com/naming/) for classNames
- [x] [N/A] Added or updated Jest tests
- [x] [N/A] Added or updated Storybook stories

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | Screenshot 2025-02-18 at 07 31 40 |
| 🖥️ | Screenshot 2025-02-18 at 07 35 55 |
| 📱  | Screenshot 2025-02-18 at 07 36 24 |

## Testing
```
$ npm run test:update
Cached:    5 cached, 7 total
  Time:    4.588s
```


libraries/ui/src/SlideList.tsx
```diff
@@ -89,7 +89,8 @@ export const SlideList: React.FC<SlideListProps> = ({
     const childArray = Array.from(container.children) as HTMLDivElement[];
 
     // Jump forward/back by itemsPerSlide
-    const currentIndex = childArray.findIndex((c) => c.offsetLeft >= scrollLeft);
+    const tolerancePx = 5;
```
tarinrickett, line 92: how did we come up with `5`?

libraries/ui/src/SlideList.tsx
```diff
@@ -107,6 +108,8 @@ export const SlideList: React.FC<SlideListProps> = ({
   // If there are multiple slides, show 15% of the next card to signal that the section is scrollable
   const peekAdjustment = allChildrenFit ? '1' : `(1 - ${0.15 / itemsPerSlide})`;
   const itemWidth = `calc((${100 / itemsPerSlide}% - var(--space-between) * ${(itemsPerSlide - 1) / itemsPerSlide}) * ${peekAdjustment})`;
+  // Threshold for scrolling to the next item when tabbing through the list, only needs to be approximate
+  const scrollPaddingPx = `${(measuredContainerWidth ?? minItemWidth * itemsPerSlide) * 0.15}px`;
```
tarinrickett, line 112: when would we not have `measuredContainerWidth`?

tarinrickett (approved):
lgtm, just some clarifying questions 🫡

libraries/ui/src/SlideList.tsx
Will-Howard: Only on the first render before the `useEffect` has run, so I don't think this would ever make a difference to end users

libraries/ui/src/SlideList.tsx
Will-Howard: It just needs 1 or 2px leeway depending on how the snapping settles, so 5 is a fairly comfortable amount of tolerance. There's no reason for the number 5 specifically

PR #459, opened 2025-02-19
Add breadcrumbs component

### LINKED ISSUE #397: Add Breadcrumb component and integrate across pages
Image

Will-Howard:
# Summary

1. Added breadcrumbs component
2. Refactored all internal navigation to use predefined route urls and titles. The idea here is to make it easier to maintain consistency between the tab name, breadcrumb title, and any other places the title might appear

## Issue

Fixes #397 

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | Screenshot 2025-02-19 at 10 27 36 |
| 🖥️ | Screenshot 2025-02-19 at 10 28 00 |
| 📱  | Screenshot 2025-02-19 at 10 28 25 |

## Testing
```
$ npm run test:update
Cached:    4 cached, 7 total
  Time:    4.765s
```

### DIFF
--- a/apps/website-25/src/lib/routes.ts
+++ b/apps/website-25/src/lib/routes.ts
@@ -1,0 +1,30 @@
+export type BluedotRoute = {
+  /**
+   * Title of the page, by convention this is the title that appears in the
+   * <title /> tag (e.g. "About us" in "About us | BlueDot Impact")
+   */
+  title: string;
+  /**
+   * Relative url of the route (e.g. /about, not https://bluedot.org/about)
+   */
+  url: string;
+};
+
+export const ROUTES = {
+  home: {
+    title: 'BlueDot Impact',
+    url: '/',
+  },
+  about: {
+    title: 'About us',
+    url: '/about',
+  },
+  privacyPolicy: {
+    title: 'Privacy Policy',
+    url: '/privacy-policy',
+  },
+  careers: {
+    title: 'Join our team',
+    url: '/careers',
+  },
+} as const;
--- a/apps/website-25/src/lib/routes.ts
+++ b/apps/website-25/src/lib/routes.ts
@@ -1,30 +1,31 @@
-export type BluedotRoute = {
-  /**
-   * Title of the page, by convention this is the title that appears in the
-   * <title /> tag (e.g. "About us" in "About us | BlueDot Impact")
-   */
-  title: string;
-  /**
-   * Relative url of the route (e.g. /about, not https://bluedot.org/about)
-   */
-  url: string;
+import type { BluedotRoute } from '@bluedot/ui';
+
+const home: BluedotRoute = {
+  title: 'BlueDot Impact',
+  url: '/',
+};
+
+const about: BluedotRoute = {
+  title: 'About us',
+  url: '/about',
+  parentPages: [home],
+};
+
+const privacyPolicy: BluedotRoute = {
+  title: 'Privacy Policy',
+  url: '/privacy-policy',
+  parentPages: [home],
+};
+
+const careers: BluedotRoute = {
+  title: 'Join our team',
+  url: '/careers',
+  parentPages: [home],
 };
 
 export const ROUTES = {
-  home: {
-    title: 'BlueDot Impact',
-    url: '/',
-  },
-  about: {
-    title: 'About us',
-    url: '/about',
-  },
-  privacyPolicy: {
-    title: 'Privacy Policy',
-    url: '/privacy-policy',
-  },
-  careers: {
-    title: 'Join our team',
-    url: '/careers',
-  },
+  home,
+  about,
+  privacyPolicy,
+  careers,
 } as const;
--- a/apps/website-25/src/pages/_app.tsx
+++ b/apps/website-25/src/pages/_app.tsx
@@ -12,6 +12,7 @@ import clsx from 'clsx';
 import { useEffect } from 'react';
 import { Router } from 'next/router';
 import { Analytics } from '../components/Analytics';
+import { ROUTES } from '../lib/routes';
 
 if (typeof window !== 'undefined') {
   if (!process.env.NEXT_PUBLIC_POSTHOG_KEY) {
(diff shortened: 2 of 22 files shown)


libraries/ui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <div className={clsx('breadcrumbs border-b border-color-divider w-full py-3', className)}>
+      <nav
+        className="breadcrumbs__nav section-base"
+        aria-label="Breadcrumbs"
+      >
+        <ol className="breadcrumbs__list flex">
+          {items.map((item, index) => (
+            <li key={item.url} className="breadcrumbs__item flex items-center text-size-xs">
```
Will-Howard, line 23: I wrote it with a 14px font size and 12px padding, according to the designs. To make it more standardised we could instead use the default font size and `py-space-between`, this is what that would look like:
Screenshot 2025-02-19 at 10 29 46
Screenshot 2025-02-19 at 10 29 58

apps/website-25/src/pages/_app.tsx
```diff
@@ -58,14 +59,14 @@ const App: React.FC<AppProps> = ({ Component, pageProps }: AppProps) => {
         logo="/images/logo/BlueDot_Impact_Logo.svg"
         courses={constants.COURSES}
       >
-        <a href="/about" className={clsx('hover:text-bluedot-normal', isCurrentPath('/about') && 'font-bold')}>About us</a>
-        <a href="/careers" className={clsx('hover:text-bluedot-normal', isCurrentPath('/careers') && 'font-bold')}>Join us</a>
+        <a href={ROUTES.about.url} className={clsx('hover:text-bluedot-normal', isCurrentPath(ROUTES.about.url) && 'font-bold')}>About us</a>
+        <a href={ROUTES.careers.url} className={clsx('hover:text-bluedot-normal', isCurrentPath(ROUTES.careers.url) && 'font-bold')}>Join us</a>
```
Will-Howard, line 63: I've attempted to replace all hardcoded strings used for internal navigation, so if you find any I have missed please let me know

Will-Howard (commented):
Here is an exhaustive list of everywhere the <Breadcrumbs /> component now appears:
Screenshot 2025-02-19 at 10 39 22
Screenshot 2025-02-19 at 10 39 31
Screenshot 2025-02-19 at 10 39 38
Screenshot 2025-02-19 at 10 39 49
Screenshot 2025-02-19 at 10 39 59

libraries/ui/src/Breadcrumbs.tsx
tarinrickett: grrrreeeeeaat callout Will. let's go with default font size and py-space-between for standardization 🙏

apps/website-25/src/lib/routes.ts
```diff
@@ -0,0 +1,30 @@
...
+  title: string;
+  /**
+   * Relative url of the route (e.g. /about, not https://bluedot.org/about)
+   */
+  url: string;
+};
+
+export const ROUTES = {
+  home: {
+    title: 'BlueDot Impact',
```
tarinrickett, line 15: the designs have this as "Bluedot.org" but I'm also thinking "Home" might be best? will sync with PM tomorrow to confirm

```suggestion
    title: 'Home',
```

apps/website-25/src/pages/careers/ai-safety-teaching-fellow.tsx
```diff
@@ -5,22 +5,30 @@ import {
...
   Section,
+  Breadcrumbs,
 } from '@bluedot/ui';
 import Head from 'next/head';
+import { BluedotRoute, ROUTES } from '../../lib/routes';
+
+export const CURRENT_ROUTE: BluedotRoute = {
+  title: 'AI Safety Teaching Fellow',
+  url: `${ROUTES.careers.url}/ai-safety-teaching-fellow`,
+};
```
tarinrickett, line 16: i actually quite like colocating the route def in the `/pages/` route file itself. what do you think about doing the same for all pages?

apps/website-25/src/lib/routes.ts
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * Title of the page, by convention this is the title that appears in the
+   * <title /> tag (e.g. "About us" in "About us | BlueDot Impact")
+   */
+  title: string;
+  /**
+   * Relative url of the route (e.g. /about, not https://bluedot.org/about)
+   */
+  url: string;
+};
```
tarinrickett, line 11: what about including the definition of parent page(s) within the BluedotRoute type, with the default pointing to `home`?
```
parentPages: [BluedotRoute]
```

tarinrickett: im concerned that requiring consumers to manually set the path introduces a fragile pattern, so i'm trying to minimize that where we can...

i poked around a little bit, and i'm also not convinced that dynamically generating these paths is a robust enough solution to this. e.g. [this one](https://medium.com/@kcabading/creating-a-breadcrumb-component-in-a-next-js-app-router-a0ea24cdb91a) breaks when we introduce page titles that differ from their route name (e.g. /about vs About Us) 😅 i think your proposal works great and serves our purposes 💯

tarinrickett (commented):
a few small notes. will get Vio's eyes on this tomorrow!

apps/website-25/src/pages/careers/ai-safety-teaching-fellow.tsx
domdomegg: sepearate idea: can we write a test that:
- uses node:fs to get all the pages recusively in the pages directory
- uses `await import()` to dynamically import them, and get the CURRENT_ROUTE
- check that the url matches what it should based on the Next.js routing logic

this should enforce that the url is actually correct.

libraries/ui/src/CookieBanner.tsx
```diff
@@ -5,6 +5,7 @@ import posthog from 'posthog-js';
 import { CTALinkOrButton } from './CTALinkOrButton';
 
 type CookieBannerProps = {
+  cookiePolicyUrl: string,
```
domdomegg, line 8: I think this is _probably_ not worth parameterising. We only have one privacy policy across everything, so hard coding it within the component seems fine (possibly to bluedot.org/privacy-policy if we expect to use it across different apps and we're worried the relative path won't work)

domdomegg (approved):
LGTM, would be happy to merge as-is and for you to make the changes in a follow-up

apps/website-25/src/lib/routes.ts
Will-Howard: Cool, just let me know 👍. I would like to try and keep the tab title the same if possible, which is why I called it "BlueDot Impact" instead of "Bluedot.org" (sorry I forgot to flag that). So if changing it to "Home" I think it would make sense to change the title from:
`<title>{CURRENT_ROUTE.title} | Industry-leading free AI courses and career support</title>`
to:
`<title>{CURRENT_ROUTE.title} | BlueDot Impact</title>`

apps/website-25/src/pages/careers/ai-safety-teaching-fellow.tsx
Will-Howard: > i actually quite like colocating the route def in the /pages/ route file itself. what do you think about doing the same for all pages?

I agree this would be better if possible. The reason I didn't do that is that it creates the potential for circular dependencies: A parent page is likely to contain a link to a child page, so will need to import it, and then the child will need to import the parent to reference the route in the breadcrumbs. I did it in this case because these are known to be "leaf" pages that don't have further children.

In the current site we get a dependency cycle between `/careers` and these two specific job listings if we define the route inside the `careers.tsx` file.

apps/website-25/src/lib/routes.ts
Will-Howard: >what about including the definition of parent page(s) within the BluedotRoute type

Great idea 😌, this does simplify things a lot. I tried setting the deafult the default as `home` by adding a function like this:
```typescript
export function defineRoute({ title, url, childOf = home }: { title: string; url: string; childOf?: BluedotRoute }): BluedotRoute {
  return {
    title,
    url,
    parentPages: childOf ? [...(childOf.parentPages ?? []), childOf] : [],
  };
}
```

Again I ended up thinking that given the number of pages currently on the site it's not worth the extra complexity to have this function, so I have kept it that you have to referece `home` manually`:
```
const about: BluedotRoute = {
  title: 'About us',
  url: '/about',
  parentPages: [home],
};
```

Will-Howard:
Thanks for the reviews 😄. I'll merge as is as Adam suggested and followup later based on the open discussions

PR #460, opened 2025-02-19
Add multistage deploy for `website-25`

### LINKED ISSUE #58: Set up staging environment

janraasch:
# Summary

* Add `bluedot-website-25-production`-container
* Introduce *multistage* deployment workflow for `website-25`

## Issue

Fixes #58, but will require a few follow ups to actually the switch to the new production environment. See `### Rollout` ⬇️.

### Deployment Workflow

* **staging** Push to `master` triggers a deploy to `bluedot-website-25`. _This is the current behavior._
* **production** Push a `website-release*`-tag to `master` to trigger a deploy to `bluedot-website-25-production`. 

We can even choose to make it a convention to create `GitHub Releases` for production deploys.

Screenshot 2025-02-19 at 11 01 42

That way we can add a quick description of the release & keep a transparent history of releases.

### Rollout

* [x] **Before merging** comment out the `no index robots.txt`-code in `deploy-staging.sh`, see 4f2dcdb05f6e0d30db1fb12312af1488f4a7ad63.
* [ ] Merge this pr. No visual changes. Check that everything is working as before.
* [ ] Push a tag (or create a release) to trigger our first production deploy.
* [ ] Check that the new production container is working.
* [ ] **Make the switch**: Update the proxy to point to the new production container.
* [ ] Re-enable `no index robots.txt`-code in `deploy-staging.sh`.
* [ ] Change stanging env vars in `.env.staging.template` to point to staging GA & Posthog.
* [ ] Use staging tracking keys for dev, i.e. update `env.local.template`
* [ ] That's it 🎉 - Double-check that the GA & PostHog keys are correct on both environments.

ℹ️ I prepared a branch implementing the steps outlined above over at https://github.com/janraasch/bluedot/pull/1.

### Reasoning

This implementation draws inspiration from [Next.js - Examples - with-docker-multi-env](https://github.com/vercel/next.js/tree/canary/examples/with-docker-multi-env). While it requires separate builds (both npm run build and Docker image creation) for staging and production environments, it preserves Next.js's built-in optimization of inlining frontend environment variables. The solution maintains code simplicity without introducing additional dependencies. I believe this approach offers a balanced trade-off, and may be extended nicely to other multistage apps in the future. Happy to get your feedback ❤️!

## Developer checklist

Testing GitHub actions before merging is ~~hard~~ impossible? 😇. Since we introduce a new workflow & only made a small change to the existing `ci_cd`-workflow, we should be good to go after thorough code review 👀 🤝.

## Testing
```
$ npm run test:update
Cached:    4 cached, 7 total
  Time:    4.992s 
```

### DIFF
--- a/.github/workflows/ci_cd.yaml
+++ b/.github/workflows/ci_cd.yaml
@@ -115,4 +115,4 @@ jobs:
 
       - name: Deploy
         if: steps.check_deployment.outputs.should_skip != 'true'
-        run: npm run deploy:prod --workspace apps/${{ matrix.app }}
+        run: npm run deploy:cd --workspace apps/${{ matrix.app }}
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Inside each package folder, the common files you'll find are:
   - `npm run lint`: Check for lint issues. Visual Studio Code should usually highlight these for you already.
   - `npm run build`: Build the application. This usually finds any type errors, which Visual Studio Code should usually highlight for you already.
   - `npm run postinstall`: Perform any extra steps to setup the application for development. Usually things like creating configuration files for local development. You usually don't need to run this manually, as it runs when you run `npm install`.
-  - `npm run deploy:cd`: Actually deploy the app into the production (real-world) environment. You usually don't need to run this manually, as it is run by CD tooling when you merge your changes into the master branch.
+  - `npm run deploy:cd`: Actually deploy the app, usually into the production (real-world) environment. You usually don't need to run this manually, as it is run by CD tooling when you merge your changes into the master branch.
 - `README.md`: documentation to explain what the package does, how to use it, and how to contribute
 - `src`: most of the code usually lives here. You usually want to edit files in this folder.
   - `pages`: pages in the web app. For example `pages/some-page.tsx` usually corresponds to `app.bluedot.org/some-page`. The `api` folder contains API routes rather than webpages ([learn more in the Next.js docs](https://nextjs.org/docs/pages/building-your-application/routing/api-routes)).
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Inside each package folder, the common files you'll find are:
   - `npm run lint`: Check for lint issues. Visual Studio Code should usually highlight these for you already.
   - `npm run build`: Build the application. This usually finds any type errors, which Visual Studio Code should usually highlight for you already.
   - `npm run postinstall`: Perform any extra steps to setup the application for development. Usually things like creating configuration files for local development. You usually don't need to run this manually, as it runs when you run `npm install`.
-  - `npm run deploy:prod`: Actually deploy the app into the production (real-world) environment. You usually don't need to run this manually, as it is run by CD tooling when you merge your changes into the master branch.
+  - `npm run deploy:cd`: Actually deploy the app into the production (real-world) environment. You usually don't need to run this manually, as it is run by CD tooling when you merge your changes into the master branch.
 - `README.md`: documentation to explain what the package does, how to use it, and how to contribute
 - `src`: most of the code usually lives here. You usually want to edit files in this folder.
   - `pages`: pages in the web app. For example `pages/some-page.tsx` usually corresponds to `app.bluedot.org/some-page`. The `api` folder contains API routes rather than webpages ([learn more in the Next.js docs](https://nextjs.org/docs/pages/building-your-application/routing/api-routes)).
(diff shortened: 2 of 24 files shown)


.github/workflows/ci_cd.yaml
```diff
@@ -114,6 +114,11 @@ jobs:
           K8S_KUBECONFIG: ${{ secrets.K8S_KUBECONFIG }}
           VULTR_CONTAINER_REGISTRY_PASSWORD: ${{ secrets.VULTR_CONTAINER_REGISTRY_PASSWORD }}
 
-      - name: Deploy
-        if: steps.check_deployment.outputs.should_skip != 'true'
+      # website-25 is a special case, it has a multistage deployment
```
domdomegg, line 117: Rather than making website-25 a special case, I think I'd rather rename the script to something like `deploy:cd` in all apps. For most apps this just runs `npm run deploy:prod`, but for website-25 this runs `deploy:staging`. Then you wouldn't have to make any changes to this file.

apps/website-25/.env.production.template
```diff
@@ -0,0 +1,3 @@
+# This file will be picked up by the deploy script for production environment.
+NEXT_PUBLIC_GA_MEASUREMENT_ID=G-C5L967C9M1
+NEXT_PUBLIC_POSTHOG_KEY=phc_NVqRwH6xZ0MiP9I5a9tx3l9zsTWBvMI1YEbvbCOXDcI
```
Will-Howard, line 3: Should these be different to the staging keys?

Will-Howard (approved):
Did you consider using [branch-based deployment](https://shipyard.build/blog/branch-based-deploys/) by adding a separate `staging` branch? This is the method I'm more used to for static sites like this, some advantages I can think of:
- It's a more familiar workflow which makes it more convenient if we want to do frequent deploys (e.g. if we want to push most things straight to prod rather than queueing up several changes)
- It's easy to set up testing to block deployments (I'm sure this is possible this way too, but again the fact that it's the same process as deploying to staging makes it more convenient)
- If necessary you can push hotfixes to prod but not staging, which is sometimes helpful if you get yourself into a mess

But this way looks good too, and I can see the advantage of using GitHub Releases to make deployments more understandable

apps/website-25/.env.production.template
janraasch: Hi @Will-Howard, thanks for looking out 👍🏻. I have not created the actually staging `KEY`s (yet). I noted this down as a `To-do` in the `Rollout`-section of the `README`.

janraasch:
Hi @Will-Howard, thank you for the detailed feedback 🙏🏻

Why not [branch-based deployments](https://shipyard.build/blog/branch-based-deploys/)? 

* My main reason here was that a tag-based approach offers more flexibility: At some point we might have a `coursehub` app which we could then deploy to production *independently* from `website` via `coursehub-release*` tags.
* I see your point about being able to push hotfixes straight to production. Actually, this implementation supports this capability as well, since we don't check for the tag to be on `master`. Although this should be used kind of in emergencies only :)

Let's proceed with this `master` branch + `release` tags setup for now - we can always switch to a two-branch setup if we discover a necessity for it.

tarinrickett (approved):
LGTM too! Let's ship this tomorrow AM 🚢 🎉

README.md
```diff
@@ -180,7 +180,7 @@ Inside each package folder, the common files you'll find are:
   - `npm run lint`: Check for lint issues. Visual Studio Code should usually highlight these for you already.
   - `npm run build`: Build the application. This usually finds any type errors, which Visual Studio Code should usually highlight for you already.
   - `npm run postinstall`: Perform any extra steps to setup the application for development. Usually things like creating configuration files for local development. You usually don't need to run this manually, as it runs when you run `npm install`.
-  - `npm run deploy:prod`: Actually deploy the app into the production (real-world) environment. You usually don't need to run this manually, as it is run by CD tooling when you merge your changes into the master branch.
+  - `npm run deploy:cd`: Actually deploy the app into the production (real-world) environment. You usually don't need to run this manually, as it is run by CD tooling when you merge your changes into the master branch.
```
domdomegg, line 183:
```suggestion
  - `npm run deploy:cd`: Actually deploy the app, usually into the production (real-world) environment. You usually don't need to run this manually, as it is run by CD tooling when you merge your changes into the master branch.
```

apps/website-25/.env.production.template
domdomegg: Maybe it's fine to use the same keys across dev and staging, at least for now? If we do really want to keep these isolated it'd probably be fine to split them out later.

domdomegg (approved):
LGTM!

apps/website-25/.env.production.template
janraasch: Good call, @domdomegg! I'll update our `.env.local.template` to use `staging` keys.

Issue #344, opened 2025-02-13
[UX] Revisit SlideList designs to reduce complexity


tarinrickett:
SlideList designs are difficulty to implement, as the toggles are located inline with titles. We have also received Fearless Leader feedback that the toggles themselves are easy to miss.

Can we revisit this to introduce toggles on the left/right and keep the slidelist on a single horizontal plane?

See [#299](https://github.com/bluedotimpact/bluedot/issues/299)

tarinrickett:
Closing as duplicate: #388

PR #481, opened 2025-02-23
SlideList redesign

### LINKED ISSUE #347: Meaningfully test responsive behaviour of SlideList (post-refactor)
Description from #239:
> I wasn't able to easily update the tests to meaningfully check the desktop and mobile versions render as expected, because the logic in SlideList now depends a lot on CSS, and this doesn't appear to be applied properly in tests. Given the short deadline for the MVP I decided to go ahead without the tests, but I would like to come back and update them later. Also cc @janraasch for applying CSS in tests, I couldn't get it to work in the ~30 mins I spent trying

### LINKED ISSUE #388: Redesign of SlideList
See #344 . This must be completed first.

Will-Howard:
# Summary

This redesigns SlideList based on the new design we have, and the issues raised in #344 and various other places (#299, [Figma comment](https://www.figma.com/design/s4dNR4ELGKPbja6GkHLVJy?node-id=788-6484#1123521535)). We only have a desktop version of the new design so I went with what seemed sensible in other places, based on the goals of reducing complexity and making it look nicer.

Main changes:
- Moving the buttons from the header to the bottom of SlideList, and adding the progress bar between the buttons
  - I kept the old design of the progress bar because of the [problems I ran into here](https://github.com/bluedotimpact/bluedot/issues/388#issuecomment-2666831453) (more complex to decide which dot to highlight + looks uncentered if we keep the `peekAdjustment`). I can change this to the new design if desired
- Removing the Section-like logic from SlideList
- Adding an option to wrap onto multiple rows before fully collapsing to the slide view, this implements the [tablet designs](https://www.figma.com/design/s4dNR4ELGKPbja6GkHLVJy/Website-Laura's-Working-File?node-id=788-6710&t=JZPfjbZzIusYYofu-0) in the Figma and removes the need for handling this manually in some places
- Fix the styling of `CourseSection` so that the featured card and slide list are the same height without any hardcoded heights
- Removed the buttons from the mobile version (because they are swipeable). This was [discussed in Figma ](https://www.figma.com/design/s4dNR4ELGKPbja6GkHLVJy?node-id=788-6484#1123521535) and I decided to go with it
- Adds tests for the key features

## Issue

Fixes #388, Fixes #347

## Developer checklist
- [X] Added or updated Jest tests
- [X] Added or updated Storybook stories

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | Screenshot 2025-02-24 at 09 47 14 |
| 🖥️ | Screenshot 2025-02-24 at 09 47 48 |
| Tablet | Screenshot 2025-02-24 at 09 50 31 |
| | Screenshot 2025-02-24 at 09 50 01 |
| 📱  | Screenshot 2025-02-24 at 09 48 37 |

## Testing
```
$ npm run test:update
Cached:    5 cached, 7 total
  Time:    4.898s
```


apps/website-25/src/components/homepage/CommunitySection/TestimonialSection.tsx
```diff
@@ -30,38 +31,40 @@ const testimonials: Testimonial[] = [
...
-                src={testimonial.imageSrc}
-                alt={testimonial.name}
-                className="size-full object-cover"
-              />
-            </div>
-            <div className="testimonial__info">
-              <div className="testimonial__name font-semibold text-bluedot-black">
-                {testimonial.name}
+    <div className="testimonial-section">
+      <SectionHeading title="What our graduates say" className="testimonial-section__heading" />
```
tarinrickett, line 35: is this missing `titleLevel="h3"`?

apps/website-25/src/components/homepage/CommunitySection/index.tsx
```diff
@@ -7,7 +7,7 @@ import CommunityValuesSection from './CommunityValuesSection';
 const CommunitySection = () => {
   return (
     <Section className="community-section">
-      <div className="community-section__sub-sections flex flex-col">
+      <div className="community-section__sub-sections flex flex-col gap-spacing-y">
```
tarinrickett, line 10: is `gap-spacing-y` enough spacing? the current website has 1.5rem padding and 1.5rem margin on the elements directly: 
image

can you include a full page screenshot for verification? the current website got visual sign-off from the team

apps/website-25/src/components/homepage/CourseSection.tsx
```diff
@@ -13,26 +14,43 @@ const CourseCardWithTracking = withClickTracking(CourseCard, {
...
         <CourseCardWithTracking
-          trackingEventParams={{ course_title: featuredCourse.title, course_url: featuredCourse.href }}
+          trackingEventParams={{
+            course_title: featuredCourse.title,
+            course_url: featuredCourse.href,
+          }}
           {...featuredCourse}
           cardType="Featured"
-          className="h-full"
+          className="course-section__featured"
```
tarinrickett, line 31:
```suggestion
          className="course-section__featured lg:basis-[50%]"
```

apps/website-25/src/components/homepage/CourseSection.tsx
```diff
@@ -13,26 +14,43 @@ const CourseCardWithTracking = withClickTracking(CourseCard, {
...
-    >
-      {constants.COURSES.filter((course) => course !== featuredCourse).map((course) => (
-        <CourseCardWithTracking trackingEventParams={{ course_title: course.title, course_url: course.href }} key={course.title} {...course} />
-      ))}
-    </SlideList>
+        <SlideList
+          maxItemsPerSlide={4}
+          maxRows={1}
+          minItemWidth={300}
+          className="course-section__carousel"
```
tarinrickett, line 37:
```suggestion
          className="course-section__carousel lg:basis-[50%]"
```

apps/website-25/src/components/homepage/CourseSection.tsx
```diff
@@ -13,26 +14,43 @@ const CourseCardWithTracking = withClickTracking(CourseCard, {
...
+      className="course-section"
       title="Our courses"
       subtitle="Our courses combine self-paced study with guided discussions, deepening your learning while fostering connections in the field"
       subtitleLevel="p"
-      className="course-section section-body"
-      maxItemsPerSlide={2}
-      minItemWidth={300}
-      featuredSlot={(
+    >
+      <div className="course-section__content flex flex-col lg:flex-row lg:[&>*]:[flex-basis:50%] gap-space-between items-stretch">
```
tarinrickett, line 23: i prefer to apply styles as directly to the element as possible. i think it improves discoverability and maintainability
```suggestion
      <div className="course-section__content flex flex-col lg:flex-row gap-space-between items-stretch">
```

tarinrickett: P.S. for future ref, there are also tailwind selectors for children: https://tailwindcss.com/docs/hover-focus-and-other-states#child-selectors
but i do agree with their preceding note:
> While it's generally preferable to put utility classes directly on child elements, you can use the * variant in situations where you need to style direct children that you don’t have control over:

apps/website-25/src/lib/withClickTracking.tsx
```diff
@@ -45,7 +45,7 @@ export const withClickTracking = <P extends object>(
     };
 
     // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
-    return <div onClick={handleClick}><WrappedComponent {...componentProps as P} /></div>;
+    return <span onClick={handleClick}><WrappedComponent {...componentProps as P} /></span>;
```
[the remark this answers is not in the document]
tarinrickett: i'm guessing that's due to the default positioning. `<div>` is a block-level element and `<span>` is inline. i think it makes sense to keep `span` here! 👍

libraries/ui/src/Section.tsx
```diff
@@ -34,13 +33,12 @@ export const SectionHeading: React.FC<BaseProps> = ({
           <SubtitleTag className="section-heading__subtitle text-bluedot-darker text-md">{subtitle}</SubtitleTag>
         )}
       </div>
-      {rightNode}
```
tarinrickett, line 37: i agree with the design decision to remove rightNode, thanks for taking that on.

libraries/ui/src/SlideList.test.tsx
```diff
@@ -1,22 +1,204 @@
...
 
+/**
+ * Mocks the ResizeObserver API for testing purposes.
+ * Based on this article: https://greenonsoftware.com/articles/testing/testing-and-mocking-resize-observer-in-java-script/
+ *
+ * @returns {Object} An object containing:
+ *  - resizeCallback: A function to manually trigger resize events.
+ *  - observerInitialized: A promise that resolves when the mock observer is set up.
+ */
+function mockResizeObserver() {
```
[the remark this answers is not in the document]
tarinrickett: nice. if this becomes a repeat pattern, we can abstract away into a util !

libraries/ui/src/SlideList.test.tsx
```diff
@@ -1,22 +1,204 @@
...
         <div>Slide 1</div>
         <div>Slide 2</div>
         <div>Slide 3</div>
       </SlideList>,
     );
+    expect(screen.queryAllByLabelText(/(Previous|Next) slide/)).toHaveLength(0);
+
+    vi.clearAllMocks();
+    // TODO: clearAllMocks() doesn't actually clear this
+    vi.spyOn(deviceDetect, 'isMobile', 'get').mockReturnValue(false);
```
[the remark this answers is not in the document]
tarinrickett: thanks Will and thanks for documenting so we can track as needed!!

libraries/ui/src/SlideList.tsx
```diff
@@ -51,8 +47,8 @@ export const SlideList: React.FC<SlideListProps> = ({
     return cleanup;
   }, []);
 
-  const itemsFit = Math.max(1, Math.floor((measuredContainerWidth ?? 800) / minItemWidth));
-  const itemsPerSlide = Math.max(1, Math.min(itemsFit, maxItemsPerSlide));
+  const itemsFitPerRow = Math.max(1, Math.floor((measuredContainerWidth ?? 800) / minItemWidth));
```
tarinrickett, line 50: where does 800 come from? would be good to have a comment explaining this

libraries/ui/src/default-config/tailwind.css
```diff
@@ -110,7 +110,7 @@
 
 /* Class for titled sections in the body of the page */
 @utility section-body {
-    @apply w-full max-w-max-width mx-auto px-spacing-x pt-6 pb-10 border-b border-color-divider overflow-hidden flex flex-col;
```
tarinrickett, line 113: i think we had to add this as there was concerns from product that visually the sections were not aligned as intended. you are right that theoretically it should be `pb-spacing-y` but `pb-10` is a workaround until we can dig deeper into the design approach.

it hurts my eng brain too to have to add this, but let's keep it as is. happy to add a comment if you'd like to explain this above

tarinrickett (commented):
big refactor that looks so good!!!

very glad to see SlideList and Section decoupled. i think will be much more maintainable moving future. appreciate all the iterative thinking on this!

apps/website-25/src/components/homepage/CommunitySection/TestimonialSection.tsx
Will-Howard: Good catch 🙏

apps/website-25/src/components/homepage/CommunitySection/index.tsx
Will-Howard: Here's a full page screenshot before and after:
Screenshot 2025-03-01 at 16 51 06
Screenshot 2025-03-01 at 16 51 38

I think you're right it does look a bit cramped on the cramped side now, let me know whether you want me to change it back

apps/website-25/src/components/homepage/CourseSection.tsx
Will-Howard: This doesn't quite work unfortunately because of the extra `<span>` wrapping. I could add the ability to give the span a className, but I think the current `lg:[&>*]:[flex-basis:50%]` is easier to follow than that

apps/website-25/src/components/homepage/CourseSection.tsx
Will-Howard: This doesn't quite work unfortunately because of the extra `<span>` wrapping. I could add the ability to give the span a className, but I think the current `lg:[&>*]:[flex-basis:50%]` is easier to follow than that.

I think the real answer here is refactoring `withClickTracking` so it doesn't have this problem, probably by making it a hook that components use to implement their own tracking rather than something that is applied externally

libraries/ui/src/SlideList.tsx
Will-Howard: I've added a comment 👍:
```typescript
  // On the first render, we won't have a `measuredContainerWidth`. Fall back
  // to 1400px as a value close to the max column width, so that the first render will tend to
  // start of un-collapsed (to avoid layout shift)
  const itemsFitPerRow = Math.max(1, Math.floor((measuredContainerWidth ?? 1400) / minItemWidth));
```
I also made the value bigger to achieve the intended goal better

libraries/ui/src/default-config/tailwind.css
Will-Howard: 👍 I've changed it back

tarinrickett (approved):
LGTM

PR #483, opened 2025-02-24
Update record-attendance.tsx

anglilian:
We updated this param in Course Runner to groupClassId in line with operation sensible names.


tarinrickett (approved):
oh man wonder if we need more auditing in the github repo for sensible names. thanks for identifying + fixing @anglilian  !!

PR #475, opened 2025-02-20
Add 404 page

Will-Howard:
# Summary

https://github.com/user-attachments/assets/db280bf9-923a-437b-9174-15079a06c8f4

## Issue

Fixes #389

## Developer checklist
- [X] Used [BEM naming convention](https://getbem.com/naming/) for classNames
- [X] Added or updated Jest tests
- [X] Added or updated Storybook stories

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | Screenshot 2025-02-26 at 09 37 08 |
| 🖥️ | Screenshot 2025-02-26 at 09 37 47 |
| 📱  | Screenshot 2025-02-26 at 09 37 55 |

## Testing
```
$ npm run test:update
 Tasks:    7 successful, 7 total
Cached:    6 cached, 7 total
  Time:    4.424s
```

PR #494, opened 2025-02-26
website-proxy: 301 redirect `www.bluedot.org` to `bluedot.org`

janraasch:
# Summary

Add 301 permanent redirect from `www.bluedot.org` to `bluedot.org`

## Issue

Fixes #487

## Testing

I've verified my solution through Claude and cross-referenced it with [this gist](https://gist.github.com/EmranAhmed/946a74c0331832040844) 😇, though I couldn't test the `nginx.template.conf` locally.

## Screenshot

📸 **No visual changes**

## Testing
```
$ npm run test:update
 Tasks:    9 successful, 9 total
Cached:    4 cached, 9 total
  Time:    5.404s 
```

PR #496, opened 2025-02-26
Add routing for 404 page to website-proxy

### LINKED ISSUE #391: Update routing to include 404 page

Will-Howard:
**This should not merged before #475 **

# Summary

The location block should make it so that:
- If a route isn't explicitly directed to the new site (https://website-25-production.k8s.bluedot.org), it tries directing it to the old site (https://45.76.132.116)
  - If this returns a 404, it redirects to the same route on the new site. This may or may not return a 404, if it does it will display the page added in #475
- If a route is directed to the new site and still returns a 404, it should do an internal redirect to the same page and then return this

I haven't tested this yet, @janraasch is there a good way of testing this nginx config that you have been using? If not I will test up a quick test server and try it out 

## Issue

Fixes #391 

## Developer checklist
- [X] [N/A] Used [BEM naming convention](https://getbem.com/naming/) for classNames
- [X] [N/A] Added or updated Storybook stories

### DIFF
--- a/apps/website-proxy/src/nginx.template.conf
+++ b/apps/website-proxy/src/nginx.template.conf
@@ -49,6 +49,12 @@ http {
             proxy_pass https://website-25-production.k8s.bluedot.org;
         }
 
+        # Redirect 404 errors to the same route on the new website
+        error_page 404 = @website_25_404;
+        location @website_25_404 {
+            proxy_pass https://website-25-production.k8s.bluedot.org$request_uri;
+        }
+
         # Default all other traffic to old site
         location / {
             proxy_ssl_name bluedot.org;


janraasch:
Hi @Will-Howard, [check out this comment](https://github.com/bluedotimpact/bluedot/pull/494#discussion_r1971290820) by @domdomegg on testing nginx config locally.

apps/website-proxy/src/nginx.template.conf
```diff
@@ -49,6 +49,12 @@ http {
             proxy_pass https://website-25-production.k8s.bluedot.org;
         }
 
+        # Redirect 404 errors to the same route on the new website
```
janraasch, line 52: Hi @Will-Howard, I think this is solid. 

An idea would be to move the `error_page` call into `location / {`:

```nginx
        location / {
            proxy_ssl_name bluedot.org;
            proxy_set_header Host bluedot.org;
            proxy_pass https://45.76.132.116;
            error_page 404 = @website_25_404;
        }

        location @website_25_404 {
            proxy_pass https://website-25-production.k8s.bluedot.org;
        }
```

Looks a little bit cleaner maybe? What do you think?

apps/website-proxy/src/nginx.template.conf
```diff
@@ -49,6 +49,12 @@ http {
...
+        error_page 404 = @website_25_404;
+        location @website_25_404 {
+            proxy_pass https://website-25-production.k8s.bluedot.org$request_uri;
+        }
+
         # Default all other traffic to old site
```
janraasch, line 58: Quick note on testing this *locally*:

1. Replace `docker run` line in `libraries/docker-scripts/src/start.sh` with `docker run --rm --publish 8001:8080 $IMAGE_ID` so we can run `website-25` on `8000`.
2. Replace `https://website-25-production.k8s.bluedot.org` with `http://host.docker.internal:8000` to use your local version of `website-25` instead of `website-25-production`.
3. Start `website-proxy` via `npm start`.
4. Start `website-25` via `npm start`.
5. Use `curl` to trigger a request with the correct `Host`, e.g. `curl -vv localhost:8001/does-not-exist --header 'Host: bluedot.org'`
6. Check. the logs :)

janraasch: Great call, @tarinrickett! 

@Will-Howard: Would you mind adding (sth like 😇) this to the `website-proxy/README.md` once you've gotten a chance to play around with it?

Will-Howard:
@janraasch 
> bluedot.org returns HTTP Status 200 instead of 404 for non-existent pages

Ah good catch! This is the sort of thing that could have easily made it to production and been very confusing. I'll get this working locally and then wait on the wordpress site being updated

domdomegg:
Ah I don't know how easy it will be to update the old Wordpress site - this is managed by an agency and not something we really have much control over. Is there a way to check the response content and route based on that? Otherwise we can try to coordinate with the agency.

Will-Howard:
> Is there a way to check the response content and route based on that?

It may be possible in principle but definitely not while using the nginx image that we are using. There is apparently an extended version of nginx called [OpenResty](https://en.wikipedia.org/wiki/OpenResty) that can do things based on the response body, but even with that changing the status is hard because the headers are sent before the body is received.

I think it would be best to coordinate with the agency, or just decide this isn't worth having if you're planning to migrate the whole site over relatively soon anyway.

Anyway apart from that @janraasch I tried the instructions you gave [here](https://github.com/bluedotimpact/bluedot/pull/496#discussion_r1971679398) and it worked perfectly, thanks for that!

apps/website-proxy/src/nginx.template.conf
Will-Howard: Yep good shout, I've changed it to this

Issue #251, opened 2025-02-10
Add semi bold Roobert woff2-file


janraasch:
**Context**:
We use `font-weight 650` quite a bit on the designs, but we currently do not have a matching `@font-face` declaration in `libraries/ui/src/shared.css`.

**To-dos**:
* Add `/fonts/Roobert-SemiBold.woff2`-file to `website-25/public/fonts` (- Do we need to add this file to all of our `apps`?)
* Add `@font-face` declaration for `font-family: "Roobert";` & `font-weight: 650;` to `libraries/ui/src/shared.css`

janraasch:
⚠️ Note to self: Add `SemiBold.woff2` to `public` folder of other apps as well, since they rely on our common `tailwind.css` config.

janraasch:
Hi @tarinrickett, could you please provide the `Roobert-SemiBold.woff2` file (for `font-weight: 650`)?

I assume BlueDot purchased the `Roobert` font family from https://displaay.net/typeface/roobert-collection/roobert/. 

CC: @domdomegg (adding you since you've preivously added these `fonts` to our other apps).

domdomegg:
@janraasch I think you can find all the fonts here: https://drive.google.com/drive/folders/1DSPImBM-ykbegC4UTYs0Rcf63d5DZUxW?usp=drive_link

janraasch:
> (there maybe is a separate question to ask about how necessary this font weight is, in terms of trade-offs against extra complexity + page load time. but I will leave you/design to make this decision)

Good point, @domdomegg 👍🏻.

💁🏻 I did a quick search for both `font-[650]` & `font-semibold` (which is short for `font-600`) in our codebase & currently we do use this for a handful of base elements & ui components.

@tarinrickett: Perhaps we could revisit the necessity of `SemiBold` with design/UX (now - *after the MVP launch*)? We could evaluate whether the visual benefits outweigh the additional complexity.

PR #502, opened 2025-02-27
Remove legacy app `bubble-proxy`

janraasch:
# Summary

Remove unused legacy app `bubble-proxy` to clean up our codebase 🧹 😁.

## Issue

Discovered unused apps while working on #251.

## Description

Removing will:

* Ease onboarding for new developers
* *Somewhat* speed up CI/CD pipelines
* Reduce maintenance overhead

We'll keep `backennd` around for later reference, see https://github.com/bluedotimpact/bluedot/pull/502#issuecomment-2687565058.

## Testing
```
 Tasks:    8 successful, 8 total
Cached:    7 cached, 8 total
  Time:    3.575s 
```


domdomegg:
`bubble-proxy` can definitely go. We ended up not replacing [CoAlias](https://coalias.com/) with it because CoAlias fixed the thing that was bugging us.

`backend` is maybe useful in so far as it has a bunch of code that is potentially relevant for if we want to build a custom course hub (or any more advanced server really). I guess we could pull it out of the git history again (but doing this is a pain). I'd be tempted to leave it there for a couple months and then delete it once we figure out what we're doing with course hub. If we do want to delete it now, could we also delete the backend-contract package and the bits in frontend-example that link to the backend?

janraasch:
Thank you for the detailed reply, @domdomegg ❤️. 

Let's keep `backend` around (for now 😇). I added a note to the top of the README as well to make sure it's clear that this is noop & kept for later reference. 

I'm looking forward to revisit this once the need for a backend arises 🔮☺️.

domdomegg (approved):

PR #514, opened 2025-03-12
Fix tab navigation in Nav

### LINKED ISSUE #286: [feedback] website: keyboard navigation is in wrong order
When using tab to navigate the page:
- the top navigation does not navigate as standard, e.g. after 'explore' is opened, the next step is not into this menu
- ~tab navigation in the course card scrolling bit doesn't work~ fixed in #386 
- ~it feels odd that the navigation buttons for the course scrolling bit come before the featured course~ will be covered by #388 
- ~keyboard navigation enters closed FAQs without opening them (either it shouldn't navigate into them at all, or should open them as it navigates into them)~ fixed in #338! 🎉

Will-Howard:
# Summary

This makes it so that after you open the Explore menu, you can tab into it immediately. There should be no significant visual change along with this (I have just aligned the desktop explore section with the Explore button rather than centering it in the page).

https://github.com/user-attachments/assets/8ccf9006-8120-453c-9ea9-074caffaf457

## Issue

Fixes #286 

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | There is a bug in the storybook at the moment which causes some tailwind breakpoint classes to be applied in the wrong order (e.g. flex overrides lg:hidden). This was already there, but I haven't been able to fix it so the story is still broken: Screenshot 2025-03-12 at 10 40 08 |
| 🖥️ | Screenshot 2025-03-12 at 10 43 05 |
| 📱  | Screenshot 2025-03-12 at 10 43 27 |

## Testing
```
$ npm run test:update
 Tasks:    10 successful, 10 total
Cached:    1 cached, 10 total
  Time:    11.207s
```

### DIFF
--- a/apps/website-25/src/components/Nav.stories.tsx
+++ b/apps/website-25/src/components/Nav.stories.tsx
@@ -39,8 +39,5 @@ export const Default: Story = {
 export const Customized: Story = {
   args: {
     logo: imgSrc,
-    children: [
-      <><a href="#">Support Us</a><a href="#">About</a><a href="#">Join us</a><a href="#">Blog</a></>,
-    ],
   },
 };
--- a/apps/website-25/src/components/Nav.tsx
+++ b/apps/website-25/src/components/Nav.tsx
@@ -48,13 +48,13 @@ const ExploreSection: React.FC<{
   >
     <div
       className={clsx(
-        'nav-explore-section___dropdown-content flex flex-col gap-[14px] w-fit overflow-hidden text-pretty',
+        'nav-explore-section___dropdown-content flex flex-col gap-[14px] text-pretty px-6',
         innerClassName,
       )}
     >
       <h3 className="nav-explore-section__dropdown-title font-bold">Our courses</h3>
       {courses?.map((course) => (
-        <a key={course.href} href={course.href} className="nav-explore-section__dropdown-link">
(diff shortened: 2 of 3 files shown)


apps/website-25/src/components/Nav.stories.tsx
```diff
@@ -39,8 +39,5 @@ export const Default: Story = {
 export const Customized: Story = {
   args: {
     logo: imgSrc,
-    children: [
-      <><a href="#">Support Us</a><a href="#">About</a><a href="#">Join us</a><a href="#">Blog</a></>,
```
Will-Howard, line 43: This was already unused since refactoring the links into the Nav component itself

domdomegg (commented):
This maybe needs more testing?

On macOS Chrome (mobile, tablet, desktop breakpoints) this puts the explore courses part in the right order. The menu appears to work fine on Android Chrome (mobile breakpoint), but I haven't tested accessibility features.

However, on the tablet and mobile breakpoints it's possible to navigate through the menu without it actually being open - which is confusing. Either it should skip over these, or it should open the menu as you navigate through them. (To reproduce: set the screen width to mobile, and then hit tab through the items from the start - you'll see it selects items in the menu that aren't visible).

On Safari with voiceover it seems to act very strangely. It seems to jump between links in unpredictable ways, often selecting things that aren't visible on the page.

domdomegg:
Closing this as stale for now - I think this ticket is not super high priority for us right now. Reopen if this is wrong!

PR #520, opened 2025-03-17
course-demos: 2018 to present

domdomegg:


domdomegg:
A pretty basic UI for now, but it works to show how these models have advanced

image

PR #531, opened 2025-03-28
website-25: Add new lander page

domdomegg:
# Summary

Added new 'classic' style landing page

## Developer checklist

## Screenshot

image
image


apps/website-25/package.json
```diff
@@ -36,6 +36,7 @@
     "react-click-away-listener": "^2.3.0",
     "react-device-detect": "^2.2.3",
     "react-dom": "^18.2.0",
+    "react-icons": "^5.5.0",
```
domdomegg, line 39: This is not actually a new dep - it's pulled in from ui already. But eslint needs this not to complain (semi rightfully so, because if we uninstall libraries/ui it would break which would be unexpected).

domdomegg:
Discussed with Tarin: okay to merge this for now into staging and Tarin will post-review.

PR #532, opened 2025-03-28
Refactor landing

domdomegg:
- Move stuff into reusable composable components so it's less chaos when others edit the copy in GitHub later
- Add another variant


apps/website-25/src/components/lander/NewNav.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
```
tarinrickett, line 1: can we just use Nav?

PR #534, opened 2025-03-28
Update Hero background

tarinrickett:
image

Mobile still broken

Screenshot 2025-03-28 at 16 06 43


domdomegg (commented):
LGTM

Possibly follow-up (not at all a priority for now): Maybe we could get an svg version so it looks pretty on high-res monitors? Canva can do svg export just as easily as png.

PR #536, opened 2025-03-28
lander: Replace NewNav with Nav

domdomegg:
image

Why draft for now: This is kinda messy - the navbar is a different width to the rest of the site, it's taller (so less above-fold content), and it gives users lots of options to get lost in other places on the site. Hardcoding the height of the navbar in also feels like a bit of a hack. Will return to this as a P1 once we sort the other more obvious bugs/problems. (also I think the navbar might be going completely so maybe wait until we see what's happening to fix this)


tarinrickett:
👍

tarinrickett:
i think we can close this now with https://github.com/bluedotimpact/bluedot/pull/674

domdomegg:
Going to merge this given:
- some of the hackiness I was irritated by before is resolved
- while the pages have some traffic, it's not lots - and we're trying to reduce it

PR #537, opened 2025-03-28
Add all variants

tarinrickett:
Subtitle doesn't look great with all variants. Need to revisit styles.

Screenshot 2025-03-28 at 16 24 02
Screenshot 2025-03-28 at 16 23 54
Screenshot 2025-03-28 at 16 23 47
Screenshot 2025-03-28 at 16 23 39
Screenshot 2025-03-28 at 16 23 32


domdomegg (commented):
LGTM

PR #558, opened 2025-03-31
devex: Fix package.json for eslint-config linking to eslint-plugin-custom

domdomegg:
# Summary

devex: Fix package.json for eslint-config linking to eslint-plugin-custom

This was causing mysterious npm errors when installing packages, but only sometimes, such as:

```
npm error Cannot read properties of undefined (reading 'extraneous')
```

The package name for our custom eslint plugin was wrong in eslint-config's package.json. I figured this out by editing the NPM source code to print out more details of the error, and it was consistently when resolving the @bluedot/eslint-config package. I'm fairly certain this was the cause of the problem, but I can't be certain - we'll see if the bug reappears.

### DIFF
--- a/libraries/eslint-config/package.json
+++ b/libraries/eslint-config/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "main": "src/index.js",
   "devDependencies": {
-    "@bluedot/eslint-plugin-bluedot-custom": "*",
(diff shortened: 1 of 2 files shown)


domdomegg:
Have opened: https://github.com/npm/cli/issues/8199

PR #600, opened 2025-04-04
Initial API config for Unit

tarinrickett:
# Summary

Initial API config for Unit.

See: http://bluedot.org/api/units

### DIFF
--- a/apps/website-25/src/pages/api/units.ts
+++ b/apps/website-25/src/pages/api/units.ts
@@ -1,0 +1,25 @@
+import { z } from 'zod';
+import db from '../../lib/api/db';
+import { makeApiRoute } from '../../lib/api/makeApiRoute';
+import { Unit, unitTable } from '../../lib/api/db/tables';
+
+export type GetUnitResponse = {
+  type: 'success',
+  units: Unit[],
+};
+
+export default makeApiRoute({
+  requireAuth: false,
+  responseBody: z.object({
+    type: z.literal('success'),
+    units: z.array(z.any()),
+  }),
+}, async () => {
+  const allUnits = await db.scan(unitTable);
+  // how do we use params arg to filter by course and/or unit for more performant fetching?
+
+  return {
+    type: 'success' as const,
+    units: allUnits,
+  };
+});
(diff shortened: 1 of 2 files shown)


apps/website-25/src/pages/api/units.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default makeApiRoute({
+  requireAuth: false,
+  responseBody: z.object({
+    type: z.literal('success'),
+    units: z.array(z.any()),
+  }),
+}, async () => {
+  const allUnits = await db.scan(unitTable);
+  // how do we use params arg to filter by course and/or unit for more performant fetching?
```
domdomegg, line 19: `filterByFormula` (click into db.scan to see the options)

OR

fetch the course, then find the linked records and do a `db.get` on them

domdomegg (approved):

apps/website-25/src/pages/api/units.ts
tarinrickett: thanks Adam! we should be able to use filter by formula to target by field ID for consistency as well, e.g. `filterByFormula=fldLmQZ0ISTr7xQUE+%3D+%22Future+of+AI%22`

i'll test this out 👍

domdomegg: I think the syntax is something like `{fldLmQZ0ISTr7xQUE} = Future of AI`

For even more robustness it might be good to filter by id values too - e.g. find the linked field column and do something like `FIND({fldLmQZ0ISTr7xQUE}, "rec1234") > -1`. Where `fldLmQZ0ISTr7xQUE` is the linked column field. (I'm not sure this is quite the right syntax, but something like this... I'm also not 100% certain this is possibleb but I think it should be?)

Also useful: https://support.airtable.com/docs/formula-field-reference

domdomegg: Actually from the FAQs I notice this very sad point:

```
Can I use field IDs in my filterByFormula API call?

No. You can only use field names. We recommend using field descriptions to call out the implications of changing any field names being used in API calls.
```

I will contact Airtable Enterprise support and ask them if this could be changed as a feature request.

PR #619, opened 2025-04-08
website: tidy up unit routing, fix unit API errors

domdomegg:
# Summary

## Issue

Fixes # (issue)

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📕 |  |
| 🖥️ |  |
| 📱  |  |

## Testing
```
$ npm run test:update
```

### DIFF
--- a/apps/website-25/src/components/courses/SideBar.tsx
+++ b/apps/website-25/src/components/courses/SideBar.tsx
@@ -1,15 +1,16 @@
 import React from 'react';
 import { Unit } from '../../lib/api/db/tables';
+import { ROUTES } from '../../lib/routes';
 
 type SideBarProps = {
   // Required
-  courseId: string;
+  courseSlug: string;
   units: Unit[];
   currentUnitNumber: number;
 };
 
 const SideBar: React.FC<SideBarProps> = ({
-  courseId,
+  courseSlug,
   units,
   currentUnitNumber,
 }) => {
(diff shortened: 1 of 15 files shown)


apps/website-25/src/components/courses/SideBar.tsx
```diff
@@ -21,12 +22,13 @@ const SideBar: React.FC<SideBarProps> = ({
...
         {units.map((unit) => (
           <div
+            key={unit.id}
             className={`sidebar__unit border ${isCurrentUnit(unit) ? 'sidebar__unit--active border-color-primary' : 'border-color-divider'} rounded-radius-md p-4 flex flex-col gap-2`}
           >
             <div className="sidebar__unit-item flex flex-col gap-2">
               <p className="sidebar__title uppercase text-size-xs font-bold">Unit {unit.unitNumber}</p>
-              <a href={`/courses/${courseId}/unit/${unit.unitNumber}`} className="sidebar__description subtitle-sm">{unit.title}</a>
-              <p className="sidebar__description text-sm">{unit.description}</p>
+              <a href={ROUTES.makeCoursePageRoute(courseSlug, undefined, Number(unit.unitNumber)).url} className="sidebar__description subtitle-sm">{unit.title}</a>
```
tarinrickett, line 30: this morning we aligned on URLs to be maintained from a single SOT in Airtable, so with that in mind, we should be able to pass around `{unit.url}` or equivalent here instead of constructing the path ourself.
```suggestion
              <a href={unit.url} className="sidebar__description subtitle-sm">{unit.title}</a>
```

tarinrickett: i worry maintaining `courseSlug` is not enough, as we would have to duplicate route creation in stuff like emails (e.g., if we want to link to specific units in emails reminders). what are your thoughts?

PR #627, opened 2025-04-09
website: Add Embed and Callout components

domdomegg:
# Summary

Add components:
- `Embed` for iframe content, like YouTube videos, demos, etc.
- `Callout` for bonus content

## Issue

Part of #616, may also help with #585

## Developer checklist

## Screenshot

image

### DIFF
--- a/apps/website-25/src/components/courses/Embed.stories.tsx
+++ b/apps/website-25/src/components/courses/Embed.stories.tsx
@@ -1,0 +1,51 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import Embed from './Embed';
+
+const meta = {
+  title: 'website/Embed',
+  component: Embed,
+  // This component will have an automatically generated Autodocs entry: https://storybook.js.org/docs/writing-docs/autodocs
+  tags: ['autodocs'],
+  parameters: {
+    // More on how to position stories at: https://storybook.js.org/docs/configure/story-layout
+    layout: 'fullscreen',
+  },
+  args: {},
+} satisfies Meta<typeof Embed>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const YouTubeEmbed: Story = {
+  name: 'YouTube Embed',
+  args: {
+    url: 'https://www.youtube.com/embed/dQw4w9WgXcQ',
+  },
+};
+
+export const DemoEmbed: Story = {
+  args: {
+    url: 'https://course-demos.k8s.bluedot.org/generate-react-component',
+  },
+};
+
+export const SunoEmbed: Story = {
+  args: {
+    url: 'https://suno.com/embed/e02aebab-1fe8-4b64-b849-7393e01eeafd',
+    className: '!h-40',
+  },
+};
+
+export const WebsiteEmbed: Story = {
+  args: {
+    url: 'https://example.com',
+  },
+};
+
+export const CustomStyleEmbed: Story = {
+  args: {
+    url: 'https://example.com',
+    className: 'invert !w-1/2 my-24 mx-auto rotate-30',
+  },
+};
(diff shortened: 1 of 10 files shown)


apps/website-25/src/components/courses/MarkdownExtendedRenderer.tsx
```diff
@@ -2,12 +2,20 @@ import React, { useEffect } from 'react';
...
+import Embed from './Embed';
+// eslint-disable-next-line import/no-cycle
+import Callout from './Callout';
 
 export interface MarkdownRendererProps {
   children?: string;
 }
 
-export const SUPPORTED_COMPONENTS = { Greeting };
+// This must be a function, rather than a constant, to avoid dependency cycles
```
tarinrickett, line 13: does Callout need full ExtendedRenderer? can we just support basic ReactMarkdown in Callouts to reduce complexity here?

we can always revisit once we have usecases that require embed, etc. in callouts

apps/website-25/src/components/courses/Embed.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  parameters: {
+    // More on how to position stories at: https://storybook.js.org/docs/configure/story-layout
+    layout: 'fullscreen',
+  },
+  args: {},
+} satisfies Meta<typeof Embed>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
```
tarinrickett, line 19: could you include storybook screenshot/screen recording? would be interested to see how these different embeds are all rendering. also, is it already adapting on mobile screens?

(we can also do these as fast follows if youd like, but want to make sure we have it documented if so)

apps/website-25/src/components/homepage/__snapshots__/FAQSection.test.tsx.snap
```diff
@@ -74,7 +74,7 @@ exports[`FAQSection > renders default as expected 1`] = `
         </div>
       </details>
       <details
-        class="collapsible max-w-max-width border-b border-color-divider py-4 last:border-none group marker:hidden [&_summary::-webkit-details-marker]:hidden faq-section__item"
+        class="collapsible max-w-max-width border-b border-color-divider py-4 last:border-b-0 group marker:hidden [&_summary::-webkit-details-marker]:hidden faq-section__item"
```
tarinrickett, line 77: can you include FAQ screenshot as a sanity check?

tarinrickett: i hope custom styling of Collapsible in Callout won't lead to any conflicts in the future. no action needed now, i think this solution works, but we may want to revisit and apply absolute bare-minimum styling to Collapsible and port this border style over to the specific FAQSection instance

tarinrickett (approved):

apps/website-25/src/components/courses/MarkdownExtendedRenderer.tsx
tarinrickett: we'll run into the same problem with MultipleChoice and FreeTextResponse, which also handle markdown

domdomegg: It works the same - we can swap them in/out pretty easily.

And yes, the current future of AI course uses videos within callouts already.

apps/website-25/src/components/courses/Embed.stories.tsx
domdomegg: https://github.com/user-attachments/assets/fceecc04-05fc-435e-a039-c3190f7ea56c

apps/website-25/src/components/courses/MarkdownExtendedRenderer.tsx
```diff
@@ -2,12 +2,20 @@ import React, { useEffect } from 'react';
 import { evaluate } from '@mdx-js/mdx';
 import { MDXContent } from 'mdx/types';
 import Greeting from './Greeting';
+import Embed from './Embed';
+// eslint-disable-next-line import/no-cycle
```
domdomegg, line 6: Also I realised I managed to get Callout working without a dependency cycle, but then didn't update the comment in MarkdownExtendedRenderer. So there isn't a circular dependency any more :tada: (but things are set up so that it can be stable if this is needed in future)

Issue #564, opened 2025-04-01
Free-answer response component


tarinrickett:
<img width="738" alt="Image" src="https://github.com/user-attachments/assets/873270a6-ec18-4652-b477-d8735e77d545" />

PR #629, opened 2025-04-09
[DRAFT] website: Add MDX editor component

domdomegg:
# Summary

Adds an MDX editor component

In draft state because I'm not sure I like this library. The editing experience feels a bit clunky (especially trying to delete blocks like code blocks / JSX elements) and it doesn't support previewing JSX components by default I think?

## Issue

Intended to make editing unit content and blog posts easier.

Possibly related to #564 if we use the same editor

## Developer checklist

## Screenshot

https://github.com/user-attachments/assets/4cc3c674-54e1-4e02-bec6-5b022bb1b24c


domdomegg:
Superceded by #665, which I think is just better

PR #631, opened 2025-04-09
Exercise API get and put handling

### LINKED ISSUE #562: Store & update exercise responses
See [[WIP] Exercise API get and put handling #631](https://github.com/bluedotimpact/bluedot/pull/631)

tarinrickett:
# Summary

Updating Exercise, FreeTextResponse, and MultipleChoice to handle saving, updating, and rendering saved user responses

## Issue
Fixes #562 

## Description
- New APIs:
  - GET /api/courses/exercises/${exerciseId} – does not require auth, returns exercise details
  - GET /api/courses/exercises/${exerciseId}/response – requires auth, returns saved exercise responses
  - PUT /api/courses/exercises/${exerciseId}/response – requires auth, upsert exercise with new submitted state
- Exercise handles API methods and passes a callback to FreeTextResponse and MultipleChoice
- Use react-hook-form to help handle form updates in FreeTextResponse and MultipleChoice

## Screen Recordings

https://github.com/user-attachments/assets/5d3748d7-ddb5-4fc7-a408-0e54357ae378

Screenshot 2025-04-22 at 18 15 30

Screenshot 2025-04-22 at 18 15 26

Screenshot 2025-04-22 at 18 15 37

https://github.com/user-attachments/assets/041ef8a3-0e75-49e7-ab74-dd79424d77aa

Screenshot 2025-04-22 at 18 14 54

Screenshot 2025-04-22 at 18 15 04

### DIFF
--- a/apps/website-25/src/components/courses/exercises/Exercise.tsx
+++ b/apps/website-25/src/components/courses/exercises/Exercise.tsx
@@ -1,8 +1,10 @@
 import React from 'react';
 import useAxios from 'axios-hooks';
+import { useAuthStore } from '@bluedot/ui';
 import FreeTextResponse from './FreeTextResponse';
 import MultipleChoice from './MultipleChoice';
-import { GetExerciseResponse } from '../../../pages/api/courses/exercises/[exerciseId]';
+import { GetExercise } from '../../../pages/api/courses/exercises/[exerciseId]';
+import { GetExerciseResponse } from '../../../pages/api/courses/exercises/[exerciseId]/response';
 
 type ExerciseProps = {
   // Required
@@ -12,23 +14,56 @@ type ExerciseProps = {
 const Exercise: React.FC<ExerciseProps> = ({
   exerciseId,
 }) => {
-  const [{ data }] = useAxios<GetExerciseResponse>({
+  const [{ data: exerciseData }] = useAxios<GetExercise>({
     method: 'get',
     url: `/api/courses/exercises/${exerciseId}`,
   });
 
-  if (!data || !data.exercise) {
+  if (!exerciseData || !exerciseData.exercise) {
     console.error('Exercise not found');
     return null;
   }
 
+  const auth = useAuthStore((s) => s.auth);
+
+  const [{ data: responseData }, executeResponse] = useAxios<GetExerciseResponse>({
+    method: 'get',
+    url: `/api/courses/exercises/${exerciseId}/response`,
+    headers: auth ? {
+      Authorization: `Bearer ${auth.token}`,
+    } : undefined,
+  }, { manual: true });
+
+  // If the user is authenticated, fetch their response
+  // This is needed as React always expects the same number of hooks to be returned
+  // so we need to return an empty array for the second hook if the user is not authenticated
+  React.useEffect(() => {
+    if (auth) {
+      executeResponse();
+    }
+  }, [auth, executeResponse]);
+
   const exerciseClassNames = 'exercise not-prose my-6';
 
-  switch (data.exercise.type) {
+  switch (exerciseData.exercise.type) {
     case 'Free text':
-      return <FreeTextResponse className={exerciseClassNames} {...data.exercise} exerciseId={exerciseId} />;
+      return (
+        <FreeTextResponse
+          className={exerciseClassNames}
+          {...exerciseData.exercise}
+          exerciseId={exerciseId}
+          exerciseResponse={responseData?.exerciseResponse?.response}
+        />
+      );
     case 'Multiple choice':
-      return <MultipleChoice className={exerciseClassNames} {...data.exercise} exerciseId={exerciseId} />;
+      return (
+        <MultipleChoice
+          className={exerciseClassNames}
+          {...exerciseData.exercise}
+          exerciseId={exerciseId}
+          exerciseResponse={responseData?.exerciseResponse?.response}
+        />
+      );
     default:
       return null;
   }
--- a/apps/website-25/src/components/courses/exercises/Exercise.tsx
+++ b/apps/website-25/src/components/courses/exercises/Exercise.tsx
@@ -26,9 +26,9 @@ const Exercise: React.FC<ExerciseProps> = ({
 
   switch (data.exercise.type) {
     case 'Free text':
-      return <FreeTextResponse className={exerciseClassNames} {...data.exercise} />;
+      return <FreeTextResponse className={exerciseClassNames} {...data.exercise} exerciseId={exerciseId} />;
     case 'Multiple choice':
-      return <MultipleChoice className={exerciseClassNames} {...data.exercise} />;
+      return <MultipleChoice className={exerciseClassNames} {...data.exercise} exerciseId={exerciseId} />;
     default:
       return null;
   }
--- a/apps/website-25/src/components/courses/exercises/Exercise.tsx
+++ b/apps/website-25/src/components/courses/exercises/Exercise.tsx
@@ -6,7 +6,6 @@ import FreeTextResponse from './FreeTextResponse';
 import MultipleChoice from './MultipleChoice';
 import { GetExercise } from '../../../pages/api/courses/exercises/[exerciseId]';
 import { GetExerciseResponseResponse, PutExerciseResponseRequest } from '../../../pages/api/courses/exercises/[exerciseId]/response';
-import { error } from 'console';
 
 type ExerciseProps = {
   // Required
@@ -33,11 +32,12 @@ const Exercise: React.FC<ExerciseProps> = ({
     } : undefined,
   });
 
-  const handleExerciseSubmit = useCallback(async (exerciseResponse: string) => {
+  const handleExerciseSubmit = useCallback(async (exerciseResponse: string, completed?: boolean) => {
     await axios.put<unknown, unknown, PutExerciseResponseRequest>(
       `/api/courses/exercises/${exerciseId}/response`,
       {
         response: exerciseResponse,
+        completed,
       },
       {
         headers: {
--- a/apps/website-25/src/components/courses/exercises/Exercise.tsx
+++ b/apps/website-25/src/components/courses/exercises/Exercise.tsx
@@ -1,10 +1,12 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import useAxios from 'axios-hooks';
+import axios from 'axios';
 import { useAuthStore } from '@bluedot/ui';
 import FreeTextResponse from './FreeTextResponse';
 import MultipleChoice from './MultipleChoice';
 import { GetExercise } from '../../../pages/api/courses/exercises/[exerciseId]';
-import { GetExerciseResponse } from '../../../pages/api/courses/exercises/[exerciseId]/response';
+import { GetExerciseResponseResponse, PutExerciseResponseRequest } from '../../../pages/api/courses/exercises/[exerciseId]/response';
+import { error } from 'console';
 
 type ExerciseProps = {
   // Required
@@ -14,36 +16,43 @@ type ExerciseProps = {
 const Exercise: React.FC<ExerciseProps> = ({
   exerciseId,
 }) => {
+  const exerciseClassNames = 'exercise not-prose my-6';
+
+  const auth = useAuthStore((s) => s.auth);
+  
   const [{ data: exerciseData }] = useAxios<GetExercise>({
     method: 'get',
     url: `/api/courses/exercises/${exerciseId}`,
   });
 
-  if (!exerciseData || !exerciseData.exercise) {
-    console.error('Exercise not found');
-    return null;
-  }
-
-  const auth = useAuthStore((s) => s.auth);
-
-  const [{ data: responseData }, executeResponse] = useAxios<GetExerciseResponse>({
+  const [{ data: responseData }, refetch] = useAxios<GetExerciseResponseResponse>({
     method: 'get',
     url: `/api/courses/exercises/${exerciseId}/response`,
     headers: auth ? {
       Authorization: `Bearer ${auth.token}`,
     } : undefined,
-  }, { manual: true });
+  });
 
-  // If the user is authenticated, fetch their response
-  // This is needed as React always expects the same number of hooks to be returned
-  // so we need to return an empty array for the second hook if the user is not authenticated
-  React.useEffect(() => {
-    if (auth) {
-      executeResponse();
-    }
-  }, [auth, executeResponse]);
+  const handleExerciseSubmit = useCallback(async (exerciseResponse: string) => {
+    await axios.put<unknown, unknown, PutExerciseResponseRequest>(
+      `/api/courses/exercises/${exerciseId}/response`,
+      {
+        response: exerciseResponse,
+      },
+      {
+        headers: {
+          Authorization: `Bearer ${auth?.token}`,
+        },
+      },
+    );
+    
+    await refetch();
+  }, [exerciseId, auth, refetch]);
 
-  const exerciseClassNames = 'exercise not-prose my-6';
+  if (!exerciseData || !exerciseData.exercise) {
+    console.error('Exercise not found');
+    return null;
+  }
 
   switch (exerciseData.exercise.type) {
     case 'Free text':
@@ -51,8 +60,8 @@ const Exercise: React.FC<ExerciseProps> = ({
         <FreeTextResponse
           className={exerciseClassNames}
           {...exerciseData.exercise}
-          exerciseId={exerciseId}
           exerciseResponse={responseData?.exerciseResponse?.response}
+          onExerciseSubmit={handleExerciseSubmit}
         />
       );
     case 'Multiple choice':
@@ -60,8 +69,8 @@ const Exercise: React.FC<ExerciseProps> = ({
         <MultipleChoice
           className={exerciseClassNames}
           {...exerciseData.exercise}
-          exerciseId={exerciseId}
           exerciseResponse={responseData?.exerciseResponse?.response}
+          onExerciseSubmit={handleExerciseSubmit}
         />
       );
     default:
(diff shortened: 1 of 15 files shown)


apps/website-25/src/pages/api/courses/exercises/[exerciseId].ts
```diff
@@ -11,21 +14,71 @@ export type GetExerciseResponse = {
   exercise: Exercise,
 };
 
+export type PutExerciseResponse = {
+  type: 'success',
+  exerciseResponse: ExerciseResponse,
+};
+
 export default makeApiRoute({
   requireAuth: false,
```
Comment (domdomegg) [inline apps/website-25/src/pages/api/courses/exercises/[exerciseId].ts:23]: Possibly you'll want to split this into two separate endpoints, so that for the `PUT`-like response you can set `requireAuth: true`.

From my Slack yesterday:

> From a strict REST niceness perspective using the same path would be ideal, but I think because of how makeApiRoute does authentication (with requireAuth) this will be a bit of a pain. I'd be happy to just have these as separate routes if that makes things easier (which I suspect it will?).

Actually reflecting more on this, I think if we were being nice it'd be `PUT /courses/exercises/[exerciseId]/response` (as you're not PUTing an exercise - unless e.g. you're a course designer. you're PUTing the response). This would also allow you to get the user's previous response, with GET /courses/exercises/[exerciseId]/response.

Alternatively to save on calls you could make `/courses/exercises/[exerciseId]` **always** require authentication, and when GETing it it also returns the exercise response, as well as the exercise. And then when PUTing to this endpoint it really puts the exercise response.

Comment (domdomegg) [reply apps/website-25/src/pages/api/courses/exercises/[exerciseId].ts:23]: (I think actually I prefer the latter approach, but happy with either)

Comment (tarinrickett) [reply apps/website-25/src/pages/api/courses/exercises/[exerciseId].ts:23]: based on latest requirements, i am updating this to:
- GET /api/courses/exercises/${exerciseId} – does not require auth, returns exercise details
- GET /api/courses/exercises/${exerciseId}/response – requires auth, returns saved exercise responses
- PUT /api/courses/exercises/${exerciseId}/response – requires auth, upsert exercise with new submitted state

apps/website-25/src/components/courses/exercises/Exercise.tsx
```diff
@@ -12,23 +15,64 @@ type ExerciseProps = {
...
+
+  const auth = useAuthStore((s) => s.auth);
+
+  const [{ data: exerciseData }] = useAxios<GetExercise>({
     method: 'get',
     url: `/api/courses/exercises/${exerciseId}`,
   });
 
-  if (!data || !data.exercise) {
+  const [{ data: responseData }, refetch] = useAxios<GetExerciseResponseResponse>({
```
domdomegg, line 27: fine for now.

as a follow-up, I think it would be good if we don't try to get a exercise response at all if not logged in. I think you can achieve this with `{ manual: auth == undefined }`, because then if auth is not defined it won't automatically make the request when loading the component (I might be wrong)

domdomegg, line 27: fine for now.

as a follow-up, I think it would be good if we don't try to get a exercise response at all if not logged in. I think you can achieve this with { manual: auth == undefined }, because then if auth is not defined it won't automatically make the request when loading the component (I might be wrong)

domdomegg, line 27: future: I think it would be good if we don't try to get a exercise response at all if not logged in. I think you can achieve this with { manual: auth == undefined }, because then if auth is not defined it won't automatically make the request when loading the component (I might be wrong).

Appreciate you can't do this with an if because of the rules of react hooks.

apps/website-25/src/components/courses/exercises/Exercise.tsx
```diff
@@ -12,23 +15,64 @@ type ExerciseProps = {
...
+        completed,
+      },
+      {
+        headers: {
+          Authorization: `Bearer ${auth?.token}`,
+        },
+      },
+    );
+
+    await refetch();
```
domdomegg, line 49: Does this definitely work well if the user is on a slow connection? I'm kind of worried about:
- user writes some text
- hits save, and the previous response starts refetching...
- writes some additional text
- their slow internet loads their response and deletes their additional text

I don't know if the above actually does happen on the refetch.

An alternative to avoid this might be to have a state variable which we initially set to the fetched value (on first component load), but then update when they want to save (as the first action, before sending it to the server). It saves us a network call and means the above problem doesn't occur.

apps/website-25/src/components/courses/exercises/FreeTextResponse.stories.tsx
```diff
@@ -21,5 +21,13 @@ export const Default: Story = {
   args: {
     title: 'Understanding LLMs',
     description: 'Why is a language model\'s ability to predict \'the next word\' capable of producing complex behaviors like solving maths problems?',
+    onExerciseSubmit: () => {},
```
domdomegg, line 24: nit: Use alert(JSON.log()) or console.log() so it's easy for people in storybook to see what the submit callback looks like?

apps/website-25/src/components/courses/exercises/FreeTextResponse.test.tsx
```diff
@@ -1,10 +1,24 @@
...
+  expect,
+  Mock,
+  test,
+  vi,
+} from 'vitest';
+import axios from 'axios';
 import FreeTextResponse from './FreeTextResponse';
 
+// Mock axios
+vi.mock('axios');
```
domdomegg, line 13: I think because we've now moved all the networking logic out of here (yay) this isn't necessary?

apps/website-25/src/components/courses/exercises/FreeTextResponse.test.tsx
```diff
@@ -17,19 +31,14 @@ describe('FreeTextResponse', () => {
...
-      expect(textareaEl.classList.contains('free-text-response__textarea--saved')).toBeTruthy();
-      expect(container.querySelector('.free-text-response__saved-msg')).toMatchSnapshot();
-    });
+    const textareaEl = container.querySelector('.free-text-response__textarea') as HTMLTextAreaElement;
+
+    expect(container).toMatchSnapshot();
+    expect(textareaEl.classList.contains('free-text-response__textarea--saved')).toBeTruthy();
+    expect(textareaEl.value).toBe('This is my saved answer.');
   });
 });
```
domdomegg, line 44: consider: Adding a test to ensure the onExerciseSubmit is called correctly?

apps/website-25/src/components/courses/exercises/FreeTextResponse.tsx
```diff
@@ -1,36 +1,55 @@
...
+import { CTALinkOrButton } from '@bluedot/ui';
+import React, { useCallback, useEffect } from 'react';
 import ReactMarkdown from 'react-markdown';
+import { useForm } from 'react-hook-form';
 
 type FreeTextResponseProps = {
   // Required
-  title: string;
-  description: string;
   className?: string;
```
domdomegg, line 9: nit: move this to the optional section

apps/website-25/src/components/courses/exercises/FreeTextResponse.tsx
```diff
@@ -1,36 +1,55 @@
...
+    defaultValues: {
+      answer: exerciseResponse || '',
+    },
+  });
 
-  const handleAnswerSubmit = () => {
-    setSubmittedOption('test');
-    setIsSaved(true);
-  };
+  useEffect(() => {
```
domdomegg, line 35: This logic will make one of my comments above a problem (about saving on slow networks).

I'm guessing you've put this here because the exerciseResponse isn't loaded at first render, as it's fetched from the network. I think maybe we'd prefer displaying a loader in Exercise.tsx if either the exercise or the response isn't loaded? So on first render the exerciseResponse is loaded.

apps/website-25/src/components/courses/exercises/FreeTextResponse.tsx
```diff
@@ -1,36 +1,55 @@
...
-  const handleAnswerChange = () => {
-    if (submittedOption && isSaved) {
-      setIsSaved(false);
+  const onSubmit = useCallback(async (data: FormData) => {
+    try {
+      await onExerciseSubmit(data.answer);
+      setIsEditing(false);
+    } catch (e) {
+      console.log('error', e);
+      // Set some state for the error and render it
```
domdomegg, line 47: nit:

```suggestion
      // TODO: Set some state for the error and render it
```

just so it's easier to find and bash these later

apps/website-25/src/components/courses/exercises/FreeTextResponse.tsx
```diff
@@ -42,23 +61,23 @@ const FreeTextResponse: React.FC<FreeTextResponseProps> = ({
...
           className={`free-text-response__textarea p-4${
-            isSaved ? ' free-text-response__textarea--saved container-active bg-[#63C96533] border-[#63C965] text-[#2A5D2A]' : ' container-lined'
+            (!isEditing && exerciseResponse) ? ' free-text-response__textarea--saved container-active bg-[#63C96533] border-[#63C965] text-[#2A5D2A]' : ' container-lined'
           }`}
           placeholder="Enter your answer here"
-          onChange={handleAnswerChange}
+          onChange={() => setIsEditing(true)}
         />
       </div>
       <CTALinkOrButton
```
domdomegg, line 72: consider: You might want to get `formState` from `useForm`. It has `formState.isSubmitting`, which you can use to disable the button and change the text to 'Saving' to give the user more feedback.

apps/website-25/src/components/courses/exercises/FreeTextResponse.tsx
```diff
@@ -42,23 +61,23 @@ const FreeTextResponse: React.FC<FreeTextResponseProps> = ({
...
         variant="primary"
-        onClick={handleAnswerSubmit}
-        type="button"
+        type="submit"
       >
         Save
       </CTALinkOrButton>
-      {isSaved && <p className="free-text-response__saved-msg">Saved! 🎉</p>}
-    </div>
+      {(!isEditing && exerciseResponse) && <p className="free-text-response__saved-msg">Saved! 🎉</p>}
```
domdomegg, line 79: consider: Similar to the above you can get `isSubmitSuccessful` and `isDirty` from `formState`, using the check `isSubmitSuccessful && !isDirty`

And in onSubmit you can call reset with keep values to mark things as 'clean'
```ts
reset({}, { keepValues: true });
```

This will save you from having to track isEditing yourself, and get you the behaviour I think you want: where by default nothing is shown. Only after submitting successfully, is the saved message shown. Once the user edits any of the fields it'll be marked as dirty and the message will hide.

apps/website-25/src/components/courses/exercises/MultipleChoice.stories.tsx
```diff
@@ -22,6 +22,21 @@ export const Default: Story = {
     title: 'Understanding LLMs',
     description: 'Why is a language model\'s ability to predict \'the next word\' capable of producing complex behaviors like solving maths problems?',
     options: 'The community\'s preference for low-tech fishing traditions\nRising consumer demand for fish with more Omega-3s\nEnvironmental regulations and declining cod stocks\nA cultural shift toward vegetarianism in the region\n',
-    answer: 'The community\'s preference for low-tech fishing traditions',
+    answer: 'The community\'s preference for low-tech fishing traditions\n',
+    onExerciseSubmit: () => {},
+  },
+};
+
+export const SavedCorrectResponse: Story = {
```
domdomegg, line 30: nice

apps/website-25/src/components/courses/exercises/FreeTextResponse.tsx
```diff
@@ -1,36 +1,55 @@
 import clsx from 'clsx';
```
domdomegg, line 1: This is so much cleaner without having to have all the networking logic here 🎉 

Excited that this also makes it easier to show in storybook and write tests for. Plus building new types of exercises should be easier.

apps/website-25/src/components/courses/exercises/MultipleChoice.test.tsx
```diff
@@ -1,14 +1,26 @@
...
+  expect,
+  Mock,
+  test,
+  vi,
+} from 'vitest';
+import axios from 'axios';
 import MultipleChoice from './MultipleChoice';
 
+// Mock axios
+vi.mock('axios');
```
domdomegg, line 13: nit: Same as above, I don't think this is necessary any more

apps/website-25/src/components/courses/exercises/MultipleChoice.stories.tsx
```diff
@@ -22,6 +22,21 @@ export const Default: Story = {
     title: 'Understanding LLMs',
     description: 'Why is a language model\'s ability to predict \'the next word\' capable of producing complex behaviors like solving maths problems?',
     options: 'The community\'s preference for low-tech fishing traditions\nRising consumer demand for fish with more Omega-3s\nEnvironmental regulations and declining cod stocks\nA cultural shift toward vegetarianism in the region\n',
-    answer: 'The community\'s preference for low-tech fishing traditions',
+    answer: 'The community\'s preference for low-tech fishing traditions\n',
+    onExerciseSubmit: () => {},
```
domdomegg, line 26: nit: same as above, picking a function that's easier to see in storybook

apps/website-25/src/components/courses/exercises/MultipleChoice.tsx
```diff
@@ -1,51 +1,84 @@
...
+  exerciseResponse,
+  onExerciseSubmit,
+  options,
+  title,
 }) => {
   /**
    * Options are stored as a string with newlines
    * Format them to be an array of strings with no empty strings (i.e., removing trailing return statements)
    */
   const formattedOptions = (options.split('\n').map((o) => o.trim()).filter((o) => o !== ''));
```
domdomegg, line 35: nit:

```suggestion
  const formattedOptions = options.split('\n').map((o) => o.trim()).filter((o) => o !== '');
```

apps/website-25/src/components/courses/exercises/MultipleChoice.tsx
```diff
@@ -1,51 +1,84 @@
...
+  });
 
   const handleOptionSelect = (option: string) => {
     setSelectedOption(option);
-    setSubmittedOption(null); // Reset the submitted option when a new option is selected
+    setIsEditing(true);
   };
 
   const isSelected = (option: string): boolean => {
+    if (!isEditing && formattedExerciseResponse) {
```
domdomegg, line 53: consider: I don't understand the purpose of this if statement - maybe remove it if not, or add a comment explaining why it's here

apps/website-25/src/components/courses/exercises/MultipleChoice.tsx
```diff
@@ -1,51 +1,84 @@
...
+
+  const onSubmit = useCallback(async (data: FormData) => {
+    try {
+      const isAnswerCorrect = data.answer === formattedAnswer;
+      console.log('submitted this answer: ', data.answer, ' and is it correct? ', isAnswerCorrect);
+      await onExerciseSubmit(data.answer, isAnswerCorrect);
+      setIsEditing(false);
+    } catch (e) {
+      console.log('error', e);
+      // Set some state for the error and render it
```
domdomegg, line 73: nit: so we can find this later

```suggestion
      // TODO: Set some state for the error and render it
```

apps/website-25/src/pages/api/courses/exercises/[exerciseId]/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  responseBody: z.object({
+    type: z.literal('success'),
+    exercise: z.any().optional(),
+  }),
+}, async (body, { raw }) => {
+  const { exerciseId } = raw.req.query;
+
+  switch (raw.req.method) {
+    case 'GET': {
+      const exercise = (await db.scan(exerciseTable, {
```
Comment (domdomegg) [inline apps/website-25/src/pages/api/courses/exercises/[exerciseId]/index.ts:26]: consider: using `db.get(exerciseTable, exerciseId)` - this will load the record from Airtable faster

apps/website-25/src/pages/api/courses/exercises/[exerciseId]/response.ts
```diff
@@ -0,0 +1,88 @@
+import { z } from 'zod';
+import createHttpError from 'http-errors';
+import db from '../../../../../lib/api/db';
+import { makeApiRoute } from '../../../../../lib/api/makeApiRoute';
+import {
+  ExerciseResponse,
+  exerciseResponseTable,
+} from '../../../../../lib/api/db/tables';
+
+export type GetExerciseResponseResponse = {
```
Comment (domdomegg) [inline apps/website-25/src/pages/api/courses/exercises/[exerciseId]/response.ts:10]: 😄

(but it makes sense and is in line with convention of others, I think you should keep it this way)

apps/website-25/src/pages/api/courses/exercises/[exerciseId]/response.ts
```diff
@@ -0,0 +1,88 @@
...
+  type: 'success',
+  exerciseResponse: ExerciseResponse,
+};
+
+export type PutExerciseResponseRequest = {
+  response: string,
+  completed?: boolean,
```
Comment (domdomegg) [inline apps/website-25/src/pages/api/courses/exercises/[exerciseId]/response.ts:17]: I'm not sure why this should be optional - I'd expect all exercises should be able to have some sense of being completed or not?

apps/website-25/src/pages/api/courses/exercises/[exerciseId]/response.ts
```diff
@@ -0,0 +1,88 @@
...
+  requireAuth: true,
+  requestBody: z.optional(
+    z.object({
+      response: z.string(),
+      completed: z.boolean().optional(),
+    }),
+  ),
+  responseBody: z.object({
+    type: z.literal('success'),
+    exerciseResponse: z.any().optional(),
```
Comment (domdomegg) [inline apps/website-25/src/pages/api/courses/exercises/[exerciseId]/response.ts:30]:
```suggestion
    exerciseResponse: z.any(),
```

I don't see a codepath where this doesn't get returned on a successful response?

domdomegg (approved):
Awesome, very excited to have exercises working on the new course hub. Great work here.

I'd be happy to merge this now, and then for you to follow up on the other comments as a separate PR. Also happy for you to continue on this branch - up to you.

tarinrickett:
merging as-is and will follow up with improvements 🚀

PR #675, opened 2025-04-23
website-25: Fix explore menu in Nav on desktop

### LINKED ISSUE #673: Nav drop open menu doesn't work on desktop
https://github.com/user-attachments/assets/422740a4-12a7-4385-98c4-6917e2b41bf0

Regression introduced in #649

domdomegg:
# Summary

website-25: Fix explore menu in Nav on desktop

The desktop explore panel was incorrectly nested inside the mobile panel. This PR moves it out of the mobile panel, so it displays properly on desktop. The `hidden lg:flex` already stops it being shown on mobile.

## Issue

Fixes #673

### DIFF
--- a/apps/website-25/src/components/Nav.tsx
+++ b/apps/website-25/src/components/Nav.tsx
@@ -278,14 +278,14 @@ export const Nav: React.FC<NavProps> = ({
               )}
             </div>
           </div>
-          <div className={clsx('nav__links-drawer', drawerBaseClassName, expandedSections.explore ? drawerOpenClassName : drawerClosedClassName)}>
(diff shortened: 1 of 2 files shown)


tarinrickett (approved):
can you add mobile and desktop screen recording for sanity check to confirm?

Ps ive been investigating AI agent QA solutions for us. would love to have some low-lift integrated process to catch bugs like this in the future

domdomegg:
> can you add mobile and desktop screen recording for sanity check to confirm?

https://github.com/user-attachments/assets/89202fa9-8bf3-435a-97ba-763793716f36

PR #685, opened 2025-04-23
login: Update keycloak version

### LINKED ISSUE #679: Update Keycloak

domdomegg:
# Summary

Update keycloak version to get:
- security updates
- performance improvements

Tried to follow https://www.keycloak.org/docs/latest/upgrading/index.html closely, and tested it ran locally. But some of the config there is quite complex, so while I'm fairly confident this is correct I'm not 100%. I think we should do this before we go live so that if there is an issue it's not affecting real users badly.

## Issue

Fixes #679

### DIFF
--- a/apps/infra/Pulumi.prod.yaml
+++ b/apps/infra/Pulumi.prod.yaml
@@ -4,8 +4,6 @@ config:
     secure: v1:xxnGOjci0H2TTfIY:p5/SIY+vels8uuUTUCJYRW3JIIYJNF9Jmr1m8usCqYgjU073BMKjQ+uq3dUb42f1iNeOs0A124ZIzDjxv9f1cJNaOd8gfvAxQf1oA1WC/A/CqnHyoNFIG6zpE1C0gNtu+u8=
   infra:alertsSlackBotToken:
     secure: v1:KHo0xEvyljpu6RQu:FxDE6U4NfjkQTTCE+yLJKpF2tnjiNqGM4eWfgnSMgD+3RKyicTHqUuVKYyEQdNYj9YZkokYeD6X0eB277AYJHQUV3ZhbrrO/vA==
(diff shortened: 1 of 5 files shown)


apps/infra/Pulumi.prod.yaml
```diff
@@ -4,8 +4,6 @@ config:
     secure: v1:xxnGOjci0H2TTfIY:p5/SIY+vels8uuUTUCJYRW3JIIYJNF9Jmr1m8usCqYgjU073BMKjQ+uq3dUb42f1iNeOs0A124ZIzDjxv9f1cJNaOd8gfvAxQf1oA1WC/A/CqnHyoNFIG6zpE1C0gNtu+u8=
   infra:alertsSlackBotToken:
     secure: v1:KHo0xEvyljpu6RQu:FxDE6U4NfjkQTTCE+yLJKpF2tnjiNqGM4eWfgnSMgD+3RKyicTHqUuVKYyEQdNYj9YZkokYeD6X0eB277AYJHQUV3ZhbrrO/vA==
-  infra:keycloakAdminPassword:
```
domdomegg, line 7: No longer needed, as admin user is already set up

apps/login/package.json
```diff
@@ -3,9 +3,10 @@
     "version": "1.0.0",
     "private": true,
     "scripts": {
-        "start:docker": "docker-scripts start",
+        "deploy:cd": "docker-scripts deploy",
         "build": "tools/getBluedotKeycloakTheme.sh",
-        "deploy:cd": "docker-scripts deploy"
+        "start": "docker-scripts start",
+        "test": "docker-scripts test"
```
domdomegg, line 9: This makes the CI actually try to build the image on PRs, giving us more confidence in CI checks.

Other script edits make it easier to do local development

PR #647, opened 2025-04-15
Link to course from profile page

Will-Howard:
# Summary

Returns coursePath from the `/api/me` route, and uses it to link to course if it is in-progress.

## Issue

#646 

This doesn't fully resolve the ticket, still to do:
- Making the referral flow work end to end (currently it doesn't) and adding a test for this

## Developer checklist
- [X] [N/A] Used [BEM naming convention](https://getbem.com/naming/) for classNames
- [ ] Added or updated Jest tests
- [ ] Added or updated Storybook stories

## Screenshot
The only visual change is adding the "Continue" button:

| 📸 |  |
|---------|---|
| 📕 |  |
| 🖥️ | Screenshot 2025-04-15 at 11 25 06 |
| 📱  | Screenshot 2025-04-15 at 11 25 21 |

## Testing
```
$ npm run test:update
 Tasks:    13 successful, 13 total
Cached:    1 cached, 13 total
  Time:    24.451s
```

PR #690, opened 2025-04-23
Add util for splitting comma separated strings

### LINKED ISSUE #648: Create util for splitting comma separated strings
See comment [here](https://github.com/bluedotimpact/bluedot/pull/647#discussion_r2044317855). There will be a few cases of this coming from Airtable

Will-Howard:
# Summary

I'm not actually convinced of the utility of this. Apart from the one here, the [other case we have](https://github.com/bluedotimpact/bluedot/pull/647#discussion_r2044317855) of a string that needs to be split coming from Airtable is too different to combine in the function, and it remains to be seen if the cases will tend to be that similar (e.g. if it makes sense to trim the strings always or not).

## Issue

Fixes #648

## Developer checklist
- [X] [N/A] Used [BEM naming convention](https://getbem.com/naming/) for classNames
- [X] [N/A] Added or updated Storybook stories

## Screenshot

| 📸 |  |
|---------|---|
| 📕 |  |
| 🖥️ |  |
| 📱  |  |

## Testing
```
$ npm run test:update
 Tasks:    13 successful, 13 total
Cached:    11 cached, 13 total
  Time:    12.222s
```

### DIFF
--- a/apps/website/src/components/courses/exercises/MultipleChoice.tsx
+++ b/apps/website/src/components/courses/exercises/MultipleChoice.tsx
@@ -4,6 +4,7 @@ import React, { useCallback, useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { ROUTES } from '../../../lib/routes';
 import { P } from '../../Text';
+import { formatStringToArray } from '../../../lib/utils';
 
 type MultipleChoiceProps = {
   // Required
@@ -36,7 +37,7 @@ const MultipleChoice: React.FC<MultipleChoiceProps> = ({
    * Options are stored as a string with newlines
    * Format them to be an array of strings with no empty strings (i.e., removing trailing return statements)
    */
-  const formattedOptions = (options.split('\n').map((o) => o.trim()).filter((o) => o !== ''));
+  const formattedOptions = formatStringToArray(options, '\n');
   const formattedAnswer = answer.trim();
   const formattedExerciseResponse = exerciseResponse?.trim();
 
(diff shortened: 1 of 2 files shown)


apps/website-25/src/lib/utils.ts
```diff
@@ -0,0 +1,3 @@
+export const splitCommaString = (input: string | null | undefined): string[] => {
+  return input ? input.split(',').filter((item) => item !== '') : [];
+};
```
tarinrickett, line 3: we could abstract this to something more like:
```suggestion
export const formatStringToArray = (input: string | null | undefined, splitBy: string): string[] => {
  return input ? input.split(splitBy).map((o) => o.trim()).filter((item) => item !== '') : [];
};
```
that way we can use it across more general airtable formatting usecases:
```
// users/me.ts
const courseNames = formatStringToArray(user?.courseSitesVisited, ',');

// MultipleChoice.tsx
const formattedOptions = formatStringToArray(options, '/n');
```

i think trim() will likely be used in all cases. if in the future we find one that doesn't require it, we could always add a new option doNotTrim or something to handle it

i find having this helper useful since we know airtable doesn't handle arrays, so it is likely we will continue encountering it.

Will-Howard: Seems reasonable 👍. I've updated it to this now and applied it the the one case there is at the moment

tarinrickett (approved):

tarinrickett:
would be good to include a screenshot/screen recording that MCQ are working before merging 👍

tarinrickett:
also see: https://github.com/bluedotimpact/bluedot/pull/707#discussion_r2058064467

PR #698, opened 2025-04-23
addQueryParam helper util

tarinrickett:
# Summary
addQueryParam helper util

## Description

- created new `utils/` directory (is there anywhere that exists where this should live instead? open to discussion)
- addQueryParam helper handles appending to URLs with or without existing params for safety
- updated login URLs in FreeTextResponse and MultipleChoice

## Testing
```
$ npm run test:update
```


libraries/ui/src/utils/addQueryParam.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(result).toBe('https://example.com/?existing=param&key=value');
+  });
+
+  test('adds query parameter to relative path URL', () => {
+    const url = '/login';
+    const result = addQueryParam(url, 'key', 'value');
+    expect(result).toBe('http://localhost:3000/login?key=value');
+  });
+
+  test('handles empty key by returning original URL', () => {
```
domdomegg, line 23: I think probably we should throw in this case

libraries/ui/src/utils/addQueryParam.test.ts
```diff
@@ -0,0 +1,38 @@
+import { describe, test, expect } from 'vitest';
+import { addQueryParam } from './addQueryParam';
+
+describe('addQueryParam', () => {
```
domdomegg, line 4: would be curious what behaviour is when:
- submitting an invalid URL (I think this throws - reasonable)
- adding a parameter that already exists on the URL (not sure what is reasonable to do here - maybe overwrite?)

domdomegg (approved):

domdomegg:
> this isn't getting passed all the way through. login pages (http://localhost:8000/login?redirect_to=http%3A%2F%2Flocalhost%3A8000%2Fcourses%2Ffuture-of-ai%2F1) reroutes to https://login.bluedot.org/ with redirect_uri=http://localhost:8000/login/oauth-callback.

This might be as designed - I think some providers e.g. Google login don't support arbitrary redirects so instead it's stored in local storage and retrieved after login success. Can you check whether it does actually redirect to future of AI on successful login? If not that is a problem.

tarinrickett:
merging this and will continue working off of it for addQueryParam and BE redirect fixes !

libraries/ui/src/utils/addQueryParam.test.ts
domdomegg: Small followup in https://github.com/bluedotimpact/bluedot/pull/741

libraries/ui/src/utils/addQueryParam.test.ts
domdomegg: Turns out this is in spec - handled in https://github.com/bluedotimpact/bluedot/pull/741

PR #704, opened 2025-04-23
ui(auth): add refresh token logic

### LINKED ISSUE #694: website-25: login timeout is too short

domdomegg:
# Summary

Adds refresh token logic to auth.tsx, enabling long-lived and secure sessions.

## Issue

Fixes #694

## Description

Uses OidcClient instead of UserManager so we have more control over the refreshing logic - with the built-in refreshing of UserManager it's not clear how we get notified of changes to the token, and we can't hydrate the refresh token from localstorage AFAIK.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ generated/
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+core
 
 # Misc
 .DS_Store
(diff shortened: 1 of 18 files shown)


domdomegg:
Merging this to avoid merge conflicts with some other work I'm doing near to login (see #696)

PR #718, opened 2025-04-24
Fix BE login redirect logic

tarinrickett:
# Summary

Fix BE login redirect logic

## Description
There were 2 issues in our implementation:
1. [useSearchParams()](https://nextjs.org/docs/app/api-reference/functions/use-search-params) is an async hook, so on first render we may not have fetched the actual query params.
2. The `state` custom data passed in during SigninRequest creation is transformed into `userState` set on the User obj during the response

### Redirects for login
https://github.com/user-attachments/assets/4e543280-5ebc-4ab8-8d7c-cdbd763e7a41

### Does not redirect for Google
https://github.com/user-attachments/assets/09b6b3fa-05b8-457a-8630-82e6f1cbadd2

### Does not redirect on new account creation
https://github.com/user-attachments/assets/5807e229-e2bd-49c9-b523-96c0e614845f

## Testing
Opened a ticket to address testing separately: [Add tests to Login.tsx](https://github.com/bluedotimpact/bluedot/issues/720)

### DIFF
--- a/libraries/ui/src/legacy/Login.tsx
+++ b/libraries/ui/src/legacy/Login.tsx
@@ -144,11 +144,18 @@ export const loginPresets = {
 
 export const LoginRedirectPage: React.FC<LoginPageProps> = ({ oidcSettings }) => {
   const searchParams = useSearchParams();
-  const redirectTo = searchParams.get('redirect_to') || '/';
+  const [redirectTo, setRedirectTo] = useState<string>('/');
   const auth = useAuthStore((s) => s.auth);
 
   useEffect(() => {
-    if (!auth) {
+    const redirect = searchParams.get('redirect_to');
+    if (redirect) {
+      setRedirectTo(redirect);
+    }
+  }, [searchParams]);
+
+  useEffect(() => {
+    if (!auth && redirectTo) {
       new OidcClient(oidcSettings)
         .createSigninRequest({
           request_type: 'si:r',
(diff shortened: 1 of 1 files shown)


domdomegg:
> [useSearchParams()](https://nextjs.org/docs/app/api-reference/functions/use-search-params) is an async hook, so on first render we may not have fetched the actual query params.

this is a dumb design decision from the next.js team :(

libraries/ui/src/legacy/Login.tsx
```diff
@@ -144,11 +144,18 @@ export const loginPresets = {
 
 export const LoginRedirectPage: React.FC<LoginPageProps> = ({ oidcSettings }) => {
   const searchParams = useSearchParams();
```
domdomegg, line 146: Better approach might just be manually doing this, and not using Next's hooks

i.e. `new URLSearchParams(window.location.href)`

We're not even using Next's hook functionality - i.e. we don't expect to continue rendering this page through URL changes so we don't lose anything (except the dumb async behaviour).

tarinrickett: yeah i think you're right. i can set up a getQueryParam() util to go with out addQueryParam() and we can just handle this manually

libraries/ui/src/legacy/Login.tsx
```diff
@@ -144,11 +144,18 @@ export const loginPresets = {
...
   useEffect(() => {
-    if (!auth) {
+    const redirect = searchParams.get('redirect_to');
+    if (redirect) {
+      setRedirectTo(redirect);
+    }
+  }, [searchParams]);
+
+  useEffect(() => {
+    if (!auth && redirectTo) {
```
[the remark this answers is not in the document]
domdomegg: My guess is that this might work temperamentally because the page unload might be slower than the second render - and then this code will run again, and try redirecting to the actual page which overrides the previous redirect. But this feels like a hack.

tarinrickett:
moved to https://github.com/bluedotimpact/bluedot/pull/727

Issue #714, opened 2025-04-24
Create additional Course feature fields in Airtable


tarinrickett:
[You will require Creator access to Course Builder for this ticket. Reach out to Tarin if you need access support]

0. Images

1. Avg Rating
- Use lookups + formula to get avg rating for all units

2. Course material last updated
- We might want to highlight freshness of material for our courses, as they are regularly updated and this is something that sets us apart from our course offerings like Coursera or LinkedIn Learning, which are largely set once and never revisited. We'll use this to highlight that.

3. Num graduates and/or Num active learners
- Whatever is easiest/accessible at this point

Will-Howard:
> Course material last updated

@tarinrickett is this something you think could be derived from data we already have or should it be updated manually? (currently assuming manual, as I expect there would be some updates where it wouldn't make sense to update the user-facing value)

Will-Howard:
I've added these to the best of my ability, I'm just blocked on adding "Num graduates":
- [X] Images: Already added by @tarinrickett 
- [X] Avg Rating:
  - Added "[*] Sum unit rating" (fldFZ46lgvJfKE7be) to Unit table
  - Added "[*] Sum all unit rating" (fldGYkNBfNpfBRu65), "[*] Num unit rating" (fldlN4x4oPYBH7zZh), and "[*] Average rating" (fldONpnyJ4OG0StDY) to the Course table
- [X] Course material last updated: Added a manually updated field, "Public last modified at" (fld8g5mMsPqOm75Vz)
- [ ] Num graduates and/or Num active learners:
  - I decided to go with num graduates, because I think this is better proof of the value of the course
  - For non-MOOC courses, adding a column that counts the linked `CourseRegistration`s with a non-empty certificate ID can do this. _I need edit access to the Applications base_ to add this though as the certificate ID isn't currently synced
  - For the MOOC, it isn't yet hooked up to the CourseRegistration table. I can see that we are planning to migrate this so I'm planning to add a temporary workaround in code that counts the users with a `completedMoocAt`

PR #781, opened 2025-04-27
website: Update CourseSearchCard with more fields

### LINKED ISSUE #710: Add "Cadence" and "Level" to Course definitions
Currently, our courses page is very simple.

Update "Course" definition to include these fields:
- Cadence: Weekly, Daily, Self-paced (MOOC)
- Level: Beginner, Intermediate, Advanced

To bed added to airtable (see #714 )
- Avg Rating: Out of 5
- Course material last updated
- Num graduates and/or Num active learners

Image

### LINKED ISSUE #711: Create CourseSearchCard
Update the UI of our current course directory page. Course cards should include:
- Title
- Img
- Description
- Cadence
- Level
- Avg rating

Guidance on UI, but open to changes!!
Image

Will-Howard:
# Summary

- Replaces duration with cadence
  - I'm calling "MOOC" "Self-paced", I thought a lot of people wouldn't know what MOOC stands for
- Adds average rating

## Issue

Fixes #710, Fixes #711

## Developer checklist
- [X] Added or updated Jest tests
- [X] Added or updated Storybook stories

## Screenshot

| 📸 |  |
|---------|---|
| 📕 | Screenshot 2025-04-27 at 15 57 02 |
| 🖥️ | Screenshot 2025-04-27 at 16 11 21 |
| | Screenshot 2025-04-27 at 16 12 58 |
| 📱  | Screenshot 2025-04-27 at 16 11 30 |

## Testing
```
$ npm run test:update
 Tasks:    14 successful, 14 total
Cached:    13 cached, 14 total
  Time:    2.638s
```


apps/storybook/.storybook/preview.tsx
```diff
@@ -1,5 +1,5 @@
-import './globals.css'
 import '../../website/src/globals.css'
+import './globals.css'
```
Will-Howard, line 2: When these were imported the other way around there was a case of `flex-col` overriding `sm:flex-row`. This is because `flex-col` was defined in './globals.css' but `sm:flex-row` wasn't, so it had precedence. Importing them the other way around should ensure that the styling is the same as when running the project directly

apps/website/src/lib/api/db/tables.ts
```diff
@@ -13,6 +13,11 @@ export type Course = {
   shortDescription: string,
   title: string,
   units: string[],
+  cadence: string,
+  level: string,
+  averageRating: number | null,
+  publicLastUpdated: string | null,
+  // numGraduates: number, // TODO
```
Will-Howard, line 20: `numGraduates` is not used yet (neither is `publicLastUpdated`), but I wanted to add all the fields that we ultimately want to have on the table. `numGraduates` is blocked by me getting access to the Applications base, see [comment here](https://github.com/bluedotimpact/bluedot/issues/714#issuecomment-2830091760)

domdomegg (approved):
This looks neat, good work! 🙌 

image

image

apps/website/src/components/courses/CourseSearchCard.tsx
```diff
@@ -29,25 +36,40 @@ export const CourseSearchCard: React.FC<CourseSearchCardProps> = ({
...
-      <div className="course-search-card__content flex flex-col gap-4 flex-1">
+      <div className="course-search-card__content flex flex-col justify-between gap-space-between flex-1">
         <div className="course-search-card__text">
-          <H3 className="course-search-card__title mb-6">
+          <H3 className="course-search-card__title mb-4">
             {title}
           </H3>
           {description && (
-            <P className="course-search-card__description overflow-hidden text-bluedot-black text-ellipsis mb-6">
+            <P className="course-search-card__description text-size-s overflow-hidden text-bluedot-black text-ellipsis mb-6">
```
domdomegg, line 52:
```suggestion
            <P className="course-search-card__description overflow-hidden text-bluedot-black text-ellipsis mb-6">
```

`text-size-s` isn't a valid class (and never was, although we used it a bunch)

The valid thing is `text-size-sm`, which P automatically comes with already

domdomegg:
reopening to try to give CI a nudge

apps/website/src/components/courses/CourseSearchCard.tsx
Will-Howard: Ah good catch, thanks for fixing 🙏

apps/storybook/.storybook/preview.tsx
Will-Howard: Tailwind prunes out classes that aren't used (in a pretty naive way, just by matching exact string). I'm not exactly sure how the compilation works in storybook, but my guess is that `import './globals.css'` only includes a few base classes and classes that are used in the actual `apps/storybook` project, so it doesn't include `sm:flex-row`.

That may not be it but either way this does empirically fix the problem 😌 .

PR #806, opened 2025-04-28
website: Add schema.org markup for courses

### LINKED ISSUE #456: Use Schema.org markup for Course List
https://developers.google.com/search/docs/appearance/structured-data/course

### LINKED ISSUE #587: SEO structured course data

domdomegg:
Adds schema.org markup for courses

This will render client-side, so we'll have to see if the Googlebot loads it properly. But I can't easily test that without deploying it to staging. If it doesn't, we'll have to do something like `getStaticProps` or `getServerSideProps`.

NB: The rating component doesn't work fully on Google's schema.org checker, because we need a total number of reviews count. This doesn't affect the courses appearing in rich text snippets and we discussed this morning getting enrolment counts would be a separate issue.

## Issue

Fixes #456, fixes #587

## Developer checklist

## Screenshot

<no visual changes>


domdomegg:
> i don't think client-side will be picked up appropriately by Google for SEO

Yeah, this is what I'm not sure about. Agree that maybe mapping static data is the easier route.

I think I'd like to try it, and see what happens (Google has a tester we can point at staging) - okay to merge on that basis, then we switch to static data if that doesn't work?

domdomegg:
From https://developers.google.com/search/docs/guides/generate-structured-data-with-javascript

> Another way you can generate structured data is by using JavaScript to either generate all of your structured data or add more information to the server-side rendered structured data. Either way, Google Search can understand and process structured data that's available in the DOM when it renders the page. To learn more about how Google Search processes JavaScript, check out the [JavaScript basics guide](https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics).

domdomegg:
Merging in the interests of getting this out and being able to test in staging - if it turns out we do need to do this server-side I will make a follow-up PR.

PR #818, opened 2025-04-28
website: Fix refresh logic by using id_token instead of access_token

### LINKED ISSUE #761: Refreshed access token from Keycloak does not have audience

domdomegg:
When initially authenticating, we used the `id_token` which has an `aud` claim. This works great.

When refreshing, we mistakenly used the `access_token` instead of `id_token`. This doesn't have an `aud` claim and was the source of our pain. I think we also got confused whether the scope was relevant, which now think wasn't.

## Issue

Fixes #761


domdomegg:
Updating tests...

PR #768, opened 2025-04-26
Add traffic routing for certificates

tarinrickett:
# Summary

Add traffic routing for certificates

## Issue
https://bluedotimpact.slack.com/archives/C04CSNYF5QX/p1745615555418379

## Description

- Add traffic routing for anything in /certification route

## Testing
Wish we could mock out traffic changes somehow?

### Staging
https://website-staging.k8s.bluedot.org/certification/?id=recg3FUpPCHAxWge5

image

Confirmed with other's certificates

https://website-staging.k8s.bluedot.org/certification?id=reckcH9U0KVA1jxMe

image

### Prod
image

image

PR #835, opened 2025-04-29
website-proxy: Fix contact page

domdomegg:
I think this got deleted accidentally in #768. This results in a weird redirect to `https://bluedot.org:8080/contact/` when trying to view this page.

### DIFF
--- a/apps/website-proxy/src/nginx.template.conf
+++ b/apps/website-proxy/src/nginx.template.conf
@@ -85,6 +85,10 @@ http {
             proxy_pass https://website-production.k8s.bluedot.org;
         }
 
+        location = /contact {
+            proxy_pass https://website-production.k8s.bluedot.org;
+        }
+
         location = /contact/ {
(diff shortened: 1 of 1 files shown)


apps/website-proxy/src/nginx.template.conf
```diff
@@ -85,6 +85,10 @@ http {
             proxy_pass https://website-production.k8s.bluedot.org;
         }
 
+        location = /contact {
+            proxy_pass https://website-production.k8s.bluedot.org;
+        }
+
         location = /contact/ {
             proxy_pass https://website-production.k8s.bluedot.org;
         }
```
tarinrickett, line 94: should this be 
```suggestion
        location /contact {
            proxy_pass https://website-production.k8s.bluedot.org;
        }
```

domdomegg: Should have both there - so both the trailing slash and non-trailing slash routes get redirected

PR #846, opened 2025-04-30
website: Fix certificate request endpoint

### LINKED ISSUE #845: website: certificate requests failing
https://bluedotimpact.slack.com/archives/C04FGBESPK8/p1745996752326049

domdomegg:
This PR fixes the certificate request endpoint.

When upgrading filterByFormula expressions (#829) I think I made a mistake here - we can't filter on the actual linked record field because then Airtable tries to match the record name rather than ID. So we need to filter on a linked field that has the record id as a string. This is what we were doing previously, but I changed it to the linked record field.

This PR adds the record id string field, and filters using that. I've tested locally and it fetches the exercises correctly.

It would be nice to figure out how to clean this up a little, or at least have a convention for how to do this.

## Issue

Fixes #845

### DIFF
--- a/apps/website/src/lib/api/db/tables.ts
+++ b/apps/website/src/lib/api/db/tables.ts
@@ -166,7 +166,8 @@ export const unitTable: Table<Unit> = {
 export type Exercise = {
   id: string,
   answer: string,
-  courseId: string,
+  courseIdWrite: string,
+  courseIdRead: string,
   exerciseNumber: string,
   description: string,
   options: string,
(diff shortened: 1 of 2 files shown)


tarinrickett:
For future ref:
- [Course ID Read](http://airtable.com/appbiNKDcn1sGPGOG/tbla7lc2MtSSbWVvS/fldc9oyPwJSkeMiAW)
- [Course ID Write](https://airtable.com/appbiNKDcn1sGPGOG/tbla7lc2MtSSbWVvS/fldxcJ5gCihs3iRyE)

PR #857, opened 2025-05-01
Add some images for blog content

tarinrickett:
# Summary

Add some images for blog content

## Description
- MarkdownExtendedRenderer does not support tables
- (This support is non trivial and not widely used)
- Replace blog tables with simple screenshots


domdomegg:
Re blog tables, I think adding remark-gfm might add support for tables

Can see a somewhat related example on my blog for configuring it: https://github.com/domdomegg/domdomegg.github.io/blob/54b94b96185347d4ca175437b66b1d8106d496c5/next.config.ts#L112

(and we already have our custom remark plugin installed for MDX escaping, so you can see how plugins are installed there)

PR #864, opened 2025-05-02
Add MultiSelect checkbox exercises

tarinrickett:
# Summary
Add MultiSelect checkbox exercises

## Description

- Work in progress. Need to debug the select logic.


tarinrickett:
not a priority; will return to fix lingering test cases when we have a use case

Issue #1002, opened 2025-06-13
[Feature] Allow user to edit name


tarinrickett:
### 🧩 Summary

Name should be editible from the profile page

### ✅ Definition of done

Users are able to edit existing names, save updates, and add a name for the first time if one was not added to their profile.

This should update for all certificates.

### 📌 Notes / Designs

<img width="1439" alt="Image" src="https://github.com/user-attachments/assets/be9d099a-13f0-4619-9614-0b49d5090601" />

### How urgent is this?

🕐 Within 1 week

Issue #1005, opened 2025-06-13
[Feature] Allow users to self-serve delete their account


tarinrickett:
### 🧩 Summary

A delete account flow should be available from the profile page

### ✅ Definition of done

- Users are able to delete their account
- Users should confirm their password before doing so
- We should include a warning that certificates will be deleted and all data will be lost
- In AirTable, this member should be deleted

### 📌 Notes / Designs

TBD - Requires designs

### How urgent is this?

🕐 Within 1 week

tarinrickett:
We've got some great WIP settings designs from our designer today to help guide you!

a full settings page creation is not necessarily in scope for these tasks, so i'll leave it up to you how you'd like to break it down. as MVP, we can update the existing Account details section to look more like the future settings page layout

![Image](https://github.com/user-attachments/assets/9a5fbf7c-b5f4-4c63-bee8-cc2182563b3e)

![Image](https://github.com/user-attachments/assets/bce42601-160f-4b48-9703-3b93eb4439bb)

PR #1022, opened 2025-06-16
[Feature] Allow user to edit name

Martin-C-Listwan:
# Description
Allow people to modify their profile names. [UX was designed with the new account settings page mind. ](https://github.com/bluedotimpact/bluedot/issues/1005#issuecomment-2970514695)

Also separated ProfileCourseList into its own component to maintain single responsibility. Keeping presentation distinct from the API/logic layer which is inside of the main ProfilePage.

Fixes https://github.com/bluedotimpact/bluedot/issues/1002

## Issue
[Related github issue](https://github.com/bluedotimpact/bluedot/issues/1002)

## Developer checklist

- [Yes] Front-end code follows the [BEM class name convention](https://getbem.com/naming/)
- [Yes] Considered adding tests
- [Yes] Considered adding Storybook stories, and decided that updates weren't needed until settings redesign was complete

## Screenshot
@tarinrickett recommended loom. So LMK how this is as a replacement. 
https://www.loom.com/share/2db34d390aae4f29839bfdc3319643b8?sid=944b0d54-9909-4ae2-9f9e-d9fad6c5ca1b


apps/website/src/pages/api/users/me.ts
```diff
@@ -21,8 +21,17 @@ export type PatchUserBody = {
 export default makeApiRoute({
   requireAuth: true,
   requestBody: z.object({
-    name: z.string().optional(),
-    referredById: z.string().optional(),
+    name: z.string()
+      .trim()
+      .min(1, 'Name cannot be empty')
+      // 50 characters for a name seemed reasonable
+      .max(50, 'Name must be under 50 characters')
```
tarinrickett, line 28: i see we already do this validation UI side. do we need to duplicate it here?

in theory there could be direct API calls set up somewhere, but i dont think we have a use case for this, so i'm more inclined to simplify things and have a single source of truth for validations

apps/website/src/pages/api/users/me.ts
```diff
@@ -21,8 +21,17 @@ export type PatchUserBody = {
...
+      .regex(/^[\p{L}\s\-'.]+$/u, 'Name can only contain letters, spaces, hyphens, apostrophes, and periods')
```
tarinrickett, line 29: can we include this check client-side to avoid unnecessary API calls?

apps/website/src/pages/api/users/me.ts
```diff
@@ -21,8 +21,17 @@ export type PatchUserBody = {
 export default makeApiRoute({
   requireAuth: true,
   requestBody: z.object({
-    name: z.string().optional(),
-    referredById: z.string().optional(),
+    name: z.string()
+      .trim()
+      .min(1, 'Name cannot be empty')
```
tarinrickett, line 26: name is actually optional, as some onboarding flows won't pass it. e.g., Google Auth doesn't include name today. we should keep this as truly optional from API POV

apps/website/src/pages/api/users/me.ts
```diff
@@ -21,8 +21,17 @@ export type PatchUserBody = {
...
+      // 50 characters for a name seemed reasonable
+      .max(50, 'Name must be under 50 characters')
+      .regex(/^[\p{L}\s\-'.]+$/u, 'Name can only contain letters, spaces, hyphens, apostrophes, and periods')
+      .optional(),
+    referredById: z.string()
+      .trim()
+      .min(1, 'Referral ID cannot be empty')
```
tarinrickett, line 33: `referredById` is optional. i don't think we need to make any changes to it here

apps/website/src/pages/profile.tsx
```diff
@@ -10,9 +10,12 @@ import {
   ErrorSection,
   addQueryParam,
   ClickTarget,
+  Input,
```
tarinrickett, line 13: thanks for reusing existing UI library components !

apps/website/src/pages/profile.tsx
```diff
@@ -83,29 +166,21 @@ const ProfilePage = withAuth(({ auth }) => {
...
-                <CTALinkOrButton url={ROUTES.courses.url}>Join a course</CTALinkOrButton>
-              </div>
-              )}
-              {enrolledCourses.length > 0 && (
-                <>
-                  {enrolledCourses.map(({ course, courseRegistration }) => <ProfileCourseCard key={courseRegistration.id} course={course} courseRegistration={courseRegistration} />)}
-                  <CTALinkOrButton url={ROUTES.courses.url}>Join another course</CTALinkOrButton>
-                </>
-              )}
+              <ProfileAccountDetails
```
tarinrickett, line 172: nice! i like simplifying this by separating it out.

since the entire new name editing updates are only relative to `ProfileAccountDetails`, i think we can extend the separation of concerns even further and migrate this API logic to live within `ProfileAccountDetails` rather than `pages/profile.tsx`. it might even be easier to create a new `ProfileAccountDetails.tsx` completely.

what do you think?

apps/website/src/pages/profile.tsx
```diff
@@ -119,6 +194,114 @@ const ProfilePage = withAuth(({ auth }) => {
...
+}) => {
+  const showButtons = tempName !== (userName || '');
+  return (
+    <>
+      <H3>Account details</H3>
+      <div className="flex flex-col gap-4 container-lined bg-white p-8 h-fit">
+        <div className="profile-name">
+          <div className="flex flex-col gap-2">
+            <P><span className="font-bold">Name:</span></P>
+            <div className="flex gap-2 items-center">
```
tarinrickett, line 228: when it comes to BEM, i recommend naming all elements. this consistency makes it easier to navigate the code, prompts developers to think critically about their HTML semantic structures, and provides ready-to-go references for selective testing or more complex CSS.

you can read more about BEM here: https://getbem.com/

it is my personal preference from my experience, as ive seen this scale very well. but feel free to open this up for discussion and come up with your own preferred standards as a new dev team!

```suggestion
    <div className="profile-account-details">
      <H3 className="profile-account-details__title">Account details</H3>
      <div className="profile-account-details__container flex flex-col gap-4 container-lined bg-white p-8 h-fit">
        <div className="profile-account-details__name-container flex flex-col gap-2">
            <P className="profile-account-details__name-label"><span className="font-bold">Name:</span></P>
            <div className="profile-account-details__name-input-container flex gap-2 items-center">
```

apps/website/src/pages/profile.tsx
```diff
@@ -119,6 +194,114 @@ const ProfilePage = withAuth(({ auth }) => {
...
+    courseRegistration: CourseRegistration;
+  }[];
+};
+
+const ProfileCourseList: React.FC<ProfileCourseListProps> = ({ enrolledCourses }) => {
+  return (
+    <>
+      <H3>Your courses</H3>
+      {enrolledCourses.length === 0 && (
+        <div className="profile__no-courses flex flex-col gap-4 container-lined bg-white p-8 mb-4">
```
tarinrickett, line 284: this block becomes "profile-course-list" as it is now in its own component outside "profile"

```suggestion
        <div className="profile-course-list__no-courses flex flex-col gap-4 container-lined bg-white p-8 mb-4">
```
```suggestion
        <div className="profile__no-courses flex flex-col gap-4 container-lined bg-white p-8 mb-4">
```

apps/website/src/__tests__/pages/profile.test.tsx
```diff
@@ -0,0 +1,327 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
tarinrickett, line 1: this is awesome. thanks so much for adding testing coverage here !!!

apps/website/src/__tests__/pages/profile.test.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+    await waitFor(() => {
+      expect(screen.getByText('Name must be under 50 characters')).toBeInTheDocument();
+    });
+  });
+
+  test('should show save/cancel buttons when name is changed', async () => {
+    render(<ProfilePage />);
+
+    const input = await screen.findByPlaceholderText('Enter your name');
```
tarinrickett, line 180: i recommend querying for elements by classname / ID like `profile-account-details__name-input` to future proof if there are text updates in the future

apps/website/src/pages/profile.tsx
```diff
@@ -211,4 +394,23 @@ const ProfileCourseCard: React.FC<ProfileCourseCardProps> = ({ course, courseReg
   );
 };
 
+const parseZodValidationError = (err: AxiosError<{ error?: string }>): string => {
```
tarinrickett, line 397: if we want to introduce JSON parsing of validation errors, i'd recommend moving this out into some sort of helper, as it seems really reusable across many API handlings in the future

tarinrickett (commented):
this is awesome! thanks for jumping in the deep end with your PR. this will be a great feature add, and the additional testing coverage and component breakdowns are exactly the kind of technical maintenance improvements we're always looking for 🥳 💯

apps/website/src/pages/profile.tsx
Martin-C-Listwan: I've modified the CSS to follow the BEM structure. 

This is my first time using BEM alongside inline Tailwind and React, and I'm still on the fence about it. I've never really felt the need for all that extra verbosity when simple comments like {/* Course details */} or {/* Sharing section */} have done the job just fine. That said, I recognize it's a common pattern throughout the codebase, so I'll roll with it for now!

joshlblue:
This is great (and an excellent choice for a first PR)! It will save us at least 3 emails a week.

apps/website/src/__tests__/pages/profile.test.tsx
```diff
@@ -0,0 +1,397 @@
...
+    const { container } = render(<ProfilePage />);
+
+    // Wait for the component to initialize with user data
+    await waitFor(() => {
+      const nameInput = getNameInput(container);
+      expect(nameInput.value).toBe('John Doe');
+    });
+    expect(screen.getByText('Name:')).toBeInTheDocument();
+    expect(container).toMatchSnapshot();
+  });
```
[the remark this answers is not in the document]
tarinrickett: i like to maintain at least 1 full snapshot of the page and use targeted assertions for the rest. in my experience, a single full snapshot will help provide coverage in the future of changes on shared elements that you may not recognize impact other pages

apps/website/src/__tests__/pages/profile.test.tsx
```diff
@@ -0,0 +1,397 @@
...
+    });
+
+    // Test length validation - client-side validation now happens before API call
+    const longName = 'a'.repeat(51);
+    fireEvent.change(input, { target: { value: longName } });
+
+    // Save button should appear and we can click it
+    await waitFor(() => {
+      expect(screen.getByText('Save')).toBeInTheDocument();
+    });
```
tarinrickett, line 189: nit: i don't think you need to reassert this; it's already covered in the test above

apps/website/src/__tests__/pages/profile.test.tsx
```diff
@@ -0,0 +1,397 @@
...
+    // Verify that no API call was made due to client-side validation
+    expect(mockedAxios.patch).not.toHaveBeenCalled();
+
+    // Test regex validation for invalid characters
+    fireEvent.change(input, { target: { value: 'John@123' } });
+    fireEvent.click(screen.getByText('Save'));
+
+    await waitFor(() => {
+      expect(screen.getByText('Name can only contain letters, spaces, hyphens, apostrophes, and periods')).toBeInTheDocument();
+    });
```
tarinrickett, line 207: nit: i'd separate this into its own test for clarity

apps/website/src/__tests__/pages/profile.test.tsx
```diff
@@ -0,0 +1,397 @@
...
+  });
+
+  test('should show save/cancel buttons when name is changed', async () => {
```
tarinrickett, line 210: imo, this is already covered in the first test, "should allow user to successfully change their name"

apps/website/src/pages/profile.tsx
tarinrickett: unresolving; i dont see this updated? it's just a small nit but LMK if you disagree here

apps/website/src/pages/profile.tsx
tarinrickett: ☝️ any thoughts on this? apologies if this PR is still in progress from the first round of comments!

apps/website/src/lib/utils.ts
```diff
@@ -9,3 +11,22 @@ export const formatStringToArray = (
       .filter((item) => item !== '')
     : [];
 };
+
+export const parseZodValidationError = (err: AxiosError<{ error?: string }>, defaultErrorMessage: string): string => {
```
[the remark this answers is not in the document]
Martin-C-Listwan: Added tests

apps/website/src/pages/profile.tsx
Martin-C-Listwan: I've moved the business logic into ProfileAccountDetails, but kept the API requests in the main component. My thinking is that this lets the network loading icon track all three API requests and only resolve once everything's loaded, following the existing pattern with the original two API requests which seemed reasonable to continue. Good callout though!

As for not extracting it to a separate file: with the upcoming profile changes that'll split this into multiple settings pages, it felt premature. Plus, most pages in the website app seemed to have at least some inline components rather than pure orchestration, and if I did decide to extract, then to be consistent than I should also be extracting ProfileCourseList, which seemed like scope creep.

apps/website/src/__tests__/pages/profile.test.tsx
Martin-C-Listwan: Agreed. I ended up removing it.

apps/website/src/__tests__/pages/profile.test.tsx
Martin-C-Listwan: Changed

apps/website/src/__tests__/pages/profile.test.tsx
Martin-C-Listwan: Changed

apps/website/src/lib/schemas/user/me.schema.ts
```diff
@@ -0,0 +1,13 @@
+import z from 'zod';
+
+export const meRequestBodySchema = z.object({
+  name: z.string()
+    .trim()
+  // 50 characters for a name seemed reasonable
+    .max(50, 'Name must be under 50 characters')
+    .regex(/^[\p{L}\s\-'.]+$/u, 'Name can only contain letters, spaces, hyphens, apostrophes, and periods')
```
domdomegg, line 8: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

apps/website/src/pages/api/users/me.ts
```diff
@@ -69,6 +67,12 @@ export default makeApiRoute({
         throw new createHttpError.BadRequest('PATCH request requires a body');
       }
 
+      // Validate that at least one field is being updated
```
domdomegg, line 70: why?

Martin-C-Listwan: It was to prevent unneeded calls to airtable since I understood [they have a hard limit of 5 QPS per base](https://airtable.com/developers/web/api/rate-limits) regardless of plan, and if you hit it, then you get a 30 second cooldown period. 

I agree that it's unusual and if we were writing to a regular DB, I wouldn't have included it. 

Open to removing it though if Airtable QPS is not a concern.

PR #1029, opened 2025-06-17
feat: pull course in navbar and front page from airbase table instead…

tolgadur:
… of constants.tsx

# Description

## Issue

Fixes #

## Developer checklist

- [ ] Front-end code follows the [BEM class name convention](https://getbem.com/naming/)
- [ ] Considered adding tests

## Screenshot

| 📸 |  |
|---------|---|
| 📱  |  |
| 🖥️ |  |


domdomegg:
Code generally looks fine - what does this look like when loading? (maybe attach a short screen recording/similar?)

tolgadur:
> Code generally looks fine - what does this look like when loading? (maybe attach a short screen recording/similar?)

I think it looks fine imo. I attached a screen recording. Is the differing text between constants.tsx and the airtable db fine? (see PR description for more info).

domdomegg:
Maybe some kind of loading indicator would be good? Otherwise there's an unexpected jump on page load which isn't that great.

I think moving the isNew/isFeatured logic into Airtable sounds sensible rather than hardcoding!

tolgadur:
> Maybe some kind of loading indicator would be good? Otherwise there's an unexpected jump on page load which isn't that great.

Done. Have a look at the updated screenshare.


> I think moving the isNew/isFeatured logic into Airtable sounds sensible rather than hardcoding!

Could you give me rights to add airtable fields to the courses table?

apps/website/src/lib/hooks/useCourses.ts
```diff
@@ -0,0 +1,63 @@
...
+const mapCadenceToCourseType = (cadence: string): CourseType => {
+  switch (cadence) {
+    case 'MOOC':
+      return 'Self-paced';
+    case 'Weekly':
+      return 'In-depth course';
+    case 'Daily':
+      return 'Crash course';
+    default:
+      return '' as CourseType;
```
tarinrickett, line 27: it feels like it makes more sense to let the AirTable CMS handle the actual verbage for this. it will help a largely non-technical team make faster updates.

what do you think about just returning the cadence value as is? we can update these namings in AirTable manually once the change is deployed

apps/website/src/lib/hooks/useCourses.ts
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  // Enrich courses with manual properties and map cadence to CourseType
+  const enrichedCourses: EnrichedCourse[] = data?.courses.map((course) => ({
+    ...course,
+    ...(COURSE_ENRICHMENTS[course.path] || {}),
+    courseType: mapCadenceToCourseType(course.cadence),
+  })) || [];
+
+  // Sort courses: featured first, then new, then alphabetically
```
tarinrickett, line 44: this is a great call!

apps/website/src/pages/_app.tsx
```diff
@@ -39,7 +41,17 @@ const App: React.FC<AppProps> = ({ Component, pageProps }: AppProps) => {
...
+              loading ? (
+                <footer className="footer bg-bluedot-darker py-10">
+                  <div className="footer__container section-base">
+                    <ProgressDots />
+                  </div>
+                </footer>
+              ) : (
+                <Footer courses={courses} logo="/images/logo/BlueDot_Impact_Logo_White.svg" />
+              )
+            )}
```
tarinrickett, line 54: instead of handling this in the base template, i'd recommend maintaining this in Footer

i don't see a use case for a variable `courses` param, so we can make these fetches directly within Footer where they will be used

apps/website/src/components/Nav/__snapshots__/Nav.test.tsx.snap
```diff
@@ -82,26 +82,6 @@ exports[`Nav > renders with courses 1`] = `
                     <div
                       class="nav-dropdown__dropdown-content flex flex-col gap-3 w-fit overflow-hidden mx-auto text-pretty hidden"
                     >
-                      <a
-                        class="bluedot-a not-prose nav-link nav-link-animation w-fit no-underline pt-1"
-                        href="/courses/future-of-ai"
-                        tabindex="0"
-                      >
-                        The Future of AI Course
-                      </a>
```
tarinrickett, line 91: we should update the Nav tests to mock out the API call and provide coverage to confirm NavDropdown renders link elements as expected

apps/website/src/components/homepage/CourseSection.test.tsx
```diff
@@ -1,54 +1,110 @@
...
-    // Verify one of the other courses is present
-    expect(screen.getByText(nonFeaturedCourses[0]!.title)).toBeDefined();
   });
 
   test('tracks clicks on course cards', () => {
     render(<CourseSection />);
 
     // Click the featured course card
-    const featuredCard = screen.getByText(featuredCourse.title);
+    const featuredCard = screen.getByText('Featured Course');
```
tarinrickett, line 91: i generally prefer `querySelector` and getting elements by BEM classnames as a more future-proof way of querying elements, as these are less likely to change than copy

in this case, it'd be [.course-card--featured](https://github.com/bluedotimpact/bluedot/blob/000857699fd1eec24b56a2e449dd2f34f61f7696/libraries/ui/src/CourseCard.tsx#L42C18-L42C39)

we could use getByText too, but if you prefer this, i'd recommend referencing a variable directly like we were previously to ensure that this is future-proofed towards updates too

apps/website/src/pages/_app.tsx
tarinrickett: 👍 ok to keep it more modular too!

PR #1044, opened 2025-06-19
fixes: cadance shown in coursecard

tolgadur:
# Description

We want to directly show the cadance set in the airtable on the website without previously changing it as done before. 

## Developer checklist

- [x] Front-end code follows the [BEM class name convention](https://getbem.com/naming/)
- [x] Considered adding tests

## Screenshot
Screenshot 2025-06-19 at 12 47 22


tolgadur:
> screenshots should be included for all impacted changes:
> 
> image image

Understood. sorry about that!

Issue #1042, opened 2025-06-18
Course units visible on website, even when their status is set to "draft"


WillSaunter:
### 🧩 Summary

In the Course Builder airtable base, course units have a [status field](https://airtable.com/appbiNKDcn1sGPGOG/tblsDKJ8VCyO619nk/viwJf5ygEvoiWOwLj/fldFJbY40IjPXer1Q) which is "draft", "active private" or "active". Only "active" units should show on the course website. Currently it is not filtering by this field so "draft" units are showing.

### 🛝 Steps to reproduce

The unit "[Pandemics - 12: [Do not delete] Frameworks for Evaluating Biosecurity Interventions](https://airtable.com/appbiNKDcn1sGPGOG/pag175sp8pu7Zxxiu?bhGc4=b%3AWzAsWyI3eFFVRSIsNixbInJlY0tuSDlxUXNHbEpoWEIwIiwicmVjTjJhc2xrWDluQzhXVUUiLCJyZWM0RXpub3BxYmJ6T3JsOSJdLCJOZGFpYyJdXQ&zfgpC=rec6IDLJP38fy67Pr)" is set to draft and is showing on the [Pandemics course webpage](https://bluedot.org/courses/pandemics).

### ✅ Definition of done

Only active units are showing on the website

### How urgent is this?

🕐 Within 1 week

tolgadur:
The issue wasn't closed it was just the first ticket.

PR #1063, opened 2025-06-24
Tolgadur/filter unit prompts

tolgadur:
# Description
Course units that were tagged as not "Active" weren't filted out. Also in the course unit page they weren't sorted. I attached before and after screenshots below.

## Issue

Fixes https://github.com/bluedotimpact/bluedot/issues/1042

## Developer checklist

- [x] Front-end code follows the [BEM class name convention](https://getbem.com/naming/)
- [x] Considered adding tests

## Screenshot
**Before:**
Screenshot 2025-06-24 at 10 50 23
Screenshot 2025-06-24 at 10 50 11

**After:**
Screenshot 2025-06-24 at 10 33 56
Screenshot 2025-06-24 at 10 33 38

### DIFF
--- a/apps/website/src/pages/api/courses/[courseSlug]/[unitId]/index.ts
+++ b/apps/website/src/pages/api/courses/[courseSlug]/[unitId]/index.ts
@@ -1,7 +1,7 @@
 import { z } from 'zod';
 import createHttpError from 'http-errors';
 import {
-  eq, asc, chunkTable, unitTable, InferSelectModel,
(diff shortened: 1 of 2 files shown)


Martin-C-Listwan (approved):
As long as you've manually tested and it works, then generally LGTM.

Why aren't we using a requestBody zod schema? That would remove our manual if checks at the top, and also validate unitId to be valid? You could look at me.ts for an example of what I mean by requestBody validation.

PR #1085, opened 2025-06-28
website: Filter unit query to only return active units

domdomegg:
# Description

Add unitStatus: 'Active' condition to unit database query to ensure only active units are retrieved when fetching course unit resources.

This fixes https://bluedot.org/courses/pandemics/12, where there is also an inactive unit 12 that causes an error (because the .get() fails as there is more than one record).


Martin-C-Listwan (approved):
LGTM, as long you’ve manually tested that the broken url now works.

Issue #1079, opened 2025-06-27
[Feature] Capturing people's name on accounbt sign up


AlHusseinSaqr:
### 🧩 Summary

It would be great if we could also capture people's names when they sign up for an account. Currently, we have to manually add people's names into the corresponding Airtable records.

### ✅ Definition of done

Newly created Airtable records have people's names + emails instead of just emails.

### How urgent is this?

🕐 Within 1 week

joshlblue:
fyi this has been partially resolved with https://github.com/bluedotimpact/bluedot/pull/1022/files#top -- users can now add their name which will sync over to airtable!

PR #1126, opened 2025-07-22
feat: modal to set name

tolgadur:
# Description
This introduces a modal that prompts the user to add his name if he hasn't already. The main question is if this is enough or if we want to persist some state in the db to track whether this is the first time they visit the profile page and only show it then.

## Issue

Fixes https://github.com/bluedotimpact/bluedot/issues/1079#issuecomment-3102136931

## Developer checklist

## Screenshot

https://github.com/user-attachments/assets/7148193d-e624-4466-a2ed-7ccd812f5b53


somepersonnamedjosh (approved):
I think we'd like to have user names so I am quite happy with this popping up on /account if their name is empty!

PR #1165, opened 2025-08-05
Fix ingress-nginx path validation for cert-manager compatibility

domdomegg:
## Summary
- Disable strict path validation in ingress-nginx configuration
- Resolves compatibility issues with cert-manager ACME HTTP01 challenges
- References upstream issues and documentation for context

## References
- https://cert-manager.io/docs/releases/release-notes/release-notes-1.18/#acme-http01-challenge-paths-now-use-pathtype-exact-in-ingress-routes
- https://github.com/kubernetes/ingress-nginx/issues/11176


marn-in-prod (approved):
I can confirm that we were getting ACME challenge issues and that was causing certs to fail being renewed.

PR #1187, opened 2025-08-12
[Bug] Increase timeout of loki deployment, attempting to fix flaky app/infra deployments

### LINKED ISSUE #1170: [Bug] Infra deployments are failing, SSL certs will continue to expire and not auto-renew
### 🧩 Summary

I've noticed that following Adam's recent cert manager fix merge, the infrastructure service is consistently failing to deploy.

Failed deployment example: https://github.com/bluedotimpact/bluedot/actions/runs/16766180733/job/47503297627
Note: This appears unrelated to Adam's changes, the issue seems to be preexisitng.

Suggested next steps:
Attempt a local deployment with the PULUMI_DEBUG_PROMISE_LEAKS environment variable set to true

Alternatively, modify the GitHub Actions workflow to include this environment variable for better debugging visibility on github. 

### 🛝 Steps to reproduce

1. Go to https://github.com/bluedotimpact/bluedot/actions/runs/16766180733/job/47503297627
2. Attempt to re-run all failed jobs.
3. Notice how bluedot/infra will still fail

### ✅ Definition of done

Deployment works. 

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P1" => "High"}

Will-Howard:
# Description

`infra` deployments have been failing recently, with this as the apparent main error ([example](https://github.com/bluedotimpact/bluedot/actions/runs/16909987805/job/47909624064)):
```
Diagnostics:
  kubernetes:helm.sh/v3:Release (loki):
    error: 1 error occurred:
    	* Helm release "monitoring/loki" failed to initialize completely. Use Helm CLI to investigate: failed to become available within allocated timeout. Error: Helm Release monitoring/loki: context deadline exceeded
```

There is at least one example of the deployment passing since this issue started: [example](https://github.com/bluedotimpact/bluedot/actions/runs/16918610369/job/47939344927).

I haven't spent that long looking into this, so I don't have a very good understanding of how the whole system works, but I think it's quite likely this is a basic timeout issue. I've added a timeout of 10 mins to the loki release (up from the [Helm default](https://helm.sh/docs/intro/using_helm/#:~:text=%2D%2Dtimeout%20%3A%20A%20Go%20duration,This%20defaults%20to%205m0s%20.) of 5 mins. We can see if this resolves the issue, and then I can investigate further if it keeps happening.

## Issue

Fixes #1170

### DIFF
--- a/apps/infra/src/k8s/observability.ts
+++ b/apps/infra/src/k8s/observability.ts
@@ -257,6 +257,7 @@ new k8s.helm.v3.Release('loki', {
     repo: 'https://grafana.github.io/helm-charts',
   },
   namespace: 'monitoring',
+  timeout: 600, // 10 minutes, added as a fix for https://github.com/bluedotimpact/bluedot/issues/1170
   values: {
     loki: {
       commonConfig: {


marn-in-prod (approved):
TLDR: Let's give this a go but I'm not convinced that this is the fix. Honestly, this might have been a transient issue but we'll see if the next deployment to infra fails again. If it starts to pass then it's fine for now.

Earlier today, by pure chance from my perspective, we were able to get a successfull deployment to infra. It only took 3.5 minutes though and out of the last 10-15+ attempts to deploy to infra, all of them timed out at the 6 minutes mark.

Also, the error when looking at the github deployment logs (this doesn't show for me if I deploy locally from CLI), doesn't make it sound like it's a timeout issue. 
> @ updating.................
    pulumi:pulumi:Stack infra-prod running error: update failed
    pulumi:pulumi:Stack infra-prod running The Pulumi runtime detected that 148 promises were still active
    pulumi:pulumi:Stack infra-prod running at the time that the process exited. There are a few ways that this can occur:
    pulumi:pulumi:Stack infra-prod running   * Not using `await` or `.then` on a Promise returned from a Pulumi API
    pulumi:pulumi:Stack infra-prod running   * Introducing a cyclic dependency between two Pulumi Resources
    pulumi:pulumi:Stack infra-prod running   * A bug in the Pulumi Runtime
    pulumi:pulumi:Stack infra-prod running Leaving promises active is probably not what you want. If you are unsure about
    pulumi:pulumi:Stack infra-prod running why you are seeing this message, re-run your program with the `PULUMI_DEBUG_PROMISE_LEAKS`
    pulumi:pulumi:Stack infra-prod running environment variable. The Pulumi runtime will then print out additional
    pulumi:pulumi:Stack infra-prod running debug information about the leaked promises.
    pulumi:pulumi:Stack infra-prod **failed** 1 error; 9 messages

Will-Howard:
> Let's give this a go but I'm not convinced that this is the fix

This is basically my take too, although I would say I think it's ~70% likely this is the fix.

The way I was interpreting these error logs was:
1. Deploying loki times out
2. This causes pulumi:pulumi:Stack to fail overall
3. Sometimes there are active promises when it fails, and this also gets reported, but this is a symptom not the cause. E.g. in [this example](https://github.com/bluedotimpact/bluedot/actions/runs/16909987805/job/47909624064) there is no report of open promises but otherwise the loki timeout -> pulumi:pulumi:Stack error pattern is the same

PR #1189, opened 2025-08-13
[Bug] Upgrade Kubernetes from v1.29.2+1 to v1.31.8+1

### LINKED ISSUE #1170: [Bug] Infra deployments are failing, SSL certs will continue to expire and not auto-renew

Will-Howard:
# Description

`infra` deployments are sometimes failing with this error:

```
  vultr:index:Kubernetes (vke-cluster):
    error: 1 error occurred:
    	* updating urn:pulumi:prod::infra::vultr:index/kubernetes:Kubernetes::vke-cluster: 1 error occurred:
    	* error upgrading VKE cluster 83dad468-420e-4bdb-a5da-5b5395b22bdc : gave up after 4 attempts, last error: "{\"error\":\"VKE server error: Cluster upgrade failed: invalid upgrade version\",\"status\":500}"
```

According to https://docs.vultr.com/products/kubernetes/changelog , the version we are using (v1.29.2+1) is now archived, which is sufficient to explain the above error (although not why it sometimes tries and fails to update the version, and sometimes just passes). This PR upgrades to a version that is shown as being supported.

Note that this error was seen independently of the "[loki timeout error](https://github.com/bluedotimpact/bluedot/pull/1187)". Here is an [example before](https://github.com/bluedotimpact/bluedot/actions/runs/16914901475/job/47926361326) the fix for that was merged, and an [example after](https://github.com/bluedotimpact/bluedot/actions/runs/16929666194/job/47972182482).

I haven't tested this, and I'm not sure exactly how to go about testing it. I don't think the version change is very likely to cause issues, but given the high downside risk (there could be bugs due to the new version, and we couldn't roll back to the previous version) it would be worthwhile to test. @marn-in-prod do you have thoughts on this?

## Issue

Fixes #1170

### DIFF
--- a/apps/infra/src/vultr/vke.ts
+++ b/apps/infra/src/vultr/vke.ts
@@ -4,7 +4,7 @@ import * as k8s from '@pulumi/kubernetes';
 export const k8sCluster = new vultr.Kubernetes('vke-cluster', {
   label: 'bluedot-prod',
   region: 'ams',
-  version: 'v1.29.2+1',
+  version: 'v1.31.8+1',
 
   // Initial node pool
   // nodePools: {


Will-Howard:
> [the IP addresses of your nodes will change whenever you do a VKE upgrade](https://docs.vultr.com/products/kubernetes/management/upgrade)

This is a good flag! I did also created an issue on the [pulumi-vultr](https://github.com/dirien/pulumi-vultr/issues/578) provider repo, as part of the problem is wrongly detecting a version change.

Vultr itself says the version is archived so I think this is expected behaviour on their part, and ultimately we will have to upgrade eventually, so IMO it's best to proceed with that and see the pulumi-vultr fix as nice-to-have.

My planned next steps to do that are (lmk if you think there's anything else risky I'm missing):
1. Find all cases where we use static IPs that come from Vultr
2. Depending on how critical these are and how hard it is to make them dynamic, either
  a. Migrate to using a dynamic url
  b. Be ready to swap them over quickly
3. Deploy a version upgrade (maybe [using the CLI](https://docs.vultr.com/products/kubernetes/management/upgrade) first so it's easier to monitor)

First though I'm going to focus on properly fixing the [loki timeout](https://github.com/bluedotimpact/bluedot/actions/runs/16979953341/job/48137828122), as that seems to happen almost-always whereas this only happens occasionally.

marn-in-prod:
Based on @domdomegg's input and the additional context provided (thanks for that!), I recommend proceeding with the upgrade sooner rather than later.

Given that live courses start in about a week, it's better to handle any potential downtime now rather than risk disruption during the courses. I'll notify the product team in tomorrow's meeting, after which you can begin the upgrade process.

Before starting:
Double-check the upgrade sequence carefully (there's significant downside risk if we get stuck mid-process)
Schedule this when you have a dedicated block of time available to monitor the entire process

apps/infra/src/vultr/vke.ts
```diff
@@ -4,7 +4,7 @@ import * as k8s from '@pulumi/kubernetes';
 export const k8sCluster = new vultr.Kubernetes('vke-cluster', {
   label: 'bluedot-prod',
   region: 'ams',
-  version: 'v1.29.2+1',
+  version: 'v1.31.8+1',
```
marn-in-prod, line 7: Double check the version that adam has provided, but the first one might actually need to be v1.30.12+1.

Will-Howard:
> I recommend proceeding with the upgrade sooner rather than later

Ok will do, I'll prioritise getting group switching out first and then move to this as m next priority when that's done

Will-Howard:
Closing this to preserve the discussion here. I'll create new PRs for the actual upgrade steps we go through, and future discussion can be on the issue #1266

PR #1213, opened 2025-08-19
Pin Helm chart versions in infra

domdomegg:
## Summary
Pin all Helm chart versions in infra to latest available versions to prevent automatic updates that can cause breakages.

## Why
Without pinned versions, Helm charts automatically pull the latest version during deployments, which can introduce breaking changes that require manual fixes (see #1202 and #1165 for examples).

## Changes
- ingress-nginx: 4.13.1
- cert-manager: v1.18.2  
- kube-prometheus-stack: 76.4.0
- opentelemetry-collector: 0.130.2
- loki: 6.36.1
- cloudnative-pg: 0.26.0

All versions pinned to latest available as of 2025-08-20.

### DIFF
--- a/apps/infra/src/k8s/certManager.ts
+++ b/apps/infra/src/k8s/certManager.ts
@@ -3,6 +3,7 @@ import { provider } from './provider';
 
 export const certManager = new k8s.helm.v3.Release('cert-manager', {
   chart: 'cert-manager',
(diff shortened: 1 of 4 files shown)


marn-in-prod (approved):
Thanks for this! It's definitely something that needed to get done and would reduce problems in the future

Issue #1210, opened 2025-08-18
[Feature] Group switching


anglilian:
### 🧩 Summary

UI designs for participants to switch to a new group. (mobile and desktop).

You do not need to code up any of the workflows for executing the switch. All the workflows are completed within Airtable. This interface allows participants to effectively create a record in the Group Switching table from the Course Runner base.

### ✅ Definition of done

- When users click "Can't make it?" from the discussion info box on the course unit page, it opens this modal
- When users click on the unit dropdown, it shows the list of all units for the course
- When users click the action dropdown, it shows 2 actions. (I'm deprioritising giving the option to "Join group for one unit" in this sprint)
- When users click the action dropdown, the action which has already been selected has a checkmark next to it
- When users have selected an action and a unit, the list of groups available should filter the groups based on the groups which match the "Who can switch into this group?" field and the participant's "Bucket". It should be arranged chronologically. The number of spots available is the difference between the "Max participants per group" field from the Round table and the number of participants in the group.
- If a group has equal to or more than the max participants per group, participants should not be able to select it
- When users select a group and the click "Confirm", the data is sent to the Course Runner table for Group switching with the "Request status" = Requested. The data for the participants past and new discussion and group should be logged.
- When users submit a manual request, the data is sent to the Course Runner table for Group switching with the "Manual request?" checkbox field checked and the "Request status" field = Resolve. The participant's current discussion should be logged.
- Only participants for discussions should be able to access this, _not facilitators_

### 📌 Notes / Designs

https://www.figma.com/design/6PJJZThivMuDTZ9xcoWvBs/Group-switching-UI?node-id=3037-441&t=s6zDvjAAKGInxNQz-0

### Suggested Priority Level

{"P1" => "High"}

anglilian:
Creating a record in the Group Switching table in Course runner base sets off the automation which actually switches the group or discussion for the participant. To be more specific, the record created needs to have this information:

Field name = `value`

**Switch for one group**
- Request status = `Requested`
- Switch type = `Switch group for one unit` 
- Notes from participant = `user's reason for change`
- Participant = `course registration recordID`
- New discussion = `new Group Discussion recordID`
- Old discussion = `old Group Discussion recordID`
- Unit = `discussion's Unit recordID

 **Switch group permanently**
- Request status = `Requested`
- Switch type = `Switch group permanently`
- Notes from participant = `user's reason for change`
- Participant = `course registration recordID`
- New discussion = `new Group's recordID`
[...]

Issue #1209, opened 2025-08-18
[Feature] Next discussion info


anglilian:
### 🧩 Summary

Here are the designs for the mobile and desktop components which show the user's upcoming discussion. The primary and secondary buttons change depending on how close the discussion is.

<img width="2714" height="1416" alt="Image" src="https://github.com/user-attachments/assets/32127a57-0aa1-4a37-bc32-841782cfe653" />

<img width="768" height="1370" alt="Image" src="https://github.com/user-attachments/assets/509e666f-36ef-44fd-93dc-193951ec38a6" />

The data for this component comes from the "Group discussion" table in the Course Runner base. The group discussion record should be filtered for the earliest discussion for this course that the participant is expected to attend.

The group discussion record will have fields with the links to the Slack channel, discussion doc and meets. It will also contain info on the group's name, linked record to the relevant unit, start time for the discussion.

### ✅ Definition of done

- Time until discussion updates dynamically while the page is open
- Clicking the button "Join discussion" opens the meeting link in a new tab
- Clicking the button "Prepare for discussion" navigates the user to the corresponding unit
- Clicking the button "Open discussion doc" opens the discussion doc in a new tab
- Clicking the button "Message your group" opens the group's Slack channel
- The time is displayed in the user's timezone
- The buttons change if the discussion is <1 hour from starting
- The component is only visible to participants on an active course round (note: I have not prioritised making the facilitator version of this UI)

### 📌 Notes / Designs

https://www.figma.com/design/6PJJZThivMuDTZ9xcoWvBs/Group-switching-UI?node-id=3003-19436&t=s6zDvjAAKGInxNQz-0

### Suggested Priority Level

{"P1" => "High"}

marn-in-prod:
FYI, we're also planning to add "number of minutes" until the discussion starts and not just the hours.

PR #1220, opened 2025-08-21
[Feature] Schema changes to support group discussion and group switching UI

Will-Howard:
# Description

These are all the schema changes that I believe are required for adding [Next discussion info](https://github.com/bluedotimpact/bluedot/issues/1209) and [Group switching](https://github.com/bluedotimpact/bluedot/issues/1210). I'm putting these out now so they are ready to use in development.

I'll shortly put up another PR which sets up the scaffolding for displaying group discussions, and then we should be able to work on individual features (buttons in [this Figma](https://www.figma.com/design/6PJJZThivMuDTZ9xcoWvBs/Group-switching-UI?node-id=3037-441&t=s6zDvjAAKGInxNQz-0)) in parallel.

## Issue

#1209, #1210

## Testing

I have tested this on a local instance of Postgres + pg-sync-service, the fields do come through and I was able to select on them in #1221

### DIFF
--- a/libraries/db/src/schema.ts
+++ b/libraries/db/src/schema.ts
@@ -313,7 +313,7 @@ export const groupDiscussionTable = pgAirtable('group_discussion', {
       airtableId: 'fldRV2aVcMiNZMViJ',
     },
     unitNumber: {
-      pgColumn: numeric({ mode: 'number' }),
+      pgColumn: text(),
       airtableId: 'fldbNYACt7S5J2QlU',
     },
     zoomLink: {
--- a/libraries/db/src/schema.ts
+++ b/libraries/db/src/schema.ts
@@ -336,7 +336,7 @@ export const groupSwitchingTable = pgAirtable('group_switching', {
   tableId: 'tblGCHwcMcDrl57OY',
   columns: {
     participant: {
-      pgColumn: text().notNull(),
+      pgColumn: text(),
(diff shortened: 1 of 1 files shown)


libraries/db/src/schema.ts
```diff
@@ -304,6 +308,73 @@ export const groupDiscussionTable = pgAirtable('group_discussion', {
       pgColumn: text(),
       airtableId: 'fldH0pKnEELPI65Qs',
     },
+    courseSite: {
```
Will-Howard, line 311: `courseSite` and `participantEmailsExpected` are used in #1221 to look up the relevant group discussion, based on the user and course slug (matching on "courseSite ~ courseSlug/?$").

I think ideally we would add courseSlug as a field on `groupDiscussionTable`, and then match on that. I decided not do this to prioritise unblocking other work on this feature (to not risk getting bogged down in errors due to adding the field to airtable). @marn-in-prod lmk if you think this is worth changing I can come back and change this

marn-in-prod: To add a courseSlug field on Airtable, create a new formula field and reference the Course Site Url while applying Airtable's Find/left/right functions.

I don't think it matters all that much TBH. By the time the performance becomes a problem, we'll have much bigger issues.

marn-in-prod (approved):
It looks good. I didn't confirm the actual AirtableIds myself though. 

I think doing this in a seperate PR btw was great

PR #1233, opened 2025-08-23
[Feature] Support group switching with v0 UI

Will-Howard:
# Description

Summary:
- New endpoint `group-switching/available.ts` to fetch the groups and discussions that are available to switch into
- New endpoint `group-switching/index.ts` for submitting switch requests
- The modal itself, `GroupSwitchModal`
- A bunch of schema changes to support this

These are the [designs](https://www.figma.com/design/6PJJZThivMuDTZ9xcoWvBs/Group-switching-UI?node-id=3066-1891&t=U7nnbjPTr0OHDICS-0). This PR doesn't match them that closely, I'm optimising for having a usable version with more basic styling out first, and will then come back and make the UI match the designs exactly

### Testing pointers

I've been testing with my own user against the "What the fish?" test course. To do this yourself (not 100% sure this is correct):
1. Visit the course page (`/courses/fish-test-course/1`) to ensure there is a "Course registration" for this course and your user
2. Add this "Course registration" as a participant in one of the [groups](https://airtable.com/appPs3sb9BrYZN69z/tblyiJSPoniwhi17T) for this course
3. Add this "Course registration" as a participant (expected) in one of the [group discussions](https://airtable.com/appPs3sb9BrYZN69z/tblDNME0bA9OoApTk) for this course

## Issue

#1210 

## Developer checklist

- [x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [ ] Considered having snapshot tests and/or happy path functional tests
- [ ] Considered adding Storybook stories

## Screenshot

| 📸 |  |
|---------|---|
| 🖥️ | Screenshot 2025-08-25 at 19 25 23 |
| | Screenshot 2025-08-25 at 19 25 38 |
| | Screenshot 2025-08-25 at 19 25 43 |
| | Screenshot 2025-08-25 at 19 34 03 |
| | Screenshot 2025-08-25 at 19 34 25 |
| | Screenshot 2025-08-25 at 19 35 19 |
| | Screenshot 2025-08-25 at 19 35 26 |
| | Screenshot 2025-08-25 at 19 35 51 |
| 📱  | Screenshot 2025-08-25 at 19 37 24 |
| | Otherwise as above, the mobile styles currently don't look great but the modal is usable |

PR #1238, opened 2025-08-25
[Feature] More schema changes to support group switching

Will-Howard:
# Description
Schema changes required for #1233 . To be merged first to reduce the risk of code/database mismatch.

## Issue
#1210

## Developer checklist

N/A

## Screenshot
No UI change

### DIFF
--- a/libraries/db/src/schema.ts
+++ b/libraries/db/src/schema.ts
@@ -262,7 +262,7 @@ export const groupTable = pgAirtable('group', {
   tableId: 'tblyiJSPoniwhi17T',
   columns: {
     groupName: {
-      pgColumn: text().notNull(),
+      pgColumn: text(),
       airtableId: 'fldv3jHyWGjR0LxLp',
     },
     groupDiscussions: {


libraries/db/src/schema.ts
```diff
@@ -378,6 +394,21 @@ export const groupSwitchingTable = pgAirtable('group_switching', {
   },
 });
 
+export const courseRunnerBucketTable = pgAirtable('course_runner_bucket', {
+  baseId: COURSE_RUNNER_BASE_ID,
+  tableId: 'tbl7Pevw79fDA7EmX',
+  columns: {
+    groups: {
+      pgColumn: text().array().notNull(),
```
marn-in-prod, line 402: Are you positive that you want to put notNull()? In general, Airtable and Airtable-ts is very finicky with this. 

I think Airtable-ts will treat empty linked rows as undefined which might end up breaking the code if you rely on it being notNull. 

It's fine if you do, but just calling it out that it's a footgun IMO and for not much benefit since you probably need to be defensive anyways.

libraries/db/src/schema.ts
```diff
@@ -269,6 +273,14 @@ export const groupTable = pgAirtable('group', {
...
       airtableId: 'fldtzy3nSP0piVApO',
     },
+    participants: {
+      pgColumn: text().array().notNull(),
+      airtableId: 'fldcEa25oCDAmgDqm',
+    },
+    whoCanSwitchIntoThisGroup: {
+      pgColumn: text().array().notNull(),
+      airtableId: 'fldVQihgKyx6nJIR5',
+    },
```
marn-in-prod, line 283: Same comment about not null. Mostly just a warning IMO

marn-in-prod (approved):
LGTM, just read the comments that I left

libraries/db/src/schema.ts
Will-Howard: This is a good point, I've changed all the ones that were non-arrays with `notNull()` applied. For these array fields, it does already insert an empty array if it gets an empty value, so I'm leaving these ones as is

Issue #1217, opened 2025-08-20
Implement updated design for profile page


anglilian:
### 🧩 Summary

Update design for profile page to be more simple and allow participants to access the group switching modal.

### ✅ Definition of done

- Users can see the list of upcoming discussions
- Users can click "Switch group" to open the group switching modal 
- The group switching modal prefills the Action = "Switch group for one unit" and the Unit with the corresponding unit
- Users can click "Join discussion" to open the meets link in a new tab <1 hour before the discussion starts up until the end of the discussion
- Users can click "Prepare for discussion" for the upcoming discussion >1 hour before discussion starts which will take them to the unit's first chunk
- The time counting down to the start of the discussion updates on a minute-by-minute basis.
- Courses tab is now called enrolled courses
- Completed courses don't expand
- Clicking on the box (instead of ONLY the arrow) expands the course
- Removed drop out courses section

### 📌 Notes / Designs

- [Loom walkthrough](https://www.loom.com/share/4dcdb6e7964c4d4fad3384e97391435f)
- [Figma design](https://www.figma.com/design/6PJJZThivMuDTZ9xcoWvBs/Group-switching-UI?node-id=3104-4090&t=BWo8ZQq1xcgOEnrO-1)

### Suggested Priority Level

{"P2" => "Medium (Default)"}

somepersonnamedjosh:
This needs to address the registrations/applications view problem from https://github.com/bluedotimpact/bluedot/issues/1169 as well! Commenting to track this.

PR #1247, opened 2025-08-26
Filter courses by status instead of displayOnCourseHubIndex in the API endpoint

anglilian:
- Add status field to course table schema
- Update API to filter courses by 'Active' status
- Move displayOnCourseHubIndex filtering to frontend hook

# Description

This endpoint is shared by the profile page so the filter shouldn't be hardcoded here. This is blocking me from testing #1217 .

## Issue

Fixes #

## Developer checklist

## Screenshot

Can confirm it filters properly!

CleanShot 2025-08-26 at 15 59 00@2x
CleanShot 2025-08-26 at 15 59 45@2x

| 📸 |  |
|---------|---|
| 📱  |  |
| 🖥️ |  |

### DIFF
--- a/apps/website/src/components/courses/CourseDirectory.stories.tsx
+++ b/apps/website/src/components/courses/CourseDirectory.stories.tsx
@@ -38,6 +38,7 @@ export const Default: Story = {
         publicLastUpdated: null,
         isNew: true,
         isFeatured: false,
+        status: 'Active',
       },
       {
         id: '2',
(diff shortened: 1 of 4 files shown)


marn-in-prod:
Oh also, make sure to fix the storybook typing issue here (which is causing the build to fail).

./src/components/courses/CourseDirectory.stories.tsx:19:5

anglilian:
> Look good! Just confirm locally that you've got the right airtableId and that you can still see courses.

I can confirm I've got the right ID and I can see the courses.

CleanShot 2025-08-26 at 23 45 51

I've also just made the storybook fix you asked for!

PR #1253, opened 2025-08-27
Add number key navigation for units in course layout

dewierwan:
# Description

- Add keyboard shortcuts 1-9 to jump directly to specific units
- Update keyboard navigation hint to include new shortcuts
- Provide accessibility announcement when navigating via number keys

I want to navigate units faster. This is the most intuitive way to do it in my mind. 

It takes the user to the first chunk of the unit number they select. E.g. if they're currently on unit 3, chunk 4, and they press the '2' key, it will take them to the first chunk in unit 2.

### DIFF
--- a/apps/website/src/components/courses/UnitLayout.test.tsx
+++ b/apps/website/src/components/courses/UnitLayout.test.tsx
@@ -251,7 +251,7 @@ describe('UnitLayout', () => {
 
     const hint = container.querySelector('.unit__keyboard-hint');
     expect(hint).toBeTruthy();
-    expect(hint?.textContent).toContain('Use ←/→ to navigate sections');
+    expect(hint?.textContent).toContain('Use ←/→ or 1-9 to navigate');
   });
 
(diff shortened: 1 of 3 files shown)


dewierwan:
@claude please review

apps/website/src/components/courses/UnitLayout.test.tsx
```diff
@@ -251,7 +251,7 @@ describe('UnitLayout', () => {
 
     const hint = container.querySelector('.unit__keyboard-hint');
     expect(hint).toBeTruthy();
-    expect(hint?.textContent).toContain('Use ←/→ to navigate sections');
+    expect(hint?.textContent).toContain('Use ←/→ or 1-9 to navigate');
```
marn-in-prod, line 254: If I understand correctly, pressing 1-9 sets the user's unit, while [navigation keys/arrow keys] allow the user to navigate between chunks and units. 

Maybe we should modify this line to specify that pressing 1-9 will move the user directly to that unit since they don't technically do the same thing

PR #1261, opened 2025-08-29
[Fix] Disabled CTA button styling

joshestein:
# Description

Currently if you set the `disabled` attribute on a `CTALinkOrButton` that is a button there is no UI to indicate that the button is disabled. For consistency with how we disable links, we should (1) reduce opacity to 50% and (2) disallow pointer events.

1. Introduce `disabled:opacity-50` and `disabled:pointer-events-none` base CTA styles (this only affects buttons).
2. Introduce `aria-disabled:opacity-50` and `aria-disabled:pointer-events-none` base CTA styles (this only affects links). Additionally remove the currently over-riding `styles={}` attribute that does the same thing.

## Developer checklist

## Screenshot

image


marn-in-prod (approved):
LGTM. Was there a specific feature where this was needed for? I can't immediately recall where we use the disable button.

PR #1263, opened 2025-08-30
[Feat] Simplifying the keyboard shortcut popup

dewierwan:
# Code updates
- Consolidate navigation shortcuts into single "Navigate sections" entry
- Remove popover title by default for cleaner UI
- Update dialog aria-label for better accessibility
- Fix test assertions to match new menu structure

# Description
I felt the new keyboard shortcuts design was wordy, so I've simplified it. 

Users don't know what "Units" and "Chunks" are, so I don't think we should use these words publicly. Sections is more intuitive imo.

"Inside course" doesn't mean anything to me, so I deleted it too. 

# Screenshot
| **Old** | **New** |
| --- | ---- | 
| CleanShot 2025-08-30 at 14 29 27@2x | CleanShot 2025-08-30 at 14 29 03@2x |

### DIFF
--- a/apps/website/next-env.d.ts
+++ b/apps/website/next-env.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference path="./dist/types/routes.d.ts" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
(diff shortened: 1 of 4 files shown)


dewierwan:
@claude please review

dewierwan:
FYI @joshestein , I made this small tweak to the keyboard shortcut thing!

joshestein (approved):
LGTM 🥳 The original wording/implementation came from the design spec - I definitely agree this is cleaner!

PR #1267, opened 2025-09-01
[Bug] Upgrade vke-cluster to the version that is actually live (1.32.1+1)

Will-Howard:
# Description

From the issue #1266 :

>The version shown in the [Vultr web UI](https://my.vultr.com/kubernetes/manage/?id=83dad468-420e-4bdb-a5da-5b5395b22bdc#subsupgrade) is actually 1.32.1+1, and running `kubectl version` locally confirms this (client/server mismatch is just a local issue):
>```
>> kubectl version
>Client Version: v1.34.0
>Kustomize Version: v5.7.1
>Server Version: v1.32.1
>Warning: version difference between client (1.34) and server (1.32) exceeds the supported minor version skew of +/-1
>```

## Issue

#1266

## Developer checklist

N/A

## Steps to test

On master, running `pulumi preview --diff --show-config` should give an output that looks like this:
```
pulumi preview --diff --show-config
Previewing update (prod):
Configuration:
    infra:airtablePat: [secret]
    infra:alertsSlackBotToken: [secret]
    infra:anthropicApiKey: [secret]
    infra:backupGcsKey: [secret]
    infra:keycloakClientId: [secret]
    infra:keycloakClientSecret: [secret]
    infra:loginProxyKeycloakClientSecret: [secret]
    infra:loginProxySharedSecret: [secret]
    infra:meetZoomClientSecret: [secret]
    infra:minioRootPassword: [secret]
    infra:openaiApiKey: [secret]
    infra:prodOnlyWebhookDeletion: [secret]
    infra:roomDisplayBearerToken: [secret]
    pulumi:disable-default-providers: ["kubernetes","minio"]
    vultr:apiKey: [secret]
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:prod::infra::pulumi:pulumi:Stack::infra-prod]
    ~ vultr:index/kubernetes:Kubernetes: (update)
        [id=THE_ACTUAL_ID]
        [urn=urn:pulumi:prod::infra::vultr:index/kubernetes:Kubernetes::vke-cluster]
        [provider=urn:pulumi:prod::infra::pulumi:providers:vultr::default_2_20_1_github_/api.github.com/dirien/pulumi-vultr::ANOTHER_ACTUAL_ID]
      ~ version: "v1.32.1+1" => "v1.29.2+1"
Resources:
    ~ 1 to update
    104 unchanged
```
And on this branch:
```
pulumi preview --diff --show-config
Previewing update (prod):
Configuration:
[...]

PR #1268, opened 2025-09-01
[Bug] Upgrade vke-cluster to latest version (v1.33.0+3)

### LINKED ISSUE #1266: [Bug] Upgrade Kubernetes to a supported version
### 🧩 Summary

The Kubernetes version is currently set to `v1.29.2+1` in the code (in 'vke-cluster' in `apps/infra/src/vultr/vke.ts`). This is ["Archived" by Vultr](https://docs.vultr.com/products/kubernetes/changelog), resulting in the 'vke-cluster' service failing to deploy.

This could eventually result in downtime if we need to deploy the service for some other reason, or it blocks deploying another service, or the version gets so out of date that things start to break because of that.

The version shown in the [Vultr web UI](https://my.vultr.com/kubernetes/manage/?id=83dad468-420e-4bdb-a5da-5b5395b22bdc#subsupgrade) is actually 1.32.1+1, and running `kubectl version` locally confirms this (client/server mismatch is just a local issue):
```
> kubectl version
Client Version: v1.34.0
Kustomize Version: v5.7.1
Server Version: v1.32.1
Warning: version difference between client (1.34) and server (1.32) exceeds the supported minor version skew of +/-1
```

There is some further discussion in [this PR](https://github.com/bluedotimpact/bluedot/pull/1189), particularly on the risks due to node IPs changing.

### 🛝 Steps to reproduce

Since [the PR to pin Helm chart versions](https://github.com/bluedotimpact/bluedot/pull/1213), the service doesn't actually attempt to deploy when we merge into master.

You can see old examples of deployments that failed [here](https://github.com/bluedotimpact/bluedot/actions/runs/16914901475/job/47926361326) and [here](https://github.com/bluedotimpact/bluedot/actions/runs/16929666194/job/47972182482). Retrying those should result in the same failure, although I don't recommend doing that because I don't know the implications of retrying actions from weeks ago.

### ✅ Definition of done

1. Deploying the 'vke-cluster' service succeeds
2. We have upgraded to Kubernetes v1.33.0+1
  a. Minimum to close this: We have upgraded to a version that is supported by Vultr (based on the list [here](https://docs.vultr.com/products/kubernetes/changelog))

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

Will-Howard:
# Description

I won't merge this until after [Upgrade vke-cluster to the version that is actually live (1.32.1+1)](https://github.com/bluedotimpact/bluedot/pull/1267) is deployed. There is some risk with this PR of the load balancer IP changing (this is the main entry point into the website and production database), copying [this comment](https://github.com/bluedotimpact/bluedot/issues/1266#issuecomment-3241563281) from the issue regarding that:

> An alternative would be to first update the load balancer to definitely use a static IP, and then proceed with the upgrade. The reasons I think it not worth doing it this way are:
>
>1. I think it's very unlikely that the load balancer IP will change. Reasoning:
>  a. It is an external resource created by the 'ingress-nginx' service, not a Kubernetes node itself, so it shouldn't be recreated during an upgrade
>  b. The fact that we are currently on v1.32.1+1 and not v1.29.2+1 means that Vultr has probably forced an upgrade itself, and this didn't cause problems
>2. Switching to a static IP would require first provisioning a new [reserved IP](https://docs.vultr.com/products/network/reserved-ips/provisioning), then updating the load balancer to use that, then doing the Kubernetes upgrade. This would itself require switching DNS settings and come with some risk of downtime

## Issue

Fixes #1266

## Developer checklist

N/A

### DIFF
--- a/apps/infra/src/vultr/vke.ts
+++ b/apps/infra/src/vultr/vke.ts
@@ -4,7 +4,7 @@ import * as k8s from '@pulumi/kubernetes';
 export const k8sCluster = new vultr.Kubernetes('vke-cluster', {
   label: 'bluedot-prod',
   region: 'ams',
-  version: 'v1.32.1+1',
+  version: 'v1.33.0+1',
 
   // Initial node pool
   // nodePools: {
--- a/apps/infra/src/vultr/vke.ts
+++ b/apps/infra/src/vultr/vke.ts
@@ -4,7 +4,7 @@ import * as k8s from '@pulumi/kubernetes';
 export const k8sCluster = new vultr.Kubernetes('vke-cluster', {
   label: 'bluedot-prod',
   region: 'ams',
-  version: 'v1.33.0+1',
+  version: 'v1.33.0+3',
 
   // Initial node pool
   // nodePools: {


marn-in-prod (commented):
The vultr website shows v1.33.0+3 as the current version.

 Also see my comment https://github.com/bluedotimpact/bluedot/issues/1266#issuecomment-3243518798

Will-Howard:
Upgrade through the UI is complete, now merging this to get the code in sync

PR #1276, opened 2025-09-02
[feat] Adding a filter for chunks, so only "Active" chunks are visible. 

dewierwan:
# Description
Right now, the only way to hide a chunk is to delete it, or detach it from the unit. I want to have some chunks be "inactive" so I can work on them while they're in the right place on the back-end.


dewierwan:
@claude please review

Issue #1154, opened 2025-07-31
[Feature] Update bottom section of course page


herrhaase:
### 🧩 Summary

New design for the bottom section

### ✅ Definition of done

1. Centered "Next: XYZ" button with the title or type of the next chunk or unit 
2. Shortcuts popup #1150 
3. Feedback option where the label should pick up on the type of page the user is looking at (unit, exercise, reflection, etc.) – related to feedback option from resource cards #1147 
4. Bonus: Bottom gradient 

### 📌 Notes / Designs

[Designs](https://www.figma.com/design/62YlFNK7QS6z7SkrPjrwVb/Blue-Dot?node-id=929-5173&m=dev&t=Cfx5dK5pDpuK4rUZ-1)

Bottom:
<img width="1736" height="572" alt="Image" src="https://github.com/user-attachments/assets/09624c74-a31a-4f5c-8c48-ee5e16102fa2" />

Feedback:
<img width="1614" height="1614" alt="Image" src="https://github.com/user-attachments/assets/e6fecd77-15da-4b5c-b734-58e3e74fd77e" />

Gradient:
https://www.figma.com/design/62YlFNK7QS6z7SkrPjrwVb/Blue-Dot?node-id=929-5170&m=dev&t=Cfx5dK5pDpuK4rUZ-1

### Suggested Priority Level

{"P2" => "Medium (Default)"}

marn-in-prod:
Looking into this - currently we don't have a relationship in Airtable that represents a user like/disliking a unit. So this would need to be added before the feature could be done. 

We previously had an "overall rating" field in the Unit feedback table. Where was this table used before? I'm thinking we would just add a new field here that would represent it. @dewierwan @somepersonnamedjosh 

https://airtable.com/appbiNKDcn1sGPGOG/tblBwjMjul1c6l7ea/viwCSHmaJUXh0yX8Y?blocks=hide

marn-in-prod:
### Scoping
This is E2E for feature that I don't think would have any blockers. 

### Implementation Steps

1. **Database Schema Update**
   - Create and merge a PR with the updated database schema
   - Include the new column created by Dewi (see earlier comment)

2. **Backend Endpoint**
   - Add a new endpoint to upsert the database row for the specific user
   - Note: This will repurpose the old unit feedback table (inserting rows might need to take those older columns into account)
   - The previous UI for user submissions has been deleted
   - Check Dewi's commit history around the 23rd to find the removal PR
[...]

PR #1278, opened 2025-09-02
[Feat] 'ghost' variant for `CTALinkOrButton`

joshestein:
# Description

For #1154 we need ghost buttons - both for the keyboard shortcuts component and the like/dislike buttons.

image

Although we could continue to duplicate styles across components I feel like it's cleaner to have a single source of truth.

I tried to replace the current 'react-aria-component' `<Button>` used within the `KeyboardNavMenu` component with this new ghost variant. Unfortunately that led to some complexity, because the `Popover` requires a [custom trigger](https://react-spectrum.adobe.com/react-aria/Popover.html#custom-trigger), which in turn requires forwarding a React ref. This made the PR much more messy, so I decided to exclude these changes for now.

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| Normal  | image |
| Hover | image |

### DIFF
--- a/libraries/ui/src/CTALinkOrButton.test.tsx
+++ b/libraries/ui/src/CTALinkOrButton.test.tsx
@@ -69,6 +69,16 @@ describe('CTALinkOrButton', () => {
     expect(button.className).includes('text-black');
   });
 
+  test('renders with ghost variant', () => {
+    render(<CTALinkOrButton variant="ghost">Click me</CTALinkOrButton>);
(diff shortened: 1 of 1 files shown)


joshestein:
@claude review

somepersonnamedjosh (approved):
lgtm - we should consider the gap/chevron tweaks in follow-up, will leave you to do that @joshestein

PR #1269, opened 2025-09-01
[Feat] Update quiz component UI

joshestein:
# Description

Closes #1153.

1. Updates quiz component styling
2. Updates `MultipleChoice` component to not use any BEM classes, and updates tests appropriately
3. Simplifies `MultipleChoice` state by using `watch('answer')` of `useForm` instead of tracking `selectedOption`
4. Adds new tests for checking quiz options are disabled/enabled, 'try again' button resetting state
5. Add `@testing-library/jest-dom` and `@testing-library/user-event` to dev dependencies, remove (now) outdated ESLint disable lines
6. Moves the 'Correct! Quiz completed. 🎉' feedback _outside_ the form element
7. Extract `BASE_LABEL_STYLES` on `<Input>` component, which handles disabled states by disabling cursor

## Developer checklist

- [x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [x] Considered having snapshot tests and/or happy path functional tests
- [x] Considered adding Storybook stories

## Screenshot

| 📸 |  |
|---------|---|
| 📱  | image image image image image |
| 🖥️ | image image image image image |

PR #1281, opened 2025-09-03
`cn` class merge utility

joshestein:
# Description

We don't have a way to ensure that overriding styles we pass via `className` are respected.

Consider the `CTALinkOrButton`, which has a 'primary' `bg-bluedot-normal`. If a caller passes a `bg-[#mycustomcolor]` via the `className` prop both classes are present in the DOM. Since Tailwind generates CSS with similar specificity for these background utilities, the order they appear in  the compiled stylesheet (not the order in your className string) determines which one actually applies - leading to unpredictable results. This means that callers need to [use the `!important` specifier,](https://github.com/bluedotimpact/bluedot/pull/1269/files#diff-9991d16143d4292f8d7d0ab1980bdfb984bca5a612ef6c4e21e26b12c42b950eR144) which is not ideal.

This `cn()` utility solves this by detecting conflicting Tailwind classes and ensuring only the last one (your custom colour override) is included in the final output. This a fairly common pattern -[ ShadCN uses it extensively.](https://ui.shadcn.com/docs/installation/manual#add-a-cn-helper)

### DIFF
--- a/libraries/ui/package.json
+++ b/libraries/ui/package.json
@@ -31,6 +31,7 @@
     "react-device-detect": "^2.2.3",
     "react-dom": "^18.2.0",
     "react-icons": "^4.3.1",
+    "tailwind-merge": "^3.3.1",
     "winston": "^3.17.0",
     "winston-console-format": "^1.0.8",
     "zod": "^3.24.2",
--- a/libraries/ui/src/CTALinkOrButton.stories.tsx
+++ b/libraries/ui/src/CTALinkOrButton.stories.tsx
@@ -178,3 +178,11 @@ export const DisabledLink: Story = {
     disabled: true,
   },
 };
+
+export const CustomStyles: Story = {
+  args: {
+    children: 'Custom Styles',
+    variant: 'primary',
+    className: 'bg-[#2244BB] hover:bg-[color-mix(in_oklab,#2244BB,#000_30%)] text-white hover:text-white',
+  },
+};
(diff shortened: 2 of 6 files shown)


marn-in-prod (approved):
Generally looks good, but I had a few questions:
Was there a specific reason that made you want to look into this? Like, were you experiencing a particular problem that led to this change?

Also, could you confirm that the other components still look the same with this change

One more thing, more of an aside, I'm not actually a fan of how often we're using barrel index files in the codebase. At a high level, I don't think calling something "Index.ts" helps with discoverability (the backend api routes for example which I think are a really frustrating example). 

I know you didn't create the file itself, just added to it, so it's not something you need to change yourself and more of an FYI.

joshestein:
Yes, the particular problem was that I kept trying to over-ride `CTALinkOrButton` styles (in line with new designs) and would run into issues where the new styles I applied via the `className` prop wouldn't be respected (as a result of CSS specificity problems). Then you run into a problem of manually needing to figure out which styles are / aren't being applied and applying specificity modifiers. Would be less painful if I know that whatever I pass via `className` will always be respected, and I don't have to worry about any modifiers.

From the tests, the components that are affects are: `CourseSection`, `ProjectSubSection`, `Nav`, `CourseListRow` and `CourseCard`. I've checked them and they all look the same. The `CTALinkOrButton` also all look the same. 

I hear you about the barrel indexing 👍 Should I extract this to a new file, or just something to keep in mind for the future?

Issue #1293, opened 2025-09-06
[Feature] pg-sync-service should send alerts to slack


marn-in-prod:
### 🧩 Summary

Pg-sync-service is an important service and currently the only way of diagnosing when issues are occurring are either through an engineer manually looking at Grafana logs, or when somebody notices that Airtable changes are not syncing to the main website. Somebody who isn't looking at Grafana currently has no insights on when a sync starts & ends making it hard for them to know when they should escalate a sync issue.

There's a whole slew of different issues that could be the reason for why pg-sync-service is stuck/down etc and I believe that having detailed and accurate notifications sent to slack can not only improve experience (making it less opaque to vibe coders who are contributing) but also reduce how long an de-sync events occur. 

### ✅ Definition of done

Any errors that occur from webhooks not being able to be created / missing fields or rows not being able to be processed should have detailed logging and error messaging sent to slack. 

In addition to sending error notifications to the slack channel, I think that whenever prod starts and ends a full-sync, we should also send notifications. That way it'll be easier for the team to track when and if a sync has completed making it so that non-developers would know exactly when they should expect the main website to have updated data.

Some examples of how the webhook processing can be improved is the following which I already happened to have had from various debugging sessions for pg-sync-service. I'm not saying that it's "production-ready" but rather just examples of the types of detailed logging that could be really useful. I'd expect that whoever implements this fix should attempt to manually trigger these errors to make sure that whatever notifications that get sent to slack are precise and helpful by themselves without needing to look at grafana for additional logs.

````
private async createWebhook(): Promise<void> {
    logger.info(`[AirtableWebhook] Creating new webhook for base ${this.baseId} with ${this.fieldIds.length} field filters`);

    const webhookSpec: unknown = {
      specification: {
        options: {
          filters: {
            dataTypes: ['tableData', 'tableFields'],
            ...(this.fieldIds.length > 0 ? { watchDataInFieldIds: this.fieldIds } : {}),
          },
        },
      },
    };

    await this.rateLimiter.acquire();
    
    try {
      const createResponse = await this.axiosInstance.post<{ id: string; cursorForNextPayload: number }>(
        `/bases/${this.baseId}/webhooks`,
        webhookSpec,
      );

      this.webhookId = createResponse.data.id;
      this.nextPayloadCursor = createResponse.data.cursorForNextPayload;
      logger.info(`[AirtableWebhook] Created webhook ${this.webhookId} for base ${this.baseId}`);
    } catch (error: any) {
[...]

PR #1318, opened 2025-09-10
More explicit website DB error information

joshestein:
# Description

When we have DB errors we just throw a generic `InternalServerError`. To provide better transparency around what went wrong we now pass the error message along too.

Also address some minor error separation between authorization errors and actual bad requests.

## Issue

Part of #1293.

### DIFF
--- a/apps/website/src/pages/api/courses/exercises/[exerciseId]/response.ts
+++ b/apps/website/src/pages/api/courses/exercises/[exerciseId]/response.ts
@@ -31,8 +31,12 @@ export default makeApiRoute({
 }, async (body, { raw, auth }) => {
   const { exerciseId } = raw.req.query;
 
-  if (!auth.email || typeof exerciseId !== 'string') {
-    throw new createHttpError.BadRequest();
+  if (!auth.email) {
+    throw new createHttpError.Unauthorized('Authentication required');
+  }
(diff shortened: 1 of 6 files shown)


apps/website/src/pages/api/courses/exercises/[exerciseId]/response.ts
```diff
@@ -31,8 +31,12 @@ export default makeApiRoute({
 }, async (body, { raw, auth }) => {
   const { exerciseId } = raw.req.query;
 
-  if (!auth.email || typeof exerciseId !== 'string') {
-    throw new createHttpError.BadRequest();
+  if (!auth.email) {
+    throw new createHttpError.Unauthorized('Authentication required');
+  }
```
Comment (marn-in-prod) [inline apps/website/src/pages/api/courses/exercises/[exerciseId]/response.ts:36]: Isn't not having auth impossible if requireAuth is set to true on the route? I know you didn't add this but just feels like it shouldn't be relevant anymore?

marn-in-prod (approved):
Great job on taking initiative for this. 

I generally want to err on the side of caution for not actually providing database error messages directly in prod because I'm not confident that there isn't some attack vector to be able to figure out people's email addresses or some other PII that we get in the future.

In local development, I'm 100% good with these changes though.

PR #1331, opened 2025-09-12
[Feat] sitemaps

### LINKED ISSUE #250: SiteMaps XML
1. Create new sitemap
2. Upload new sitemap to bluedot.org
3. Submit sitemap URL to Google Search Console

joshestein:
# Description

Sitemaps are used by search engines to improve SEO. After implementing [I found this](https://developers.google.com/search/docs/crawling-indexing/sitemaps/overview#do-i-need-a-sitemap), which maybe means we don't actually need to do this? Or at least not for the statically generated map?

Changes:

1. New statically generated `sitemap.xml` for our most important routes (e.g. '/', '/about', '/contact', '/courses', etc.). This  updates on new builds.
2. New dynamically generated `sitemap.xml` for courses, blogs and projects. These are constructed and returned per-request, which means they will always be up-to-date with CMS/DB.
3. Add `AIRTABLE_PERSONAL_ACCESS_TOKEN` to GH secrets, and use that same variable during the build process (needed since we need to fetch courses/blogs/projects from Airtable)

`<lastmod>`, `<changefreq>` and `<priority>` tag definitions [are all optional.](https://www.sitemaps.org/protocol.html#xmlTagDefinitions) For priority, I arbitrarily chose '/' (home) as 1.0, first level routes (e.g. /blog) as 0.8 and third level routes (e.g. /blog/my-awesome-post) as 0.7. Courses don't seem to have a `publishedAt` so I've excluded `lastmod` from their sitemap file.
 
## Issue

Fixes #250.

## Developer checklist

NA

## Screenshot

NA


marn-in-prod:
Mentioned this in slack DMs, but the reason for ci_cd to be failing is because github actions has a different set of env variables than production (you can see the variables it has access to through the settings tab on this repo)

apps/website/src/pages/sitemap-generator.ts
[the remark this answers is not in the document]
joshestein: Will move outside. The root sitemap.xml is statically generated at build time and placed in `/public/sitemap.xml` within sitemap-generator.ts.

joshestein:
> Mentioned this in slack DMs, but the reason for ci_cd to be failing is because github actions has a different set of env variables than production (you can see the variables it has access to through the settings tab on this repo)

Updates the repo vars to include `AIRTABLE_PERSONAL_ACCESS_TOKEN` - I just used the one I generated during onboarding. CI/CD seems to be passing 🥳

somepersonnamedjosh:
@claude please review :D

joshestein:
@claude are you just reading CodeRabbit's nonsensical comments and then reforming them as your own thoughts? Tsk tsk. Let me respond to each appropriately:

1. We are purposefully using `sed` to inject secrets. I tried directly sourcing the secret using `env: AIRTABLE_TOKEN: {{ secrets.AIRTABLE_TOKEN }}` but because `.env.local` has higher precedence than the env variable the value in the file was still being used. Then I tried appending the secret to the file, hoping that if there are two entries the last one would be used but this didn't work. Eventually I tried injecting using `sed` which I agree is not very pleasant but it did work. Are there alternatives you would suggest?
2. I will push a commit that uses fallback, but we don't have a current env variable that specifies the URL, so this is just wasteful / strange to future readers who won't understand where the variable is supposed to be set. I also don't think we need sitemaps for staging/preview environments.
3. What are you talking about? `execFileSync` is 'generally identical to execFile' (from the [docs](https://nodejs.org/api/child_process.html#child_processexecfilesyncfile-args-options)).  We are not exposing this function to end users, no one is going to be able to inject malicious strings. I will push a commit making the change but feel like is really not necessary.
4.  `publishedAt` is:  pgColumn: numeric({ mode: 'number' }).notNull() - it cannot be null, and TS shows this is a number, not `number | null`
5. https://developers.google.com/search/docs/crawling-indexing/sitemaps/large-sitemaps gives an example with a `sitemap_index.html` file. What circular reference are you referring to?
6. You are right, I will update.
7. I'm not sure what you're talking about.
8. Cache headers were added in https://github.com/bluedotimpact/bluedot/pull/1331/commits/ce3c60f90e2f0094dfb0bbe3ac4947958cd9ba93. Are they incorrect?
9. Public exists
10. Outside scope of PR
11. 0.7 is in dynamically generated sitemaps
12. I have removed login (unsure, don't we want to index the login page?)
13. That is not part of this PR

joshestein:
@claude

5. We have `sitemap.xml` which contains the URLs and `sitemap_index.xml` which contains the sitemaps. I don't think this is self-referential. Nonetheless, I've made the change to appease you (again).
7. I didn't change anything about the API endpoint, only extracted the DB part for re-use.
11. Look at the dynamically generated `sitemap.xml.ts` files. They all use 0.7

apps/website/scripts/sitemap-generator.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/** Parse git history to determine when file was last updated */
+const getLastModified = (filePath: string): string => {
+  try {
+    const result = execFileSync('git', ['log', '-1', '--format=%cI', '--', filePath], { encoding: 'utf8' });
+    return result.trim() || new Date().toISOString();
+  } catch {
+    return new Date().toISOString();
+  }
+};
```
marn-in-prod, line 28: Wouldn't the last modified date be different than the git history date because content can be modified through airtable and thus without a commit?

marn-in-prod: Is that better or worse than not including it?

apps/website/scripts/sitemap-generator.ts
marn-in-prod, line 1: Can you confirm that you've looked at the sitemap.xml that get's produced and it looks reasonable?

marn-in-prod (approved):
Generally looks good, I think there's a slightly higher chance than normal that this will break the build when you merge and so try to merge it when you're more available.

apps/website/scripts/sitemap-generator.ts
joshestein: This is not for content pulled through airtable (blogs, projects, courses), only the root pages (/courses, /blogs/, /projects, /about, etc.).

apps/website/scripts/sitemap-generator.ts
joshestein: Yep 👍 I've looked each time I make changes to ensure it's what we expect.

PR #1338, opened 2025-09-15
[Fix]: `isMobile` from device detect not reset between tests

### LINKED ISSUE #482: Mock deviceDetect in a way that can be cleared between tests
Currently some tests use `vi.spyOn(deviceDetect, 'isMobile', 'get').mockReturnValue(true);` to mock the value of `isMobile`. `vi.clearAllMocks();` doesn't reverse this which means it can interfere with other downstream tests. We should find a way to make tests that use this properly atomic

joshestein:
# Description

Two problems:

1. We were calling `afterAll`, not `afterEach`, so the call to `clearAllMocks` was needing to be directly invoked
2. Prefer `restoreAllMocks` to `clearAllMocks`. `isMobile` is not a function, which means it is treated as a property descriptor. From the [vitest docs](https://vitest.dev/api/vi.html#vi-restoreallmocks): `restoreAllMocks` will '...  restore original descriptors of spied-on objects.' This is not the case for `clearAllMocks`. I tested this by logging the value of `isMobile` within the component and validating that it is always reset to `false` unless you explicitly over-ride the mock.

## Issue

Fixes #482 

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/apps/website/src/components/about/HistorySection.test.tsx
+++ b/apps/website/src/components/about/HistorySection.test.tsx
@@ -20,6 +20,6 @@ describe('HistorySection', () => {
     const { container } = render(<HistorySection />);
     expect(container).toMatchSnapshot();
     expect(container.querySelector('.history-section__event-container--mobile')).not.toBeNull();
-    vi.clearAllMocks();
+    vi.restoreAllMocks();
   });
 });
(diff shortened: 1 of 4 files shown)


Will-Howard (approved):
LGTM, thanks for clearing this up!

PR #1339, opened 2025-09-15
Slack info messages

joshestein:
# Description

Currently `slackAlert` can only send messages to our 'prod' channel (which we use for important error alerts). As part of #1293 I implemented 'info' to alerts to our 'dev' channel using [Grafana alerts.
](https://grafana.com/docs/grafana/latest/alerting/) After discussing with Martin we agreed it would be better to introduce a new env variable and allow callers of `slackAlert` to have a way to control if they want to send an error or an info message.

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/libraries/ui/src/utils/makeMakeApiRoute.ts
+++ b/libraries/ui/src/utils/makeMakeApiRoute.ts
@@ -146,15 +146,10 @@ export const makeMakeApiRoute = <AuthResult extends BaseAuthResult>({ env, verif
         });
 
         logger.error('Internal error handling request:', err);
-
-        try {
-          await slackAlert(env, [
-            `Error: Failed request on route ${req.method} ${req.url}: ${err instanceof Error ? err.message : String(err)}`,
-            ...(err instanceof Error ? [`Stack:\n\`\`\`${err.stack}\`\`\``] : []),
-          ]);
-        } catch (slackError) {
-          logger.error('Failed to send Slack alert', slackError);
-        }
+        slackAlert(env, [
+          `Error: Failed request on route ${req.method} ${req.url}: ${err instanceof Error ? err.message : String(err)}`,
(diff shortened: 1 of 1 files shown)


libraries/utils/src/slackNotifications.ts
```diff
@@ -2,32 +2,36 @@ type SlackAlertEnv = {
...
+      await sendSingleSlackMessage(env, messages[i]!, level, res.ts);
     }
   } catch (error) {
     // eslint-disable-next-line no-console
     console.error('Error sending Slack alert:', error);
   }
 };
 
-const sendSingleSlackMessage = async (env: SlackAlertEnv, message: string, threadTs?: string): Promise<{ ts: string }> => {
+const sendSingleSlackMessage = async (env: SlackAlertEnv, message: string, level?: 'error' | 'info', threadTs?: string): Promise<{ ts: string }> => {
```
marn-in-prod, line 23: Nit: Can you make level have a default value equal to error? That way it's clear to the developer what the "sane default" is. 

` level: 'error' | 'info' = 'error',`

Then when figuring out channel you can directly just do `level === 'error'`

marn-in-prod (approved):
I don't think we needed to add the INFO_SLACK_CHANNEL_ID to every template and package in the monorepo. I was thinking that we would only include it when it was actually being used in that specific package. I think just adding it everywhere but not actually using it is a bit confusing.

marn-in-prod:
> @marn-in-prod since makeApiRoute calls slackAlert we need INFO_SLACK_CHANNEL_ID to be present in the env passed to any makeApiRoute() call. Since all apps use makeApiRoute all of them need the new env variable, otherwise TS is unhappy.

That makes sense. You should just go ahead and submit this PR then!

PR #1341, opened 2025-09-15
[Feat] Updated landing page structure and copy

dewierwan:
# Description

- Lots of AGI Strategy landing page improvements. 
- Updated the quotes section to be clickable, so users can see the original source. 
- Updated the "community members" section with new people. 
- Updated the copy for the value prop section and moved it to the top.
- The mobile version looks ~identical as before, just with new copy.

## Screenshot

| 📸 | **Before and After** |
|---------|---|
| 🖥️ | CleanShot 2025-09-15 at 17 22 44@2x |
| 🖥️ | CleanShot 2025-09-15 at 17 22 54@2x |


dewierwan:
@greptile-apps please do a final review

PR #1343, opened 2025-09-15
Missing login tests

### LINKED ISSUE #720: Add tests to Login.tsx
```
...
const CUSTOM_REDIRECT_PATH = '/custom-path';

describe('LoginRedirectPage', () => {
  test('should maintain the redirect_to value in the signinRedirect call', async () => {
    ...
    expect(mockCreateSigninRequest).toHaveBeenCalledWith({
        state: { redirectTo: CUSTOM_REDIRECT_PATH },
      });
  });

  test('should navigate to the redirect_to value when auth is present', () => {
    (useAuthStore as any).mockReturnValue({ auth: { token: 'test-token' } });
    const mockGet = vi.fn().mockImplementation((param) => {
      if (param === 'redirect_to') return CUSTOM_REDIRECT_PATH;
      return null;
    });
    (useSearchParams as any).mockReturnValue({ get: mockGet });

    render(<LoginRedirectPage oidcSettings={mockOidcSettings} />);

    expect(Navigate).toHaveBeenCalledWith({
      url: CUSTOM_REDIRECT_PATH
    }, expect.anything());

    // Verify createSigninRequest is never called when auth is present
    expect(mockCreateSigninRequest).not.toHaveBeenCalled();
  });

  test('should use default redirect path when redirect_to is not provided', async () => {
    ...
    expect(mockCreateSigninRequest).toHaveBeenCalledWith({
        state: { redirectTo: '/' },
      });

  });
});

describe('LoginOauthCallbackPage', () => {
  test('should handle successful login callback and redirect to custom path', async () => {
      ...
      expect(mockRouter.push).toHaveBeenCalledWith(CUSTOM_REDIRECT_PATH);
  });
});
```

joshestein:
# Description

Fixes #720.

## Developer checklist

## Screenshot

NA


libraries/ui/src/Login.test.tsx
marn-in-prod, line 1: window.location is a global object across all tests right? So if one of these tests are modifying it, then they wouldn't be reset across the tests currently? Shouldn't it be necessary to reset the window.location in an afterEach for these tests?

marn-in-prod (approved):
Generally looks good. I think some of these might be harder to test any more than what you've already done. 

Just note the one comment about window.location which I think is a quick change.

libraries/ui/src/Login.test.tsx
joshestein: This was not as simple as doing `window.location = originalWindowLocation` because of type errors. And trying to use `Object.defineProperty` with a spread operator was leading to `window.location.href` being undefined instead of being reset to original value.

It's a little complicated but I think it's working correctly using this commit: https://github.com/bluedotimpact/bluedot/pull/1343/commits/2d6500c2358c85d599545da086e28bbf1aa3425e

How does it look to you?

joshestein:
@claude review

PR #1387, opened 2025-09-25
[Feature] Show GroupDiscussionBanner for facilitators

Will-Howard:
# Description
- Make GroupDiscussionBanner display for facilitators
- Show the host key in the button for facilitators, copy it to the clipboard on click

## Issue

No issue, requested by @anglilian in Slack

## Developer checklist

- [X] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [X] Considered having snapshot tests and/or happy path functional tests (will add tests for this component overall in #1389 )
- [X] Considered adding Storybook stories (also covered in #1389 )

## Screenshot

| 📸 |  |
|---------|---|
| 📱  | Screenshot 2025-09-25 at 11 42 50 |
| 🖥️ | Screenshot 2025-09-25 at 11 42 44 |

### DIFF
--- a/apps/website/src/components/courses/GroupDiscussionBanner.tsx
+++ b/apps/website/src/components/courses/GroupDiscussionBanner.tsx
@@ -23,7 +23,7 @@ type GroupDiscussionBannerProps = {
   unit: Unit;
   groupDiscussion: GroupDiscussion;
   userRole?: 'participant' | 'facilitator';
-  hostKey?: string;
+  hostKeyForFacilitators?: string;
   onClickPrepare: () => void;
 };
 
@@ -31,7 +31,7 @@ const GroupDiscussionBanner: React.FC<GroupDiscussionBannerProps> = ({
   unit,
   groupDiscussion,
   userRole,
-  hostKey,
+  hostKeyForFacilitators,
   onClickPrepare,
 }) => {
   const [groupSwitchModalOpen, setGroupSwitchModalOpen] = useState(false);
@@ -89,9 +89,9 @@ const GroupDiscussionBanner: React.FC<GroupDiscussionBannerProps> = ({
     : '';
 
   const copyHostKeyIfFacilitator = async () => {
-    if (userRole === 'facilitator' && hostKey) {
+    if (userRole === 'facilitator' && hostKeyForFacilitators) {
       try {
-        await navigator.clipboard.writeText(hostKey);
+        await navigator.clipboard.writeText(hostKeyForFacilitators);
       } catch (error) {
         // eslint-disable-next-line no-console
         console.warn('Failed to copy host key to clipboard:', error);
@@ -99,7 +99,7 @@ const GroupDiscussionBanner: React.FC<GroupDiscussionBannerProps> = ({
     }
   };
 
-  const joinButtonText = userRole === 'facilitator' && hostKey ? `Join discussion (Host key: ${hostKey})` : 'Join discussion';
+  const joinButtonText = userRole === 'facilitator' && hostKeyForFacilitators ? `Join discussion (Host key: ${hostKeyForFacilitators})` : 'Join discussion';
 
   return (
     <div className="flex flex-col p-2 border-1 border-charcoal-light gap-2">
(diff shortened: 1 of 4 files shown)


apps/website/src/components/courses/GroupDiscussionBanner.tsx
```diff
@@ -84,6 +88,19 @@ const GroupDiscussionBanner: React.FC<GroupDiscussionBannerProps> = ({
...
+        await navigator.clipboard.writeText(hostKey);
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.warn('Failed to copy host key to clipboard:', error);
+      }
+    }
+  };
+
+  const joinButtonText = userRole === 'facilitator' && hostKey ? `Join discussion (Host key: ${hostKey})` : 'Join discussion';
+
```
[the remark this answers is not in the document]
Will-Howard: This is fine, it was explicitly asked for, and I don't believe the host key is a very high security secret

somepersonnamedjosh:
@claude review pls!

apps/website/src/components/courses/GroupDiscussionBanner.tsx
```diff
@@ -84,6 +88,19 @@ const GroupDiscussionBanner: React.FC<GroupDiscussionBannerProps> = ({
     ? `https://app.slack.com/client/T01K0M15NEQ/${groupDiscussion.slackChannelId}`
     : '';
 
+  const copyHostKeyIfFacilitator = async () => {
+    if (userRole === 'facilitator' && hostKey) {
+      try {
+        await navigator.clipboard.writeText(hostKey);
+      } catch (error) {
+        // eslint-disable-next-line no-console
```
marn-in-prod, line 96: Is there a way to surface this info to the end user easily? Specifically, the error in a error boundary

apps/website/src/components/courses/UnitLayout.tsx
```diff
@@ -31,7 +32,6 @@ type Unit = InferSelectModel<typeof unitTable.pg>;
 type Chunk = InferSelectModel<typeof chunkTable.pg>;
 type UnitResource = InferSelectModel<typeof unitResourceTable.pg>;
 type ExerciseType = InferSelectModel<typeof exerciseTable.pg>;
```
marn-in-prod, line 34: ARen't these types already exported in `schema.ts`? Can you remove them and just import the types directly

apps/website/src/pages/api/courses/[courseSlug]/[unitNumber]/groupDiscussion.ts
Comment (marn-in-prod) [inline apps/website/src/pages/api/courses/[courseSlug]/[unitNumber]/groupDiscussion.ts:1]: I know the current file has a common pattern done in the past, but can we instead stop making all of the response types optional and instead just throw errors when something isn't found. Moving forward I think all of our backend api routes will be better of doing something like:

So things like
```
if (!course) {
  throw new createHttpError.NotFound('Course not found');
}
```

```
if (!isRegistered) {
  throw new createHttpError.Forbidden('Not registered for course');
}
```
That way the ```GroupDiscussionWithZoomInfo``` would only have an optional hostKey since that's only there for facilitators. Fraknly maybe even rename hostKey to something like ```hostKeyForFaciliators``` because right now it's not clear why it's optional or when it should be

apps/website/src/components/courses/UnitLayout.tsx
Will-Howard: Yep, done

apps/website/src/components/courses/GroupDiscussionBanner.tsx
Will-Howard: IMO this isn't worth doing, because:
- It will be very rare that copying to clipboard fails. The other place we do this we don't handle errors at all
- The key is just a 6 digit number, so it's easy to copy manually if needed. It is also shown again on the next screen once the facilitator clicks this button

PR #1417, opened 2025-10-04
[Bug] Add tests for SEO related content being populated during SSR

### LINKED ISSUE #1365: [Feature] Render SEO related content server side
### 🧩 Summary

There are several places where we set SEO-relevant head tags, such as meta descriptions, based on client side API calls. This means that bots that don't run the javascript won't get the correct head tags. We should make sure these are all populated server side by `getStaticProps` or `getServerSideProps`.

Example from Martin:
>To illustrate what I mean, all of the blog pages should have a <meta> tags but because they're only populated with correct information after an API call, non-sophisticated bots won't actually see detailed SEO data. It's also [why we don't see correct blog titles in notion preview](https://github.com/bluedotimpact/bluedot/issues/1099)

### ✅ Definition of done

- All pages render with proper <meta> tags (title, description, Open Graph tags, etc.) without requiring JavaScript
- OpenGraph preview (https://www.opengraph.xyz/) show correct metadata for all public pages

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P3" => "Low"}

Will-Howard:
# Description
Part of an overarching [ticket](https://github.com/bluedotimpact/bluedot/issues/1365) to render all content that may be needed for SEO or link previews server side. These two previous PRs refactored all pages where this wasn't already the case:
- #1408 
- #1416 

This PR adds tests to assert the SEO relevant content is in the DOM after the first render. The `<Head>` component doesn't work in tests by default, so I've added a `renderWithHead` mocking utility, which was inspired by this github discussion comment: https://github.com/vercel/next.js/discussions/11060#discussioncomment-33628

## Issue

Fixes #1365 

## Developer checklist

## Screenshot

N/A

### DIFF
--- a/apps/website/src/__tests__/pages/blog/[slug].test.tsx
+++ b/apps/website/src/__tests__/pages/blog/[slug].test.tsx
@@ -1,0 +1,77 @@
+import {
+  describe, expect, test, beforeEach, vi,
+} from 'vitest';
+import { Blog } from '@bluedot/db';
+import BlogPostPage from '../../../pages/blog/[slug]';
+import { renderWithHead } from '../../testUtils';
+
+// Mock <Head>, which doesn't work in tests. See docstring of
+// `renderWithHead` for more details.
+vi.mock('next/head', () => ({
+  __esModule: true,
+  default: ({ children }: { children: React.ReactNode }) => {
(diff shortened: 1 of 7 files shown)


marn-in-prod (approved):
LGTM

Issue #1405, opened 2025-10-01
[Bug] Next discussion banner doesn't appear in Unit 1


anglilian:
### 🧩 Summary

My next discussion banner shows in all other units but Unit 1.

<img width="2306" height="1714" alt="Image" src="https://github.com/user-attachments/assets/0dcf5db2-c373-4ba3-a02e-9ad3b4bff343" />

<img width="2310" height="1768" alt="Image" src="https://github.com/user-attachments/assets/a7f4ab63-e8c5-4bb5-bfed-78af341ba572" />

### 🛝 Steps to reproduce

If you're able to access my account or any other facilitator's account.

### ✅ Definition of done

Facilitators should be able to see the next discussion banner in unit 1

### Suggested Priority Level

{"P2" => "Medium (Default)"}

joshestein:
@anglilian are you still seeing this? Are there any errors in your console? I'm seeing the banner in unit 1.

<img width="1423" height="852" alt="Image" src="https://github.com/user-attachments/assets/80138169-eafd-4881-a6cd-495f0802ce37" />

anglilian:
I saw it an hour after I posted this issue. I assume it was because someone fixed it? I'll mark this as done!

PR #1421, opened 2025-10-04
`CourseDetails` cleanup

joshestein:
# Description

Was taking a look at the `CourseDetails` component as part of #1405. Made some changes:

1. Re-use existing DB types
2. Prefer axios `loading` instead of dedicated loading state
3. Re-use existing time formatting utils
4. Prefer `last:` tailwind CSS class to prop drilling `isLast`. We can also make this change for parent components, but I didn't want this PR to get too large
5. Mock HTTP requests for stories (so that they actually work, which they didn't before?!)

## Developer checklist

## Screenshot

As a result of re-using some of the time formatting utils the abbreviations for time units have been expanded:

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/apps/website/src/components/settings/CourseDetails.tsx
+++ b/apps/website/src/components/settings/CourseDetails.tsx
@@ -6,6 +6,8 @@ import { GroupDiscussion, GetGroupDiscussionResponse } from '../../pages/api/gro
 import GroupSwitchModal from '../courses/GroupSwitchModal';
 import { formatDateMonthAndDay, formatTime12HourClock } from '../../lib/utils';
 
+const HOUR_IN_SECONDS = 60 * 60; // 1 hour in seconds
+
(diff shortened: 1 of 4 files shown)


joshestein:
@claude review

apps/website/src/components/settings/CourseDetails.tsx
```diff
@@ -221,7 +168,7 @@ const CourseDetails = ({
                     size="small"
                     url={buttonUrl}
                     disabled={!discussion.zoomLink && isStartingSoon}
-                    target={openInNewTab ? '_blank' : undefined}
+                    target={isStartingSoon ? '_blank' : undefined}
```
joshestein, line 171: Tbh I feel like we should always open in new tab...

marn-in-prod (approved):
LGTM

PR #1426, opened 2025-10-07
tRPC foundations

### LINKED ISSUE #1260: [Feature] Migrate us from Axios to React Query/Trpc
### 🧩 Summary

Implement React Query for server state management, something that we've been needing to do for a while and has been talked about several times already. Finally putting it down as a pre-req for #1223 since it'll make the implementation of that feature simpler, plus any other features that require multiple components to share server-side state.

This will let us use the backend as the single source of truth instead of needing to sync server state through zustand. 

### ✅ Definition of done

We no longer use Axios for http calls but rather just trpc or react query directly.

### 📌 Notes / Designs

None

### Suggested Priority Level

{"P2" => "Medium (Default)"}

joshestein:
# Description

This PR adds tRPC (based on the [guide for Next.js](https://trpc.io/docs/client/nextjs/setup)) and migrates two existing API routes:

1. `/pages/api/cms/jobs/index.ts`, which was responsible for getting all published jobs
2. `/pages/api/users/change-password/index.ts`, which was responsible for sending a PATCH request to update user password

The first route does not require auth and makes use of `publicProcedure`. This should be the model for all requests that DON'T need auth. [The router](https://github.com/bluedotimpact/bluedot/blob/josh/1260-trpc/apps/website/src/server/routers/jobs.ts) also demonstrates a tRPC query, used to fetch data.

The second route does require auth and makes use of `protectedProcedure`. This should be the model for requests that DO need auth. [The router](https://github.com/bluedotimpact/bluedot/blob/josh/1260-trpc/apps/website/src/server/routers/users.ts) also demonstrates a tRPC mutation, used to mutate data.

There is a lot of noise in the PR because I also updated tests to lay the foundation for how we should be testing tRPC components going forward. Apologies for this - I do think it's important.

A summary of the most important changes:

1. Adds all tRPC boilerplate, including a context middleware that checks auth
2. Creates two routers - `routers/jobs.ts` that adds a procedure for querying published jobs and `routers/uers.ts` that adds a procedure for mutating user password
3. Updates `JobListSection` and `join-us/index.ts` to fetch jobs using tRPC and re-use tRPC types
4. Updates `PasswordSection.tsx` to use tRPC to update password
5. Uses the tRPC `onError` handler to send slack alerts when we get some error
6. Create infrastructure for mocking tRPC in tests and update `PasswordSection.test.tsx` as a guide for how to mock tRPC calls

## Issue

Fixes #1260.

## Developer checklist

## Screenshot

NA


apps/website/src/server/routers/jobs.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const jobsRouter = router({
+  getJobs: publicProcedure
+    .query(async () => {
+      const allJobs = await db.scan(jobPostingTable, {
+        publicationStatus: 'Published',
+      });
+
+      // Sort jobs alphabetically by title and remove the body field
+      // TODO: just fetch these fields from the DB in the first place
```
joshestein, line 13: I'm not sure we actually have a way to only select particular fields using the airtable wrapper? Is the recommended way to directly query the wrapped db?

apps/website/src/server/context.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  try {
+    const auth = await loginPresets.keycloak.verifyAndDecodeToken(token);
+    return { auth };
+  } catch (error) {
+    // TODO: do we need this fallback check to googleBlueDot?
+    // try {
+    //   auth = await loginPresets.googleBlueDot.verifyAndDecodeToken(token);
+    // } catch {
+    // }
```
joshestein, line 22: I'm not sure how to handle this, or what the strategy regarding two login presets should be? Should we just try to authenticate both, and only if both fail do we set auth to null?

marn-in-prod:
Framing: I haven't gone through everything in the PR. I saw your comment about auth, and as I mentioned initially, I think a single simple route would have been sufficient given the complexity already here (there's already significant setup and feature parity work). Remember, there's very little downside to a small PR, but quite a bit of downside to a larger one.

Ideally, we'd want to achieve feature parity with the existing backend functionality (makeMakeApiRoute and related classes). We should also avoid requiring the frontend to maintain both older useAxios API calls alongside tRPC calls (FYI in case that was being considered).

I'd recommend splitting this into separate PRs. As suggested earlier, let's start with just the setup + feature parity work for a non-authenticated, low-priority route, and then build from there.

PR #1429, opened 2025-10-07
[Bug] Use 'warn-and-retry' option in airtable-ts to gracefully handle errors due to edits to Airtable bases

### LINKED ISSUE #1190: [Bug] Runtime type checking of Airtable results causes errors in production
### 🧩 Summary

From @marn-in-prod in slack (I will update the description if the problem looks different after investigating):

> I believe it's Airtable-ts that throws exceptions when our schema.ts doesn't match the Airtable schema (missing fields, different expected values, etc.). I really don't think it should be throwing exceptions at all - at least not for querying data. Since these are runtime errors, they typically don't show up until after deployment.

> I understand validating inserts, but for queries, the data could be coerced or we could at least keep parts of the site working instead of everything breaking. That would be hugely beneficial.
You can see examples in https://internal-rbg4723.slack.com/archives/C04SAGM4FN1 (the updated-tech-prod channel for me).

> Maybe instead of throwing an exception, our best case scenario would be to alert the channel that there's a schema mismatch but still have the backend return whatever data it could get instead of throwing 500 and falling over

### 🛝 Steps to reproduce

TBD pending looking at past examples

### ✅ Definition of done

Runtime type mismatches down throw errors, but do trigger slack alerts

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

Will-Howard:
# Description

_Pending merging a PR against the airtable-ts repo, I'm putting this up first so I can explain how we plan to use this_

We have seen errors in production due to:
- Unused fields being deleted from Airtable
- The types of fields being edited in Airtable

This PR adopts a change to [airtable-ts](https://github.com/domdomegg/airtable-ts), to support 'warn-and-retry' for these specific errors during queries (but not during write operations).

When one of these errors occurs, an `onWarning` callback will be called to send a slack alert with the content of the error. The query will then be retried, attempting to bypass the above validation steps. If this succeeds it will return the object with the affected fields set to null/undefined, if it fails it will throw another error which will also appear in slack.

## Issue

Fixes #1190

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📱  |  |
| 🖥️ |  |

### DIFF
--- a/apps/website/src/lib/api/db/index.ts
+++ b/apps/website/src/lib/api/db/index.ts
@@ -1,7 +1,12 @@
 import { PgAirtableDb } from '@bluedot/db';
+import { slackAlert } from '@bluedot/utils';
 import env from '../env';
 
 export default new PgAirtableDb({
   pgConnString: env.PG_URL,
   airtableApiKey: env.AIRTABLE_PERSONAL_ACCESS_TOKEN,
+  airtableTsOptions: {
+    validationMode: 'warn-and-retry',
+    onWarning: (warning: string) => slackAlert(env, [warning]),
+  },
 });
--- a/libraries/db/package.json
+++ b/libraries/db/package.json
@@ -12,7 +12,7 @@
     },
     "dependencies": {
         "@bluedot/utils": "*",
-        "airtable-ts": "^1.5.0",
+        "airtable-ts": "^1.7.0",
         "drizzle-orm": "^0.43.1",
         "happy-dom": "18.0.1",
         "pg": "^8.15.6"
--- a/libraries/db/src/lib/client.ts
+++ b/libraries/db/src/lib/client.ts
@@ -3,8 +3,9 @@ import {
 } from 'drizzle-orm';
 import { PgInsertValue, PgUpdateSetSource, PgColumn } from 'drizzle-orm/pg-core';
 import { drizzle } from 'drizzle-orm/node-postgres';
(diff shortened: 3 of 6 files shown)


Will-Howard:
@coderabbitai review

apps/pg-sync-service/src/lib/webhook.ts
```diff
@@ -157,6 +157,17 @@ export class AirtableWebhook {
         await this.recreateWebhookWithoutDeletedFields(deletedFields);
       }
     } else {
+      // Remove any fields that have been deleted in Airtable to avoid UNKNOWN_FIELD_NAME errors
+      // Note: duplicate functionality with `extractDeletedFieldsFromPayload`, see https://github.com/bluedotimpact/bluedot/issues/1464
```
marn-in-prod: Nit: I dont think we need an issue to refactor/clean up single digit lines of code. I just don't think it's worth doing. 

That's seperately from wanting to add tests though which given how we don't change this frequently, maybe makes sense to have as a p4.

marn-in-prod (approved):
I directionally like this PR. 

My biggest concern is that we NEED to be on top of the slack alerts so that we revert any accidentally Airtable column deletions within 7 days since the data will be gone after that point. 

I think in the past week we haven't gotten too many slack alerts (normally from mapping related errors) and so it's probably fine. 

When you merge, please confirm by looking at grafana that it is working correctly

PR #1439, opened 2025-10-08
tRPC Mock Service Worker for tests

joshestein:
# Description

We need a good way to mock tRPC requests. Although we could mock procedures using vitest this is complicated and tedious. Better to use a Mock Service Worker and use actual tRPC with control over the returned data.

## Issue

Part of #1260.

## Developer checklist

## Screenshot

NA


joshestein:
@claude ignore all other bots. Review

Issue #1427, opened 2025-10-07
Migrate remaining Axios routes to tRPC


joshestein:
### 🎨 Summary

In https://github.com/bluedotimpact/bluedot/pull/1426/ we begun the process of migrating to tRPC.

We need to migrate our remaining routes.

### User Management
- [ ] `GET/PATCH /api/users/me` - User profile retrieval and updates

### Course Management
- [ ] `GET /api/courses` - List all active courses
- [ ] `GET /api/courses/[courseSlug]` - Get course details by slug
- [ ] `GET /api/courses/[courseSlug]/[unitNumber]` - Get unit details
- [ ] `POST /api/courses/[courseSlug]/[unitNumber]/feedback` - Submit unit feedback
- [ ] `GET /api/courses/[courseSlug]/[unitNumber]/resources` - Get unit resources
- [ ] `GET/POST /api/courses/[courseSlug]/[unitNumber]/groupDiscussion` - Group discussion operations
- [ ] `GET /api/courses/[courseSlug]/units/[unitId]` - Get unit by ID
- [ ] `GET /api/courses/[courseSlug]/group-switching/available` - Get available groups for switching
- [ ] `POST /api/courses/[courseSlug]/group-switching` - Switch course groups

### Course Registrations
- [ ] `GET /api/course-registrations` - List user's course registrations
- [ ] `GET/POST /api/course-registrations/[courseId]` - Manage course registration by ID

### Exercises
- [ ] `GET /api/courses/exercises/[exerciseId]` - Get exercise details
- [ ] `GET/PUT /api/courses/exercises/[exerciseId]/response` - Manage exercise responses

### Resource Completion
- [ ] `GET/POST /api/courses/resource-completion/[unitResourceId]` - Track resource completion

### Certificates
- [ ] `POST /api/certificates/request` - Request course certificate
- [ ] `GET /api/certificates/[certificateId]` - Get certificate by ID

### CMS - Blogs
- [ ] `GET /api/cms/blogs` - List all published blogs
- [ ] `GET /api/cms/blogs/[slug]` - Get blog post by slug

### CMS - Projects
- [ ] `GET /api/cms/projects` - List all published projects
- [ ] `GET /api/cms/projects/[slug]` - Get project by slug

### Group Discussions
- [ ] `GET/PATCH/DELETE /api/group-discussions/[id]` - Manage group discussion by ID

### Admin
- [ ] `POST /api/admin/request-sync` - Request data sync
[...]

PR #1455, opened 2025-10-13
tRPC migrate `/api/course-registrations` routes

joshestein:
# Description

Migrate `/api/course-registrations` to tRPC.

I tried for some time to fix tests in `CertificateLinkCard.test.tsx` using tRPC MSW. I kept running into problems where the component would be stuck in a loading state. I've given up, and mocked tRPC directly.

We currently have 3 API route handlers:

1. `/api/course-registrations/index.ts` to handle getting all course registrations (replaced with `courseRegistrations.getAll`)
2. GET `/api/course-registrations/[courseId]/index.ts` to handle getting a single course registration (used in `CertificateLinkCard.tsx` (replaced with `courseRegistrations.getById`)
3. POST `/api/course-registrations/[courseId]/index.ts` to handle updating a single course registration (used in `[[...chunkNumber]].tsx` (replaced with `courseRegistrations.getOrCreate`)

## Issue

Part of #1260 and #1427.

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/apps/website/src/pages/courses/[courseSlug]/[unitNumber]/[[...chunkNumber]].tsx
+++ b/apps/website/src/pages/courses/[courseSlug]/[unitNumber]/[[...chunkNumber]].tsx
@@ -75,7 +75,7 @@ const CourseUnitChunkPage = ({
     if (shouldRecordCourseRegistration) {
       createCourseRegistrationMutation({ courseId: unit.courseId, source: latestUtmParams.utm_source });
     }
-  }, [auth, unit.courseId, latestUtmParams.utm_source, createCourseRegistrationMutation]);
+  }, [auth, unit.courseId, latestUtmParams.utm_source]);
 
   useEffect(() => {
     if (chunks && (chunkIndex < 0 || chunkIndex >= chunks.length)) {


apps/website/src/server/routers/course-registrations.ts
```diff
@@ -0,0 +1,47 @@
+import { applicationsCourseTable, courseRegistrationTable } from '@bluedot/db';
+import z from 'zod';
+import db from '../../lib/api/db';
+import { protectedProcedure, router } from '../trpc';
+
+export const courseRegistrationsRouter = router({
+  getById: protectedProcedure
```
joshestein: @Will-Howard In https://github.com/bluedotimpact/bluedot/pull/1476/files#diff-7b85b6c5a16b247e3fdc25e7d011eb9e4be0a5179d098934a793c15106f64dabL22 you made the API route handle both GET and POST requests with the same behaviour.

Can you confirm that the GET request is supposed to early return after `db.getFirst(courseRegistrationTable)`? That is, it's not supposed to insert into the course registration table (which would be weird for a GET request but that's what it was doing before so...).

Will-Howard: No I was intending to fully migrate the route to POST, and just leave the GET case there for backwards compatibility (for clients that haven't received the new code yet).

I see I forgot to migrate the case in `CertificateLinkCard`, so it is still using GET. I would use `getOrCreate` for this one too and then remove the `getById` route. Up to you how to handle this though, it shouldn't make any difference practically

joshestein: Okay. Is that the same for https://github.com/bluedotimpact/bluedot/pull/1442/files#diff-a0447e699130c5572389e32a48564d1478b33007a64f639d736a9a29ffb353aaR43?

That we are intending to only have a POST request, and we deprecate the GET route?

Will-Howard: > Okay. Is that the same for https://github.com/bluedotimpact/bluedot/pull/1442/files#diff-a0447e699130c5572389e32a48564d1478b33007a64f639d736a9a29ffb353aaR43?

Hm I didn't think about it in that but I think splitting out the query/mutation cases would make sense there too, thanks for catching that!

>I've created 3 procedures (or API routes):
>
>1. `getById` - called by `CertificateLinkCard`
>2. `getAll` - called by `CoursesContent`
>3. `create` - called by `[[...chunkNumber]]`
>
>How does that seem to you?

That seems good semantically 👍. Minor nits on naming:
- I think `getByCourseId` would be better to be clear it's not the course registration id
- Suggest `ensureExists` rather than `create`, because it's expected that the registration will already exist most of the time this is called

apps/website/src/components/courses/CertificateLinkCard.test.tsx
```diff
@@ -1,19 +1,32 @@
...
-import { useRouter } from 'next/router';
-import CertificateLinkCard from './CertificateLinkCard';
-import { TrpcProvider } from '../../__tests__/trpcProvider';
+import { createMockCourseRegistration } from '../../__tests__/testUtils';
 import { FOAI_COURSE_ID } from '../../lib/constants';
+import CertificateLinkCard from './CertificateLinkCard';
+
+// I tried pretty hard to use tRPC MSW but kept running into issues where the component would be permanently stuck in
+// loading state for authenticated tests. I spent a couple hours trying to fight it, have decided to give up and mock
+// tRPC directly.
```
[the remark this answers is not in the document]
joshestein: Good call 👍 https://github.com/bluedotimpact/bluedot/issues/1528

marn-in-prod (approved):
Looks good. See minor comments

Issue #1457, opened 2025-10-14
[Feature] Admin certificate creation for facilitated courses


somepersonnamedjosh:
### 🧩 Summary

Currently, certificates for facilitated courses can only be created through a mostly deprecated Airtable-only flows. We need a system where course operators can efficiently create certificates through Airtable that:

- Uses the website/backend as the source of truth (not Airtable)
- Shows certificates at the unified bluedot.org/certification page
- Syncs data back to Airtable automatically via pg-sync

Proposed Solution:
Create an API endpoint that Airtable automations can call to create certificates for students in facilitated courses.

### ✅ Definition of done

Requirements:

API endpoint at /api/admin/certificates/create that (very similar to the FoAI course):

- Takes courseRegistrationId as input
- Requires authentication (decide: admin secret, API key, or other)
- Sets certificateId = courseRegistrationId on the course registration
- Sets certificateCreatedAt timestamp
- Returns the certificate details

Airtable integration:

- Checkbox or button in Course Registration table to trigger certificate creation
- Automation that calls the API endpoint
- Fields sync back to Airtable via pg-sync

Notes:
I tried this in PR #1456 attempted this but closed it (managed to exposed my secret in test doc (roatated and no longer used) + got stuck in testing). I think the infrastructure configuration approach was sound

Success Criteria:

- [ ]  Course operator can click button/checkbox in Airtable
- [ ]  Certificate is created in website database
- [ ]  Certificate appears at bluedot.org/certification?id={certificateId}
- [ ]  Data syncs back to Airtable within 10 minutes

### Suggested Priority Level

{"P1" => "High"}

PR #1463, opened 2025-10-14
tRPC migrate `/api/admin` routes

joshestein:
# Description

We have two admin routes that are used by our sync dashboard. We should migrate these as part of our general tRPC migration, but it's also useful for #1457 which needs a dedicated admin route.

1. Introduce a new `adminProcedure` for routes that only administrators should be able to use
2. Migrate `/api/admin/request-sync` to tRPC
3. Migrate `/api/admin/sync-history` to tRPC
4. Updates the frontend component to use tRPC to make requests
5. Updates frontend to derive state from `useQuery` data, errors

## Issue

Part of #1260 and #1427.

## Developer checklist

- [x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [x] Considered having snapshot tests and/or happy path functional tests
- [x] Considered adding Storybook stories

## Screenshot

NA

PR #1470, opened 2025-10-17
Public admin certificate creation procedure

joshestein:
# Description

Adds a new public procedure to the certificates router that allows for creation of a certificate.

Adds some basic authentication by (1) checking the provided email (intended to be the user who calls the API) is an admin and (2) there is a matching secret token. The use case is calling the new route from an airtable script:

```ts
const inputConfig = input.config()
let courseRegistrationId = inputConfig.record;
let studentName = inputConfig.fullName || 'Student';

const response = await fetch('https://bluedot.org/api/trpc/certificates.create', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
  },
  body: JSON.stringify({
    courseRegistrationId,
    token: input.secret('CERTIFICATE_CREATION_TOKEN'),
    email: inputConfig.runnerEmail,
  }),
});

if (response.ok) {
  const { result: { data }} = await response.json();
  console.log('✅ Certificate created successfully!');
  console.log(`Certificate ID: ${data.certificateId}`);
  console.log(`Created at: ${new Date(data.certificateCreatedAt * 1000).toISOString()}`);
  console.log(`View at: https://bluedot.org/certification?id=${data.certificateId}`);
} else {
  const error = await response.text();
  console.error('❌ Failed to create certificate');
  console.error(`Status: ${response.status}`);
  console.error(`Error: ${error}`);
  throw new Error(`Failed to create certificate: ${error}`);
}
```

Airtable has support for shared secrets: https://community.airtable.com/announcements-6/introducing-secrets-in-automations-secure-your-api-keys-and-sensitive-data-45092. We have created a shared secret called `CERTIFICATE_CREATION_TOKEN`, which is supposed to be shared for all BlueDot users. It's not working for me. If it's also not working for you, create a personal Airtable secret - you can find the value in 1pass.

Note that `courseRegistrationId` depends on which base you are running this from. If you are running from the 'Applications' base the column name is 'Airtable record ID'. If you are running from the 'Course runner' base the column name is 'Applications ID'. We have two scripts, setup for both bases. I've modified them accordingly.

TODO:

- [ ] Add field ID to schema - not sure this is actually necessary for this PR
- [ ] Wait for .env documentation to be complete, confirm that we don't need to add to `.env.*.template` since this is never intended for local usage - pretty sure we don't but would be nice to have cemented best practices for .env stuff

## Issue

Part of #1457.

## Developer checklist

## Screenshot

NA

### DIFF
--- a/apps/website/src/server/routers/certificates.ts
+++ b/apps/website/src/server/routers/certificates.ts
@@ -13,8 +13,8 @@ export const certificatesRouter = router({
   create: publicProcedure
     .input(z.object({ adminEmail: z.string().email(), courseRegistrationId: z.string(), publicToken: z.string() }))
     .mutation(async ({ input: { adminEmail, courseRegistrationId, publicToken } }) => {
-      // This is similar to the `request` procedure below, but it does not rely on authentication and allows a
-      // certificate to be created even if not all exercises are complete.
+      // This is similar to the `request` procedure below, but it relies on a shared secret for authentication and
+      // allows a certificate to be created even if not all exercises are complete.
       if (!env.CERTIFICATE_CREATION_TOKEN) {
         throw new TRPCError({ code: 'INTERNAL_SERVER_ERROR', message: 'Certificate creation not configured' });
       }


apps/website/src/server/trpc.ts
```diff
@@ -70,6 +73,19 @@ const openTelemetryMiddleware = t.middleware(async (opts) => {
   }
 });
 
+export const checkAdminAccess = async (email: string) => {
```
joshestein: https://github.com/bluedotimpact/bluedot/pull/1463

joshestein:
@claude review

joshestein:
@claude https://github.com/bluedotimpact/bluedot/pull/1470/commits/c2630fc64eaba4f156af88dd08ef91a1c9457fbe

apps/website/src/server/routers/certificates.ts
```diff
@@ -1,10 +1,66 @@
...
+      if (tokenBuf.length !== secretBuf.length || !timingSafeEqual(tokenBuf, secretBuf)) {
+        throw new TRPCError({ code: 'UNAUTHORIZED', message: 'Invalid token' });
+      }
+
+      const hasAdminAccess = await checkAdminAccess(email);
+      if (!hasAdminAccess) {
+        throw new TRPCError({ code: 'FORBIDDEN', message: 'Forbidden' });
+      }
+
+      const courseRegistration = await db.get(courseRegistrationTable, {
```
marn-in-prod: Switch this to getFirst since I think users can have multiple registrations and this will throw if they have this bug

apps/website/src/server/routers/certificates.ts
```diff
@@ -1,10 +1,66 @@
...
-import { protectedProcedure, router } from '../trpc';
+import env from '../../lib/api/env';
+import {
+  checkAdminAccess, protectedProcedure, publicProcedure, router,
+} from '../trpc';
 
 export const certificatesRouter = router({
+  // This is a public procedure because it's called from an Airtable script, not from within the app
+  create: publicProcedure
+    .input(z.object({ email: z.string().email(), courseRegistrationId: z.string(), token: z.string() }))
```
marn-in-prod: Super nit: can we just rename the email and token to be something like, adminEmail and publicToken or something similar.

PR #1472, opened 2025-10-20
Don't expose sensitive tRPC internal server error or resource not found messages

joshestein:
# Description

We shouldn't expose sensitive information when we have internal server errors. To prevent this we replace the error message with a generic 'An internal server error occurred'.

Edit: updating to also exclude 404 errors with a generic 'Resource not found' based on [Martin's feedback.](https://github.com/bluedotimpact/bluedot/pull/1472?notification_referrer_id=NT_kwDOAaN71rQxOTc3NzU2NDAwNzoyNzQ5MTI4Ng#discussion_r2449581252)

## Issue

NA

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/apps/website/src/server/trpc.ts
+++ b/apps/website/src/server/trpc.ts
@@ -10,11 +10,8 @@ import { Context } from './context';
 // For instance, the use of a t variable is common in i18n libraries.
 const t = initTRPC.context<Context>().create({
   errorFormatter({ shape, error }) {
-    // Return full error details in development
-    if (process.env.NODE_ENV !== 'production') return shape;
-
     // Hide internal server error details in production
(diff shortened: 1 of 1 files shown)


apps/website/src/server/trpc.ts
```diff
@@ -8,7 +8,22 @@ import { Context } from './context';
...
+        message: 'An internal server error occurred',
+        data: {
+          ...shape.data,
+          stack: undefined,
+        },
+      };
+    }
+    return shape;
+  },
+});
```
marn-in-prod: I feel like we might want to add something similar to "NOT_FOUND" errors or something similar. 

The risk is low IMO, but since we're already making changes here we should add it

marn-in-prod (approved):
LGTM.

Issue #1422, opened 2025-10-05
[Code cleanup/deletion + devex] Consolidate backend practices while moving to tRPC


marn-in-prod:
### 🧩 Summary

This should probably be added to our PR checklist eventually...

## Backend Code Standards (tRPC Migration)

This should be added to our PR checklist. Our backend code is currently inconsistent, and these patterns should be standardized once we move to tRPC.

Context: https://github.com/bluedotimpact/bluedot/issues/1260#issuecomment-3369137624

### Standards:

1. **Separate GET/PUT requests onto different endpoints** - No more method switching in a single handler

2. **Remove `type: 'success' as const` wrappers** - Return data directly since errors are thrown exceptions

3. **Replace `createHttpError` with `TRPCError`**
   ```typescript
   // Old
   throw new createHttpError.BadRequest('Expected PUT request to include body');
   
   // New
   throw new TRPCError({
     code: 'BAD_REQUEST',
     message: 'Expected PUT request to include body',
   });
   ```

4. **Always use Zod for input validation** - No manual type checking or string validation.
   ```typescript
   // Old
   const { courseId } = raw.req.query;
   if (typeof courseId !== 'string' || !courseId) {
     throw new createHttpError.BadRequest('Invalid courseId parameter');
   }
   
   // New
   .input(z.object({
     courseId: z.string().min(1),
   }))
   ```

5. **Return database responses immediately** when not doing additional logic
   ```typescript
   //  Old
   const job = await db.get(jobPostingTable, { slug });
   return {
     type: 'success' as const,
     job,
   };
   
   // New
   return await db.get(jobPostingTable, { slug });
   ```

6. **Remove explicit response body schemas** - tRPC infers types automatically. Use `inferProcedureOutput<AppRouter['jobs']['getJob']>` on the frontend only when explicitly typing variables. tRPC provides correct typing automatically in most cases.

---

**Note:** These standards only apply to new tRPC code. Don't refactor existing REST endpoints unless you're migrating them to tRPC.

### 🛝 Steps to reproduce

Not a bug, but you can see us using a lot of different patterns in our backend currently. 

### ✅ Definition of done

Once something is moved to tRPC it follows these patterns.

### Suggested Priority Level

{"P1" => "High"}

Issue #1366, opened 2025-09-19
[Feature] Est time + number of exercises per chunk in course curriculum component


alyssia-x:
### 🧩 Summary

Currently blocked by #1260. Branched off from #1364. Related to other blocked progress features such as #1164 and #1163 

<img width="585" height="159" alt="Image" src="https://github.com/user-attachments/assets/0c5e96ac-78da-4c43-853a-904f3c2c80f1" />

### ✅ Definition of done

Estimated time and number of exercises per chunk is displayed for each course unit title in the Course Curriculum Component in the AGI Strategy Lander

### 📌 Notes / Designs

[Figma](https://www.figma.com/design/62YlFNK7QS6z7SkrPjrwVb/Blue-Dot?node-id=1275-11437&p=f&t=J0qUys7aAVRow99A-0)

### Suggested Priority Level

{"P2" => "Medium (Default)"}

Issue #1164, opened 2025-08-05
[Feature] Track total course progress


herrhaase:
### 🧩 Summary

As a user I want to be able to track the progress that I made across a whole course. 

### ✅ Definition of done

1. Calculate the total progress a user has made on a course (in %).
2. Display the progress (label on Desktop and Mobile)
3. Display a visual progress bar (Mobile only)

### 📌 Notes / Designs

<img width="896" height="242" alt="Image" src="https://github.com/user-attachments/assets/76f39910-864a-4bf2-b856-9b649033cf02" />

<img width="1548" height="354" alt="Image" src="https://github.com/user-attachments/assets/d091ec94-f194-4674-b0d6-463dc99000af" />

### Suggested Priority Level

{"P2" => "Medium (Default)"}

marn-in-prod:
I'm going to say that this should be blocked until https://github.com/bluedotimpact/bluedot/issues/1260 since that will make this easier to implement

marn-in-prod:
@alyssia-x Could I take these progress related issues? I think I remember you mentioning in slack that you have less availability this month

Issue #1163, opened 2025-08-05
[Feature] Track progress of exercises / quizzes in course sidebar (menu on Mobile)


herrhaase:
### 🧩 Summary

As a user I want to be able to track the progress that I made on ecompleting exercises / quizzes ("1 of 3 completed").

### ✅ Definition of done

1. State of exercises / quizzes is tracked properly.
2. When a user re-opens an exercise / quiz the tracker resets as well.

### 📌 Notes / Designs

<img width="890" height="294" alt="Image" src="https://github.com/user-attachments/assets/cb98340d-d6b7-4e6a-9812-983a973810c4" />

### Suggested Priority Level

{"P2" => "Medium (Default)"}

alyssia-x:
@herrhaase This blue bar is in-scope as well, correct? 

<img width="1548" height="354" alt="Image" src="https://github.com/user-attachments/assets/3be01c2f-105b-4a90-9932-04d9eb9f2279" />

Or is this tracked in the new #1172? cc @YanniBDI

herrhaase:
@alyssia-x Lart of the new issue. 😊

PR #1477, opened 2025-10-21
Exercise API route migration for tRPC

alyssia-x:
# Description
Moving exercise API routes over to tRPC. Doing this to unblock progress related features. This is a 1:1 migration from the existing codebase. Changes to the original code and how it works is out of scope for this PR. 

## Issue

Progress related issues: https://github.com/bluedotimpact/bluedot/issues/1164, https://github.com/bluedotimpact/bluedot/issues/1163, https://github.com/bluedotimpact/bluedot/issues/1366

Following best practices here: https://github.com/bluedotimpact/bluedot/issues/1422.
Part of https://github.com/bluedotimpact/bluedot/issues/1260 and the checklist here https://github.com/bluedotimpact/bluedot/issues/1427

## Developer checklist
N/A

## Screenshot
N/A, No visual changes

### DIFF
--- a/apps/website/src/components/courses/exercises/Exercise.tsx
+++ b/apps/website/src/components/courses/exercises/Exercise.tsx
@@ -72,7 +72,7 @@ const Exercise: React.FC<ExerciseProps> = ({
           {...exerciseData}
           description={exerciseData.description || ''}
           title={exerciseData.title || ''}
-          exerciseResponse={responseData?.response || undefined}
+          exerciseResponse={responseData?.response}
           isLoggedIn={!!auth}
           onExerciseSubmit={handleExerciseSubmit}
         />


apps/website/src/server/routers/exercises.ts
```diff
@@ -0,0 +1,57 @@
...
+    .input(z.object({ exerciseId: z.string().min(1) }))
+    .query(async ({ input }) => {
+      return db.get(exerciseTable, { id: input.exerciseId });
+    }),
+
+  getExerciseResponse: protectedProcedure
+    .input(z.object({ exerciseId: z.string().min(1) }))
+    .query(async ({ input, ctx }) => {
+      const exerciseResponse = await db.getFirst(exerciseResponseTable, {
+        filter: { exerciseId: input.exerciseId, email: ctx.auth.email },
```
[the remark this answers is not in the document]
alyssia-x: We have tRPC middleware that we're using to catch this. You can see in our other trpc routers that we're doing the same thing.

apps/website/src/server/routers/exercises.ts
```diff
@@ -0,0 +1,57 @@
+import { z } from 'zod';
+import { exerciseTable, exerciseResponseTable } from '@bluedot/db';
+import { TRPCError } from '@trpc/server';
+import db from '../../lib/api/db';
+import { publicProcedure, protectedProcedure, router } from '../trpc';
+
+export const exercisesRouter = router({
+  getExercise: publicProcedure
+    .input(z.object({ exerciseId: z.string().min(1) }))
```
joshestein: Does this need to be an object or just a string? Also, what is the reason for the `min(1)` on the strings?

apps/website/src/components/courses/exercises/Exercise.tsx
```diff
@@ -21,58 +18,52 @@ const Exercise: React.FC<ExerciseProps> = ({
...
+  // Only fetch user response when authenticated
+  const {
+    data: responseData,
+    error: exerciseResponseError,
+    refetch: fetchExerciseResponse,
+  } = trpc.exercises.getExerciseResponse.useQuery(
+    { exerciseId },
+    {
+      enabled: !!auth,
+      retry: false,
```
joshestein: Why?

marn-in-prod: TLDR: I think this is technically correct given the intent is a 1:1 migration. Axios doesn't enable retries by default and I don't think we ever configured it to attempt retries.

marn-in-prod: I know it was out of scope for this PR, but I do think for future cleanup @alyssia-x, removing this after we can confirm that it's working in prod makes sense. In general, we'd want to rely on trpc global defaults so that we can easily configure all of them at the same time

apps/website/src/components/courses/exercises/Exercise.tsx
```diff
@@ -21,58 +18,52 @@ const Exercise: React.FC<ExerciseProps> = ({
...
-        completed: completed ?? true,
-      },
-      {
-        headers: {
-          Authorization: `Bearer ${auth?.token}`,
-        },
-      },
-    );
-  }, [exerciseId, auth]);
+    await saveResponseMutation.mutateAsync({
```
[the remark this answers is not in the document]
marn-in-prod: @joshestein Regarding the refetch.

When you've queried data, and then later mutating it. You need to let tRPC know that the UX that you're displaying should be updated with the mutated data. You can do this in three ways (off the top of my head), marking the keys in the query cache as invalidated, manually changing the query cache, or calling the refetch function. Without the refetch function, I don't think the multiple choice component (maybe there's others, just off the top of my head) won't work because the component won't re-render with the updated submitted MC.

The general best practice is to invalidate the query cache (if you don't want optimistic updates), or manually update the query cache (if you do want optimistic updates but then have to handle rolling back the UX) and then invalidating the query. 

So @alyssia-x, I do think you should do something like ``utils.exercises.getExerciseResponse.invalidate({ exerciseId });`` instead of specifically calling the refetch from the query key instead. You'd probably also want to  do this inside of tRPC's callbacks. You'd get utils from the `useUtils` hook in trpc `const utils = trpc.useUtils();`

Although I do understand that you probably did what you did because it's just following the previous pattern 1:1 with axios and that's what you mentioned in the PR that you were doing + I understand that lots of this code is going to change anyways.

apps/website/src/server/routers/exercises.ts
joshestein: Alrighty. Just want to have consistency across all our procedures. I will update the routes I migrated; @marn-in-prod please point these things out in your reviews.

apps/website/src/components/courses/exercises/Exercise.tsx
```diff
@@ -21,58 +18,52 @@ const Exercise: React.FC<ExerciseProps> = ({
...
-      Authorization: `Bearer ${auth?.token}`,
+  // Only fetch user response when authenticated
+  const {
+    data: responseData,
+    error: exerciseResponseError,
+    refetch: fetchExerciseResponse,
+  } = trpc.exercises.getExerciseResponse.useQuery(
+    { exerciseId },
+    {
+      enabled: !!auth,
```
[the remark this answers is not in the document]
joshestein: If you never want to run the query I think the canonical way is to use `skipToken`:

https://trpc.io/docs/client/react/disabling-queries

apps/website/src/components/courses/exercises/Exercise.tsx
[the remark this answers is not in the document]
joshestein: Okie, regardless of if you call `invalidate()` or `refetch()` I think it's best practice to put it in the `onSuccess` callback of `useMutation` and call `mutate` not `mutateAsync`.

https://tkdodo.eu/blog/mastering-mutations-in-react-query#mutate-or-mutateasync

apps/website/src/server/routers/exercises.ts
marn-in-prod: @joshestein @alyssia-x Totally my bad. I used your laptop earlier when prepping for the FLF presentation and didn't realize that I on the wrong computer lol.

@joshestein I'll rewrite this comment now.

joshestein:
@alyssia-x great, thank you! Can you please incorporate this diff before merging: https://github.com/bluedotimpact/bluedot/pull/1566/files#diff-747746ecf2c2f8858f1c82073d6d0ce153afb709195d94bff3a0852ae8338d39

joshestein:
@greptileai review

PR #1492, opened 2025-10-23
[Bug] Refactor access token refresh logic: Simplify, add regular retries

Will-Howard:
# Description

## Issue

Fixes #

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📱  |  |
| 🖥️ |  |

### DIFF
--- a/libraries/ui/src/utils/auth.test.tsx
+++ b/libraries/ui/src/utils/auth.test.tsx
@@ -4,7 +4,9 @@ import {
 import { OidcClient } from 'oidc-client-ts';
 import { render } from '@testing-library/react';
 import * as React from 'react';
-import { Auth, useAuthStore, withAuth } from './auth';
+import {
+  Auth, useAuthStore, withAuth,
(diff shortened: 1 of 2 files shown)


libraries/ui/src/utils/auth.test.tsx
```diff
@@ -89,36 +94,6 @@ describe('auth', () => {
...
-      const newAuth = createAuth({
-        expiresAt: Date.now() + 3600_000, // 1 hour
-        token: 'new-token',
-      });
-      useAuthStore.getState().setAuth(newAuth);
-
-      // Advance 10 seconds (past when initial token would have expired)
-      vi.advanceTimersByTime(10_000);
-
-      // Should still have the new token (old expiry timer should have been cancelled)
```
Will-Howard: The old expiry timer is no longer a concern with the new logic, as there is always just one setInterval running using whatever the latest token is

PR #1495, opened 2025-10-24
tRPC migrate  POST `/api/users/me`

joshestein:
# Description

We have a route called when we create users. Migrate this to tRPC.

The original API route returns the entire new user object when created. However, since we component only cares about `isNewUser`, I decided to only return this field.

## Issue

Part of  #1260 and #1427.

## Developer checklist

NA

## Screenshot

NA


joshestein:
@greptileai I've added https://github.com/bluedotimpact/bluedot/pull/1495/commits/0913497db73527cc14f79fb8903fd12ec34056dc, what do you think now?

PR #1498, opened 2025-10-24
tRPC migrate POST `/api/courses/[courseSlug]/group-switching/index.ts`

joshestein:
# Description

We have a few routes for group switching. This is an initial migration to handle requesting a group switch.

Changes from the original:

1. `courseSlug` is passed as a required input parameter, not a query parameter
2. We do not return 'success' when the mutation succeeds

## Issue

Part of #1260 and #1427.

## Developer checklist

NA

## Screenshot

NA


marn-in-prod (approved):
LGTM

PR #1502, opened 2025-10-24
[Bug] Only get-or-create course registrations for the Future of AI course

### LINKED ISSUE #1500: [Bug] Records being created by logged in users who click apply
### 🧩 Summary

Whenever a logged in user clicks the "browse curriculum" button it creates an application record for them.

This should not happen.

### 🛝 Steps to reproduce

1. Go to https://bluedot.org/courses/technical-ai-safety
2. Click "Browse curriculum"
3. Look at the [course registrations table](https://airtable.com/appnJbsG1eWbAdEvf/tblXKnWoXK3R63F6D/viwhz9oMemImIdLXZ?blocks=hide)

### ✅ Definition of done

When users click the browse curriculum button, it does not create an application for them.

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P1" => "High"}

Will-Howard:
# Description

For the Future of AI course, it logically makes sense for a course registration to be created when someone starts browsing the course, and this is needed (as written) to generate a certificate. For other courses, this isn't needed, and @anglilian has asked (#1500) to not auto-create course registrations as this clogs up the applications table.

## Issue

Fixes #1500

## Developer checklist

## Screenshot

N/A

### DIFF
--- a/apps/website/src/pages/api/course-registrations/[courseId]/index.tsx
+++ b/apps/website/src/pages/api/course-registrations/[courseId]/index.tsx
@@ -58,7 +58,14 @@ export default makeApiRoute({
 
       // Get-or-create if this is the future-of-ai course. Users are allowed to complete the course independently
       if (courseId === FOAI_COURSE_ID) {
-        const applicationsCourse = await db.get(applicationsCourseTable, { courseBuilderId: courseId });
+        const applicationsCourse = await db.getFirst(applicationsCourseTable, {
+          sortBy: 'id',
+          filter: { courseBuilderId: courseId },
+        });
+
+        if (!applicationsCourse) {
+          throw new createHttpError.InternalServerError(`Course configuration not found for course: ${courseId}`);
+        }
 
         const newCourseRegistration = await db.insert(courseRegistrationTable, {
           email: auth.email,


Will-Howard:
Sidenote regarding the P1-ness and urgency of this bug: I think this issue has been around for a long time, there are examples of non-MOOC course registrations created this way going back several months (with `decision: 'Accept'` but no round).

So I wasn't treating this as that urgent to fix, and wasn't intending this as a stop-gap measure. I checked all uses of the course registration object to see if they still work ok without the course registration being guaranteed to exist. `CertificateLinkCard` was the only one that had an issue, which is fixed here.

So I think this should be fine as a ~long term solution, provided the assumption about course registrations being created via the application form is correct

joshestein:
> > Once this PR is merged, how will course registrations be made for user-accepted courses since I joining discussions requires registrationId (At least, from the top of my head it does)?
> 
> The application form inserts a row in the course registrations table. I believe this will mean we'll never need to create course registrations for non-MOOC courses, though @anglilian or @dewierwan can confirm?

@anglilian or @dewierwan can you please confirm?

PR #1515, opened 2025-10-28
Public `certificates.create` does not rely on admin email

joshestein:
# Description

This endpoint will be called by an Airtable automation when row conditions are met, not necessarily via a user calling the script (via checking a checkbox). Because of this we cannot guarantee that a user email will be provided. We still rely on the shared secret for authorisation.

## Issue

Part of #1457.

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/apps/website/src/server/trpc.ts
+++ b/apps/website/src/server/trpc.ts
@@ -101,7 +101,7 @@ const openTelemetryMiddleware = t.middleware(async (opts) => {
   }
 });
 
-export const checkAdminAccess = async (email: string) => {
+const checkAdminAccess = async (email: string) => {
   try {
     const admin = await db.pg.select()
       .from(adminUsersTable)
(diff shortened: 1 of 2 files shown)


apps/website/src/server/trpc.ts
```diff
@@ -101,7 +101,7 @@ const openTelemetryMiddleware = t.middleware(async (opts) => {
   }
 });
 
-export const checkAdminAccess = async (email: string) => {
+const checkAdminAccess = async (email: string) => {
```
marn-in-prod: Nit: Can we just inline this with adminProcedure  not that it's only being used in one place?

marn-in-prod (approved):
LGTM

apps/website/src/server/trpc.ts
joshestein: I would prefer to keep it extracted as a function. Even though it's not reused I think the abstraction is a helpful one.

PR #1517, opened 2025-10-28
tRPC migrate `/api/group-discussions/[id]`

### LINKED ISSUE #1485: [Bug] 'No upcoming discussions' incorrectly shown
### 🧩 Summary

In `/settings/courses` when course details are being fetched there is an intermediary stage where our loading state is finished but we haven't finished processing the data. This means we show 'No upcoming discussions' when there are actually upcoming courses.

https://github.com/user-attachments/assets/e94b21af-7c5c-4f15-98d9-7663f4b03ae7

### 🛝 Steps to reproduce

1. Go to /settings/courses while logged in and enrolled
2. See 'no upcoming discussions' flash before course details

### ✅ Definition of done

'No upcoming discussions' is not flashed between loading and processing.

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P3" => "Low"}

joshestein:
# Description

Part of ongoing migration to tRPC.

Backend changes:

1. Allow for passing multiple discussion IDs and querying the DB in one go, instead of sequentially passing in IDs and querying the DB for each discussion ID
2. Don't return `groupId` as part of the return payload, it is not used by any callers

Component changes:

1. Remove state handlers for `expectedDiscussions`, `attendingDiscussions` (derive state from `useQueries` results)

## Issue

Part of #1260 and #1427. Fixes #1485.

## Developer checklist

NA

## Screenshot

NA


apps/website/src/server/routers/groupDiscussionsRouter.ts
```diff
@@ -0,0 +1,51 @@
...
+      if (!discussion.courseBuilderUnitRecordId) {
+        throw new TRPCError({ code: 'INTERNAL_SERVER_ERROR', message: 'Discussion is missing unit reference' });
+      }
+
+      const group = await db.getFirst(groupTable, {
+        filter: { id: discussion.group },
+      });
+
+      if (!group) {
+        throw new TRPCError({ code: 'INTERNAL_SERVER_ERROR', message: 'Related group not found' });
```
joshestein: This (and the unit not found below) were setting `raw.res.status(500)`, so I kept as `INTERNAL_SERVER_ERROR` for consistency but they should maybe be `NOT_FOUND`?

apps/website/src/components/settings/CourseListRow.tsx
```diff
@@ -36,55 +33,32 @@ const CourseListRow = ({
...
+  // Use expectedDiscussionsFacilitator if the user is a facilitator, otherwise use expectedDiscussionsParticipant
+  const isFacilitator = courseRegistration.role === 'Facilitator';
+  const expectedDiscussionIds = isFacilitator
+    ? meetPerson?.expectedDiscussionsFacilitator || []
+    : meetPerson?.expectedDiscussionsParticipant || [];
 
-      const expectedResults = await Promise.all(expectedPromises);
+  const expectedResults = trpc.useQueries((t) => {
+    return expectedDiscussionIds.map((id) => t.groupDiscussions.getByDiscussionId({ discussionId: id }));
+  });
```
marn-in-prod: Why not an endpoint that takes multiple IDs? I understand that it's a 1:1 migraiton but feels like a minor but overall beneficial change to make

apps/website/src/server/routers/groupDiscussionsRouter.ts
marn-in-prod: +1 to not found

apps/website/src/server/routers/groupDiscussionsRouter.ts
```diff
@@ -0,0 +1,51 @@
...
+      });
+
+      if (!unit) {
+        throw new TRPCError({ code: 'INTERNAL_SERVER_ERROR', message: 'Associated unit not found' });
+      }
+
+      return {
+        discussion: {
+          ...discussion,
+          groupId: discussion.group,
```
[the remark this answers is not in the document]
marn-in-prod: We should just go ahead and remove it if it's not being used. Now with tRPC, we can actually easily tell if it's not being used! haha

marn-in-prod (approved):
Generally looks great!

I think all of my questions are more-so around non 1:1 parity stuff but I think are easy changes to make

apps/website/src/server/routers/groupDiscussionsRouter.ts
```diff
@@ -0,0 +1,79 @@
...
+        return { discussions: [] };
+      }
+
+      const discussions = await db.scan(groupDiscussionTable, {
+        OR: discussionIds.map((id) => ({ id })),
+      });
+
+      if (discussions.length === 0) {
+        return { discussions: [] };
+      }
```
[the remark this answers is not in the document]
marn-in-prod: It feels like it should throw an error that we can get alerted on.

marn-in-prod (approved):
Generally looks good, just take a look at the comment. 

Do confirm that this is working locally before pushing as well.

PR #1530, opened 2025-10-29
tRPC migrate `/api/courses`

joshestein:
# Description

Part of ongoing migration to tRPC. Migrate the API handler responsible for fetching all active courses.

## Issue

Part of #1260 and #1427.

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/apps/website/src/components/courses/CourseSearchCard.stories.tsx
+++ b/apps/website/src/components/courses/CourseSearchCard.stories.tsx
@@ -19,8 +19,6 @@ export const Default: Story = {
     title: 'Advanced TypeScript',
     url: '/courses/advanced-typescript',
     description: 'Deep dive into advanced TypeScript features like generics, decorators, and conditional types.',
-    courseLength: 'Self-paced',
-    cadence: 'Daily',
     averageRating: 4.8,
     imageSrc: '/images/courses/default.jpg',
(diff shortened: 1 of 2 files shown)


apps/website/src/components/courses/CourseDirectory.tsx
```diff
@@ -1,15 +1,15 @@
...
 } from '@bluedot/ui';
-import type { GetCoursesResponse } from '../../pages/api/courses';
+import {
+  FC,
+} from 'react';
 import { CourseSearchCard } from './CourseSearchCard';
 
 export type CourseDirectoryProps = {
-  courses: GetCoursesResponse['courses'] | undefined;
+  courses?: Course[];
```
joshestein: Maybe this should be inferred from the router output instead.

marn-in-prod (approved):
The migration looks good, comment is in regards to extra stuff outside of the migration/

apps/website/src/components/courses/CourseDirectory.tsx
joshestein: The course pages injects some course information into the document head (I guess for SEO and rich link rendering?).

Because of this I'm going to keep passing the courses as a prop but will update to use router output.

PR #1531, opened 2025-10-29
Cleanup course components outdated BEM classes

joshestein:
# Description

While working on #1530 I noticed a lot of outdated BEM classes that are no longer used.

1. Remove component BEM classes
2. Update tests to use `getByRole`/`getByText` to find elements
3. Small other cleanups (removing unused vars, unused props, fixing imports)

## Issue

NA

## Developer checklist

NA

## Screenshot

NA


marn-in-prod (approved):
Generally looks good but I wonder if we need to add comments to delineate major sections by removing these. 

I think this is fine for now and that's something people can do in the future when they're building new components

PR #1539, opened 2025-10-30
Remove `ResourceListCard`, unused `/api/courses/[courseSlug]/[unitNumber]/`  API routes

joshestein:
# Description

I started to migrate `/api/courses/[courseSlug]/[unitNumber]/index.ts` to tRPC as part of #1427.

I realised the actual API endpoint is only used by one component - `ResourceListCourseCard`. As far as I can see this component is actually unused. So I deleted it and removed the API route handler.

The component is the only one to call `/api/courses/[courseSlug]/[unitNumber]/resources`, so I have removed this API handler as well.

## Issue

Part of #1427.

## Developer checklist

NA

## Screenshot

NA


marn-in-prod (approved):
Looks good and I replied to your own comment

joshestein:
@marn-in-prod `[[...chunkNumber]].ts` was getting too chunky with all the extra types and definition of `getUnitWithContent`, so I moved it into `/lib/api/utils` instead.

apps/website/src/lib/api/utils.ts
```diff
@@ -44,3 +50,74 @@ export async function unitFilterActiveChunks({
...
+  const allUnitsWithAllChunks = await db.scan(unitTable, { courseSlug, unitStatus: 'Active' });
+  const allUnits = await unitFilterActiveChunks({ units: allUnitsWithAllChunks, db });
+
+  // Sort units numerically since database text sorting might not handle numbers correctly
+  const units = allUnits.sort((a, b) => parseInt(a.unitNumber) - parseInt(b.unitNumber));
+
+  const unit = units.find((u) => parseInt(u.unitNumber) === parseInt(unitNumber));
+  if (!unit) {
+    throw new TRPCError({ code: 'NOT_FOUND', message: 'Unit not found' });
+  }
```
joshestein: Changed to return TRPCError, not `creatHTTPError`

joshestein:
@greptileai review

apps/website/src/lib/api/utils.ts
```diff
@@ -44,3 +50,71 @@ export async function unitFilterActiveChunks({
     chunks: unit.chunks?.filter((c) => allActiveChunkIds.has(c)) ?? [],
   }));
 }
+
+export type UnitWithContent = Awaited<ReturnType<typeof getUnitWithContent>>;
+export async function getUnitWithContent(courseSlug: string, unitNumber: string) {
```
marn-in-prod: I understand that this was previously called getUnitWithContent but that's not what it does anymore. We should rename it.

This is because unit Content is a specific "thing" inside of Airtable and a specific thing when we'd communicate to each other and this function is returning more things that just that.

apps/website/src/lib/api/utils.ts
```diff
@@ -20,15 +26,15 @@ type Unit = InferSelectModel<typeof unitTable.pg>;
  */
 export async function unitFilterActiveChunks({
   units,
-  db,
+  db: dbInstance,
```
marn-in-prod: Nit: Can we remove the dbInstance renaming? Parameter shadowing of module imports is perfectly fine and conventional especially with something like ``db``

apps/website/src/lib/api/utils.ts
```diff
@@ -1,12 +1,18 @@
...
+  unitResourceTable,
   unitTable,
+  type Exercise,
+  type UnitResource,
 } from '@bluedot/db';
+import { TRPCError } from '@trpc/server';
 import type dbAsType from './db';
+import db from './db';
 
 type DB = typeof dbAsType;
```
marn-in-prod: I think the typing situation here is overly complicated.

Can't we delete these three lines and just replace them with
```
import db from './db';

type DB = typeof db;
```

Or potentially simpler
```
import { PgAirtableDb} from '@bluedot/db';

import db from './db';
```
And you have the class and the actual instance itself.

apps/website/src/lib/api/utils.ts
```diff
@@ -1,12 +1,18 @@
...
 type Unit = InferSelectModel<typeof unitTable.pg>;
```
marn-in-prod: This can be just imported from the ``schema.ts`` file

apps/website/src/lib/api/utils.ts
```diff
@@ -44,3 +50,71 @@ export async function unitFilterActiveChunks({
...
+
+  const unit = units.find((u) => parseInt(u.unitNumber) === parseInt(unitNumber));
+  if (!unit) {
+    throw new TRPCError({ code: 'NOT_FOUND', message: 'Unit not found' });
+  }
+
+  // Get chunks for this unit, filter for active status, and sort by chunk order
+  const allChunks = await db.scan(chunkTable, { unitId: unit.id });
+  const activeChunks = allChunks.filter((chunk) => chunk.status === 'Active');
+  const chunks = activeChunks.sort((a, b) => (a.chunkOrder || '').localeCompare(b.chunkOrder || '', undefined, { numeric: true, sensitivity: 'base' }));
```
marn-in-prod: Nit: I think you're just moving this and not your fault but can we standardize between parseInt and localcompare?  There's two different methods within the same function for very similar data.

apps/website/src/lib/api/utils.ts
marn-in-prod: `utils.ts`` becomes the default place for new code too easily. I don't agree with utils in this scenario because it's just a single call site. 

I think the options are to either make ``getUnitWithContent`` inline with the component that's using it or putting it inside of trpc and creating a new caller so that we don't actually make the api call. I think I prefer inline since it's really just a bunch of one-off logic that doesn't get called from anywhere else.

joshestein: I moved it there because that is where the definition of `unitFilterActiveChunks` was, which this code depends on.

Now I've moved it into [[...chunks]].

apps/website/src/pages/courses/[courseSlug]/[unitNumber]/[[...chunkNumber]].tsx
Comment (marn-in-prod) [inline apps/website/src/pages/courses/[courseSlug]/[unitNumber]/[[...chunkNumber]].tsx]: Nit: Can we move const CourseUnitChunkPage to be the first function then getServerSideProps and then getUnitWithChunks?

marn-in-prod (approved):
Generally looks good. Just minor nits really

joshestein:
@greptileai final review

PR #1541, opened 2025-10-30
Remove `/api/courses/[courseSlug]/[unitNumber]/feedback` API handler

joshestein:
# Description

No components use this API route. I think it's dead code?

## Issue

Part of #1427.

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/libraries/db/src/schema.ts
+++ b/libraries/db/src/schema.ts
@@ -156,6 +156,7 @@ export const courseTable = pgAirtable('course', {
   },
 });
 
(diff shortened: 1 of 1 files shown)


joshestein:
@marn-in-prod it seems that entire table is not being used at all? Should I remove the entire table, or just the fields you've specified?

marn-in-prod:
@joshestein If you can confirm that the table isn't being used at all then you could remove it. You might want to remove it when you know you're going to be available to revert in case it does break something

PR #1566, opened 2025-11-03
Prevent log spam: Return 200 and undefined for missing resource completion or exercise response, rather than 404

### LINKED ISSUE #1462: [Bug] Resource completion not found
### 🧩 Summary

Our logs are being completely spammed with: 'Client error handling request: Resource completion not found'

It seems we are somehow requesting resource completion resource that don't exist.

Figure out where the non-existent IDs are being saved and cleanup these rows.

### 🛝 Steps to reproduce

1. Go to https://grafana.k8s.bluedot.org/a/grafana-lokiexplore-app/explore/service/bluedot-website-production/logs?patterns=%5B%5D&from=now-15m&to=now&var-lineFormat=&var-ds=fekp33muo5pfka&var-filters=service_name%7C%3D%7Cbluedot-website-production&var-fields=&var-levels=&var-metadata=&var-jsonFields=&var-patterns=&var-lineFilterV2=&var-lineFilters=caseInsensitive,0%7C__gfp__%3D%7Cresource%20completion%20not%20found&timezone=browser&var-all-fields=&displayedFields=%5B%5D&urlColumns=%5B%5D&visualizationType=%22logs%22&prettifyLogMessage=false&sortOrder=%22Descending%22&wrapLogMessage=false
2. Observe the hideousness of so many logs

### ✅ Definition of done

1. We know why outdated resources are being fetched
2. Broken records are cleaned up. Alternatively, resources that we know do not exist are not fetched

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

Will-Howard:
# Description
As noted in #1462, it is expected that resource completions don't exist if the user hasn't visited the page before so return undefined rather than returning a 404 (which was handled already, but spammed the logs with "Resource completion not found").

There was an almost identical case with exercise responses (they don't exist if the user hasn't visited the page, and trigger a 404), so I fixed that in this PR too. The change is simple enough that I don't think it's worth adding tests.

## Issue

Fixes #1462

## Developer checklist

## Screenshot

N/A

### DIFF
--- a/apps/website/src/pages/api/courses/exercises/[exerciseId]/response.ts
+++ b/apps/website/src/pages/api/courses/exercises/[exerciseId]/response.ts
@@ -51,17 +51,13 @@ export default makeApiRoute({
   switch (raw.req.method) {
     // Get exercise response
     case 'GET': {
-      if (!exerciseResponse) {
-        throw new createHttpError.NotFound('Exercise response not found');
-      }
-
       return {
         type: 'success' as const,
-        exerciseResponse: {
+        exerciseResponse: exerciseResponse ? {
           ...exerciseResponse,
           // Preserve trailing whitespace and line feeds as entered by the user
           response: exerciseResponse.response,
-        },
+        } : undefined,
       };
     }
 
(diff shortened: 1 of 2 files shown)


marn-in-prod (approved):
This is fine for now especially considering we do have a trpc migration for this.

Can you just confirm that locally this still works as intended before submitting?

PR #1591, opened 2025-11-07
Make mobile CourseCarousel respond to user swipes

### LINKED ISSUE #1589: [Bug] Homepage mobile course carousel not responding to swipes
### 🧩 Summary

The mobile-responsive course carousel does not respond to user swipes.

### 🛝 Steps to reproduce

1. Go to bluedot.org
2. Mobile responsiveness
3. Try to swipe the course carousel, it does not respond

### ✅ Definition of done

Mobile responsive course carousel responds to swipes

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

joshestein:
# Description

This implementation is copied from `PhotoCarousel`. It is the same but doesn't have any mouse handlers, since this component is only shown on mobile devices.

Personally I don't think this component should auto-scroll at all. It's hard to read the information on mobile devices. I think it would make more sense to only progress on swipe / button press.

A quick browse through our codebase shows that we have several carousels, all with slightly different behaviours. I would strongly suggest that:

1. We extract a standardised `Carousel` component (or use an off-the-shelf one)
2. All carousels conform to having arrow buttons for navigation

## Issue

Fixes #1589.

## Developer checklist

## Screenshot

| 📸 | Before | After |
|---------|---|---|
| 📱 | before | after |

### DIFF
--- a/apps/website/src/components/homepage/CourseSection.tsx
+++ b/apps/website/src/components/homepage/CourseSection.tsx
@@ -164,19 +164,19 @@ const CourseCarousel = ({
   const infiniteCourses = createInfiniteScrollData();
 
   // Card width + gap
-  const getCardWidth = useCallback(() => {
+  const getCardWidth = () => {
     if (typeof window === 'undefined') return 276;
     // 768px is TW md breakpoint
     return window.innerWidth >= 768 ? 400 : 276;
-  }, []);
+  };
 
-  const getGap = useCallback(() => {
+  const getGap = () => {
     if (typeof window === 'undefined') return 20;
     const width = window.innerWidth;
     if (width >= 1024) return 32;
     if (width >= 768) return 24;
     return 20;
-  }, []);
+  };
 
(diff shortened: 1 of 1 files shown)


marn-in-prod:
> All carousels conform to having arrow buttons for navigation

I'd defer to Alex here, but I generally disagree with adding arrow buttons to all carousels. For example, the course cards don't need them IMO, swiping works fine there. Same with the events photos on the homepage. Those feel more like "style" elements rather than something people are meant to actively navigate through.

> We extract a standardized Carousel component (or use an off-the-shelf one)

I'm generally fine with this; although, I don't think I'd want to actively migrate existing components to it unless we're adding new functionality (like we did here), or there was some bug that needed fixing that instigated the move to a pre-built (library) one.

joshestein:
As I haven't heard back from anyone I'm going to merge this and create a separate issue / PR for the nit comments.

PR #1597, opened 2025-11-10
[Feature] Install eslint-plugin-react-hooks, adopt in apps/website, libraries/ui

Will-Howard:
# Description

I've put up this PR to adopt react-hooks rules first in website/ and ui/ first, to simplify reviewing it as there are some places with non-trivial changes.

I'll put up a followup to adopt it in the rest of the repo, I have most of those changes stashed already.

## Issue

#1488

## Developer checklist

- [x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [x] Considered having snapshot tests and/or happy path functional tests
- [x] Considered adding Storybook stories

PR #1603, opened 2025-11-11
[Feature] Adopt eslint-plugin-react-hooks throughout repo

### LINKED ISSUE #1488: [Feature] ESLint react rules
### 🧩 Summary

We don't have warnings when we don't list all exhaustive React dependencies (e.g. in `useEffect`) because we don't have the ESLint rule for 'exhaustive-deps' (https://react.dev/reference/eslint-plugin-react-hooks/lints/exhaustive-deps) setup.

We should definitely have this!

Consider first upgrading to ESLint v9: https://github.com/bluedotimpact/bluedot/issues/120

### ✅ Definition of done

Add https://www.npmjs.com/package/eslint-plugin-react-hooks

Ensure we get an ESLint warning if you create an effect without explicitly listing all dependencies

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

Will-Howard:
# Description

Followup to #1597 which added eslint-plugin-react-hooks. This PR clears up the remaining places where there were ignored errors (or more permanently ignores them in some cases that were difficult to fix).

## Issue

Fixes #1488 

## Developer checklist

### DIFF
--- a/apps/editor/src/components/BodyEditor.tsx
+++ b/apps/editor/src/components/BodyEditor.tsx
@@ -23,7 +23,6 @@ export const BodyEditor: React.FC<BodyEditorProps> = ({ auth, children, onSave }
     if (body === null && children) {
       setBody(children);
     }
-    // TODO note that I haven't tested, test in prod once deployed
   }, [body, children]);
 
   // Add beforeunload event listener to warn about unsaved changes


apps/meet/src/components/SelectPersonView.tsx
```diff
@@ -38,32 +40,17 @@ const SelectPersonView: React.FC<SelectPersonViewProps> = ({ page: { groupId },
     );
   }
 
-  if (data.type === 'redirect') {
```
Will-Howard: This was dead code as the 'redirect' case was no longer present in the API route

apps/meet/src/components/SelectPersonView.tsx
```diff
@@ -16,6 +17,7 @@ export type SelectPersonViewProps = {
 };
 
 const SelectPersonView: React.FC<SelectPersonViewProps> = ({ page: { groupId }, setPage }) => {
+  const currentTimeMs = useCurrentTimeMs();
```
joshestein: Just to make sure I'm 100% understanding - this will force this component to re-render every 30s, right? Contrasting to this before where using `Date.now()` would only cause one render?

joshestein (approved):
LGTM 👍

apps/meet/src/components/SelectPersonView.tsx
Will-Howard: Yep that's right 👍

apps/frontend-example/src/pages/dewi.tsx
```diff
@@ -1,13 +1,14 @@
...
+const images = ['bdi.png', 'aisf.png', 'bsf.png']; // Array of images
+
 const Dewi = () => {
   const [value, setValue] = useState(0);
-  const colors = ['red', 'blue', 'green', 'yellow', 'purple', 'orange']; // Add more colors as needed
-  const images = ['bdi.png', 'aisf.png', 'bsf.png']; // Array of images
 
   // Function to get the next color based on the current value
-  const getNextColor = () => colors[value % colors.length]!;
+  const getNextColor = useCallback(() => colors[value % colors.length]!, [value]);
```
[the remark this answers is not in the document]
Will-Howard: This function being recreated every render would cause the `useEffect` below to trigger on every render, to the `useCallback` is still needed

PR #1621, opened 2025-11-14
Allow users to give detailed resource feedback

### LINKED ISSUE #1432: [Design] Allow users to provide written feedback after the rating
### 🎨 Summary

It provides us with much richer feedback if users can tell us what exactly they did or did not find useful with the resource.

Image

### ✅ Definition of done

Design for including a text field that nudges users to provide more details on their rating.

FYI @herrhaase 

### 📌 Notes / References

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

anglilian:
# Description

- Turned the autosave text area into a component
- Converted the free text response component to use the autosave text component
- Added free text response to resource feedback

I'm least confident about the tests that were generated. 😅 

## Issue

Fixes #1432 

## Developer checklist

- [ x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [x ] Considered having snapshot tests and/or happy path functional tests

## Screenshot

| 📸 |  |
|---------|---|
| 📱  | 
CleanShot 2025-11-14 at 15 00 07
 |
| 🖥️ | 
CleanShot 2025-11-14 at 14 46 39
 |

### DIFF
--- a/apps/website/src/components/courses/ResourceListItem.tsx
+++ b/apps/website/src/components/courses/ResourceListItem.tsx
@@ -166,8 +166,8 @@ export const ResourceListItem: React.FC<ResourceListItemProps> = ({ resource }) 
   const handleFeedback = useCallback((feedbackValue: ResourceFeedbackValue) => {
     // Toggle off if clicking the same feedback button
     const newFeedback = resourceFeedback === feedbackValue ? RESOURCE_FEEDBACK.NO_RESPONSE : feedbackValue;
-    handleSaveCompletion(true, newFeedback);
-  }, [resourceFeedback, handleSaveCompletion]);
+    handleSaveCompletion(true, newFeedback, feedback || '');
+  }, [resourceFeedback, feedback, handleSaveCompletion]);
 
   if (completionLoading) {
     return <ProgressDots />;


apps/website/src/components/courses/exercises/AutoSaveTextarea.tsx
```diff
@@ -0,0 +1,243 @@
...
+      });
+    };
+  }, []);
+
+  // Inject style for Firefox to hide drag notches if needed
+  useEffect(() => {
+    if (!hideResizeHandle) return undefined;
+
+    const styleId = 'auto-save-textarea-firefox-styles';
+    if (document.getElementById(styleId)) return undefined;
```
joshestein: This ID doesn't seem to be used anywhere, so I think this element will ever be found and we will never inject these custom styles (unless I'm missing where this is defined)

apps/website/src/components/courses/exercises/AutoSaveTextarea.tsx
```diff
@@ -0,0 +1,243 @@
...
+  value,
+  onChange,
+  onSave,
+  placeholder = 'Enter text here',
+  minHeight = '140px',
+  className,
+  disabled = false,
+  showSaveStatus = true,
+  autoSaveDelay = 20000, // 20 seconds
+  periodicSaveInterval = 180000, // 3 minutes
```
marn-in-prod: Nit: Mention the timescale for these parameters/variables.

So something like ``autoSaveDelayInMs`` and ``periodicSaveIntervalInMs``

apps/website/src/components/courses/exercises/AutoSaveTextarea.tsx
```diff
@@ -0,0 +1,243 @@
...
+  minHeight?: string;
+  className?: string;
+  disabled?: boolean;
+  showSaveStatus?: boolean;
+  autoSaveDelay?: number; // milliseconds for inactivity timer (default 20000)
+  periodicSaveInterval?: number; // milliseconds for periodic saves (default 180000)
+  ariaLabel?: string;
+  hideResizeHandle?: boolean;
+  savedText?: string; // Custom text for saved state (default: "Answer saved")
+};
```
marn-in-prod: Some of these parameters are really just constants in the codebase or can be combined (minHeight is really just a className parameter), given that there's only two callers for this. I think it should be simplified to something like this. Otherwise, I think it's just over-engineering the component

 type AutoSaveTextareaProps = {
    value: string;
    onChange: (value: string) => void;
    onSave: (value: string) => Promise<void>;
    placeholder?: string;
    className?: string;
    disabled?: boolean;
  };

anglilian:
Do you think I need to apply [this fix](https://github.com/bluedotimpact/bluedot/pull/1621#pullrequestreview-3474849156) @marn-in-prod ? I tested it and I don't get this overwriting error.

Otherwise, maybe it's good to go?

marn-in-prod (approved):
Looks much better! I do get a minor bug where the textbox clears itself if I start typing right after liking/disliking but otherwise looks good

anglilian:
Yay!! I've fixed the minor bug and addressed the nit comment so I'll merge this now. Thank you Martin :)

PR #1640, opened 2025-11-17
[Bug] Remove confusing Join group button on settings page

Will-Howard:
# Description

Discussed in Slack: This button was resulting in a lot of slack warnings, due to people clicking it after they had been accepted but before they had a group assigned. From looking at the metrics in posthog ([notebook](https://eu.posthog.com/project/21762/notebooks/HhDk8Bin)), it appears that there are few if any legit uses of this button and quite a lot due to user confusion, so we're removing it.

@anglilian requested "We're working on assigning you to a group. You'll receive an email from us within the next few days." as replacement text. I went with "We\'re assigning you to a group, you'll receive an email from us within the next few days" to make it fit on one line.

## Issue

No issue

## Developer checklist

## Screenshot

| 📸 |  |
|---------|---|
| 📱  | Screenshot 2025-11-17 at 18 51 58 |
| 🖥️ | Screenshot 2025-11-17 at 18 51 50 |

### DIFF
--- a/apps/website/src/components/settings/CourseListRow.tsx
+++ b/apps/website/src/components/settings/CourseListRow.tsx
@@ -105,7 +105,7 @@ const CourseListRow = ({
   const primaryCtaButton = getPrimaryCtaButton();
 
   const getSubtitle = (): ReactNode | null => {
-    if (!isCompleted && nextDiscussion && !isLoading) {
+    if (!isCompleted && nextDiscussion) {
       if (isExpanded || isLoading) return null;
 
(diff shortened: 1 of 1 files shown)


anglilian:
I can see the text which is probably not supposed to happen @Will-Howard 

CleanShot 2025-11-17 at 22 46 43

Will-Howard:
@anglilian Ah good catch, that was a bug that would happen for facilitators. I've fixed that now and also combined all the possible subtitles into one `getSubtitle` function to avoid this sort of thing.

Note this also means changing some of the colours as the different subtitles were using slightly different shades of grey before, these are now all the possibilities (including the "Completed..." one below):

Screenshot 2025-11-18 at 07 43 04
Screenshot 2025-11-18 at 07 43 25

PR #1646, opened 2025-11-18
[Bug] Change undefined to null to fix JSON serialisation error

Will-Howard:
# Description

We were getting this error on the live site due to `getResourceCompletion` returning undefined, which isn't JSON serialisable
image


Will-Howard:
I'm going to merge this so that we can safely deploy again

PR #1650, opened 2025-11-19
[Bug] Remove tests for Join group button on settings page (button has been removed)

Will-Howard:
# Description

This is causing CI to fail due to a semantic conflict between  #1615 and #1640. Otherwise the code itself in `CourseListRow` is still correct as far as I can tell.


Will-Howard:
Ok I'm going to merge this one now to clear this up!

PR #1656, opened 2025-11-20
Upgrade Zod to v4 after revert

### LINKED ISSUE #1533: [Feature] Upgrade to Zod v4
### 🧩 Summary

Zod v4 is [significantly faster](https://zod.dev/v4?id=benchmarks) and has a reduced bundle size. We should upgrade.

### ✅ Definition of done

We are using Zod v4, not v3.

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

joshestein:
# Description

In https://github.com/bluedotimpact/bluedot/pull/1562 we upgraded Zod to v4. This change was [reverted](https://github.com/bluedotimpact/bluedot/pull/1637) since it led to broken CD: https://github.com/bluedotimpact/bluedot/actions/runs/19424561714/job/55574428338 and I didn't want to block deployment. CD fails during `npm ci`. The core problem is that our `package-lock.json` had a root dependency on Zod v3.25.76. But `package.json` specifies v4. These are in conflict, so CD bombs out.

I updated root Zod (in `package-lock.json`) to v4. I thought that would do the trick but there were still problems with `@ts-rest/core` and `@ts-rest/fastify`.

To fix I've used `turbo prune`, which [generates a partial monorepo for a target package](https://turborepo.com/docs/reference/prune). This is better because instead of relying on the root `package-lock.json` we create a new `package-lock.json` for `pg-sync-service`, which specifies exactly what this package needs to be built. I think we should make this change across all sub-repos since it should lead to faster build times (instead of needing to install _all_ packaged used by any sub-repo only install the packages that each sub-repo needs).

To test I ran `npx turbo prune --scope=@bluedot/pg-sync-service --docker --out-dir=apps/pg-sync-service/dist_tools` from our root dir, then `cd apps/pg-sync-servce/dist_tools/json` and `npm ci --omit dev`. This succeeds.

## Issue

Fixes #1533

## Developer checklist

## Screenshot

NA

### DIFF
--- a/apps/website/src/components/courses/GroupDiscussionBanner.stories.tsx
+++ b/apps/website/src/components/courses/GroupDiscussionBanner.stories.tsx
@@ -71,6 +71,18 @@ const meta = {
 export default meta;
 type Story = StoryObj<typeof meta>;
 
+// Participant view with discussion live
+export const ParticipantLive: Story = {
+  args: {
+    userRole: 'participant',
+    groupDiscussion: {
+      ...mockGroupDiscussion,
+      startDateTime: Math.floor(Date.now() / 1000) - 600, // Started 10 minutes ago (LIVE)
+      endDateTime: Math.floor(Date.now() / 1000) + 2700, // Ends in 45 minutes
+    },
+  },
+};
+
 // Participant view with discussion starting soon
 export const ParticipantStartingSoon: Story = {
   args: {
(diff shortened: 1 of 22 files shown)


joshestein:
@marn-in-prod I kind of disagree. [The gains are significant.](https://zod.dev/v4#benchmarks) Especially because we've now migrated to tRPC and are using Zod more extensively, I think it's worth trying to get as much benefit and speedup as possible.

The only thing failing was a ts-rest incompatibility specifically within `pg-sync-service`; it was only failing because of dependencies from another app. Pruning the dependencies appropriately doesn't lead to any problems and as I've argued in the PR description we should be doing anyways.

But more importantly, the ts-rest packages are only used for outdated backends. The README.md int `apps/backend` says that it is no longer used. The `apps/frontend-example` can easily (and should be) updated. And the only other package that depends on ts-rest is `pg-sync-service`. If you dive into the code you will see it is to spinup a server that has one endpoint, `/hello`. I honestly don't think this is ever used and should probably be removed. If you disagree and want to keep it there is [a builtin fastify adapter for tRPC](https://trpc.io/docs/server/adapters/fastify) that we should probably prefer to using `ts-rest`.

If you prefer me to create a separate PR that first removes our dependencies on ts-rest I am happy to do that.

marn-in-prod:
> @marn-in-prod I kind of disagree. [The gains are significant.](https://zod.dev/v4#benchmarks) Especially because we've now migrated to tRPC and are using Zod more extensively, I think it's worth trying to get as much benefit and speedup as possible.

Which gains specifically? Since you linked to the benchmarks section, I'm assuming you mean runtime parsing performance. But CPU utilization isn't a bottleneck for us, and Zod parsing is likely a small fraction of our overall CPU usage anyway.

My concern with upgrading is the ecosystem risk. Zod 4 is new, and we're likely to hit compatibility issues with libraries that haven't caught up yet, leading to strange errors or blocked integrations. This isn't specifically about ts-rest; it's a general concern with breaking changes in the JS/React ecosystem.

The type checking speedups could be something, but I'm skeptical we'd notice them given that we're not doing anything particularly complex with Zod. If slow IDE performance has been a pain point for you, that's worth discussing but it wasn't mentioned in the original PR description or earlier.

To summarize: the benefits you've highlighted (faster parsing, 7KB smaller bundle) aren't addressing actual bottlenecks we have, while the upgrade carries real compatibility risk. I'd prefer to wait until the ecosystem matures

marn-in-prod (changes requested):
See New comment - tag me again for review if you've got new questions

joshestein:
Closing until we are more confident that v4 is mature and well-supported by our dependencies.

PR #1670, opened 2025-11-21
Dynamic OG preview links for courses and jobs

### LINKED ISSUE #1290: [Feature] Auto-generated open graph preview images for roles and courses
### 🧩 Summary

Rather than having to generate the link preview image ourselves, could we have this be auto-generated for our roles and courses? 

Example:
Image

### ✅ Definition of done

The link preview shows the BlueDot logo and for our job shows the job title and subtitle while for courses, it shows the course name and short course description.

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P4" => "Trivial/Wishlist"}

joshestein:
# Description

OG tags are important for previews on social media etc. [According to the protocol](https://ogp.me/#metadata) there are 4 required tags: title, type, image and url. Also,[ images should only be one of PNG, JPG or GIF](https://developers.facebook.com/docs/sharing/webmasters/#images). 

New OG tags are added for job postings and standard courses.

Also makes the following improvements:

1. Adds a new PNG for our logo (needed since we cannot use our SVG as image type)
2. New images for agi strategy, alignment, technical ai safety, biosecurity, economics of ai, FoAI, governance courses
3. New images for biosecurity specialist jobs
4. Fallback to new logo PNG if course / job image not found
5. Missing `og:url` (required) for updated course landers
6. Missing `og:image` for blogs, use new logo PNG
7. Invalid image URLs for updated course landers, requires full path (not relative to base site)
8. Missing OG tags for Ops lander
9. Missing OG tags for FoAI lander

Honestly I'm not really a fan of the approach I took of having a separate prop for `ogImage` and depending on `images/<type>/link-preview/slug.png` for finding images. First prize would be updating the Airtable library to handle the 'attachment' type column, so that we could pull images directly from Airtable (maybe we can already do this but I wasn't sure how). Otherwise maybe we can take an approach that is somehow less fragile. The coures lander image approach is to define the path to the image as a field on the course, but this also feels brittle to me.

## Issue

Fixes #1290

## Developer checklist

## Screenshot

This is a bit hard to test without deploying. I built our website and validated that expected `og:image` tags exist for job postings and courses, and that they point to the new images. Using the[ FB debugger](https://developers.facebook.com/tools/debug/), this is what I am seeing (before merging - this is just for reference so we know what we are working with).

https://bluedot.org/courses/agi-strategy (new course lander) - uses first found image (not explicit):
image

https://bluedot.org/courses/alignment (standard course) - uses first found image (not explicit):
image

https://bluedot.org/join-us/biosecurity (job posting) - no image, no description:
image

https://bluedot.org/blog/course-portfolio-vision (blog) - no image:
image

### DIFF
--- a/apps/website/src/__tests__/pages/courses/[courseSlug]/index.test.tsx
+++ b/apps/website/src/__tests__/pages/courses/[courseSlug]/index.test.tsx
@@ -66,4 +66,16 @@ describe('CoursePage SSR/SEO', () => {
     const ogUrl = document.querySelector('meta[property="og:url"]');
     expect(ogUrl?.getAttribute('content')).toBe(`https://bluedot.org/courses/${encodeURIComponent(mockCourse.slug)}`);
   });
+
+  test('uses fallback logo image when no `courseOgImage` provided', () => {
+    renderWithHead(
+      <CoursePage
+        courseSlug="ai-safety-fundamentals"
+        courseData={{ course: mockCourse, units: mockUnits }}
+      />,
+    );
+
+    const ogImage = document.querySelector('meta[property="og:image"]');
+    expect(ogImage?.getAttribute('content')).toBe('https://bluedot.org/images/logo/icon-on-blue.png');
+  });
 });
--- a/apps/website/src/__tests__/pages/join-us/[slug].test.tsx
+++ b/apps/website/src/__tests__/pages/join-us/[slug].test.tsx
@@ -78,4 +78,16 @@ describe('JobPostingPage SSR/SEO', () => {
       expect(structuredData.hiringOrganization.name).toBe('BlueDot Impact');
     }
   });
+
+  test('uses fallback logo image when no `jobOgImage` provided', () => {
+    renderWithHead(
+      <JobPostingPage
+        slug="ai-safety-researcher"
+        job={mockJob}
+      />,
+    );
+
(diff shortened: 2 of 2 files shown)


marimeireles:
Because I'm building on your PR I found out a way of testing previews!
You can use the `ngrok` service to expose localhost to the web, copy the link from there and post it in the services: https://www.linkedin.com/post-inspector/ and https://developers.facebook.com/tools/debug/
Maybe that helps!

marn-in-prod:
> First prize would be updating the Airtable library to handle the 'attachment' type column, so that we could pull images directly from Airtable (maybe we can already do this but I wasn't sure how)

For handling attachment columns in Airtable, use `text().array()`. When there are no attachments, you'll get an empty array []. When there are attachments, you'll get an array of URLs that point directly to the files.

**Broader issue**: We actually need to use text().array() a lot more across our codebase. Many of our sync bugs (and the alerts for things not syncing correctly) happen because we're not using text().array() for lookup/linked fields. Since Airtable can return multiple values for these field types (depending on how they're configured), we need to handle them as arrays. I'll create an issue for soon, there's just a bunch of places where it's happening and we should address it.

**One caveat on attachments:** If loading attachment URLs counts against our airtable api quota, I'd rather not pull images directly from Airtable given how limited their QPS (queries per second) rate limits are. I haven't looked into whether this is the case, but something to be aware of.

PR #1680, opened 2025-11-24
Bugfix - courses with both certificate and active status do not show upcoming discussions

### LINKED ISSUE #1677: Fix conditions for the completed and in progress courses
### 🧩 Summary

In the user's [profile page](https://bluedot.org/settings/courses) the list of completed and in progress courses should be displayed based on these conditions.

I think the conditions are currently set so that if the user is issued a certificate AND the course status is still Active, then the course is not in the "In progress" section anymore.

### ✅ Definition of done

All of them require the user to have a course registration for that round.

In progress courses:

- Course registration, Decision = Accept
- Round status = Active
- Should still show whether or not there is a certificate, so long as the round status is Active

Completed courses:
- Certificate ID != null
- Doesn't matter if the course is active or not, it should show they have a certificate if they have a certificate.

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P1" => "High"}

joshestein:
# Description

We get `inProgressCourses` by filtering on `roundStatus = 'Active'`.

We get `completedCourses` by filtering on a valid certificate created at.

These are not mutually exclusive; it is possible for a registration to be both active and have an early certificate assigned, in which case it appears in both sections in a 'completed' state.

It should appear in the 'in progress' section with upcoming discussion info and in the 'completed' section with the certificate link.

## Issue

Fixes #1677

## Developer checklist

## Screenshot

| 📸 | Before | After |
|---------|---|---|
| 🖥️ | image | image |

### DIFF
--- a/apps/website/src/components/settings/CourseListRow.stories.tsx
+++ b/apps/website/src/components/settings/CourseListRow.stories.tsx
@@ -16,6 +16,11 @@ const meta: Meta<typeof CourseListRow> = {
   parameters: {
     layout: 'padded',
   },
+  args: {
+    isFirst: true,
+    isLast: false,
+    isCompleted: false,
+  },
 };
 
 export default meta;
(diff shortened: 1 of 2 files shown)


marn-in-prod (approved):
LGTM

PR #1693, opened 2025-11-25
[Feature] GroupSwitchModal: custom text for discussions that have passed

### LINKED ISSUE #1692: [Design] Improve past course labels in group switching
### 🎨 Summary

Image

When we display a group discussion in the group switching UI **that already has happened in the past** it should say "This discussion already has happened" instead of the "5 slots left". This way, users will immediately see the reason why they can't select the group slot.

### ✅ Definition of done

- [ ] adjust label with logic that detects past group discussion
- [ ] remove icon

### 📌 Notes / References

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

Will-Howard:
# Description

I went with "This discussion has passed" rather than "This discussion has already happened" because it fits better on small screens + sounds more natural imo.

## Issue

Fixes #1692

## Developer checklist

## Screenshot

| 📸 | Before | After |
|---------|---|---|
| 📱  |  | Screenshot 2025-11-25 at 22 06 57 |
| 🖥️ |  | Screenshot 2025-11-25 at 22 07 02 |

### DIFF
--- a/apps/website/src/components/courses/GroupSwitchModal.test.tsx
+++ b/apps/website/src/components/courses/GroupSwitchModal.test.tsx
@@ -432,9 +432,11 @@ describe('GroupSwitchModal', () => {
       expect(fullGroupOption).toHaveClass('opacity-50');
       expect(startedGroupOption).toHaveClass('opacity-50');
 
-      // Verify "No spots left" message appears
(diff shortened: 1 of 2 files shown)


Will-Howard:
Merging as this seems unobjectionable

Issue #1631, opened 2025-11-17
[Feature] Improve failing Lighthouse performance scores


joshestein:
### 🧩 Summary

<img width="781" height="705" alt="Image" src="https://github.com/user-attachments/assets/994b0690-cddb-4b2f-864e-928f6901a359" />

https://pagespeed.web.dev/analysis/https-bluedot-org/p77s4lswhb?form_factor=mobile

We have a LCP of 13.7s which is really bad. Some of the JS things are more difficult but I think some of the image optimisations are easy wins.

Honestly I think this should be P1 priority, see https://www.conductor.com/academy/page-speed-resources/faq/amazon-page-speed-study/ and https://www.thinkwithgoogle.com/_qs/documents/9757/Milliseconds_Make_Millions_report_hQYAbZJ.pdf

### ✅ Definition of done

There are many ways we can improve our performance.

Easiest TODOS are to use .webp for images, use smaller images (yes I'm looking at you 'large-photo.jpg' and 'hero.jpg'). For the gradient images I think we should just use CSS, consider overlaying on top of an SVG that generates noise.

We can also use `fetchpriority=high` for the hero image.

The events images also seem to be (overly) large.

### Suggested Priority Level

{"P2" => "Medium (Default)"}

PR #1694, opened 2025-11-25
[Feature] Optimise large homepage images

Will-Howard:
# Description

There were two large images on the front page which could cause performance issues on slow connections (and [hurts our Lighthouse score](https://github.com/bluedotimpact/bluedot/issues/1631)!). This PR replaces them with optimised webp versions.

For both I've checked that they look the same in context.

## Issue

#1631

## Developer checklist

N/A


Will-Howard:
Merging as I don't think there's any real risk here and I want to see the new score 😃

Issue #1698, opened 2025-11-26
[Bug] Remove duplicate resource completions


joshestein:
### 🧩 Summary

If a user does not have an entry in the 'resource completion' the very first time they complete a resource a new entry is created. However, during the in-flight time while this request is being made it is possible that they could uncomplete, complete again, etc. The end result is that a user can have multiple completion records in the 'resource completion' table for the same resource.

We should have an automation that runs to ensure that we only have one unique row for each resource completion. This is important because we want to consistently pull the same data to the frontend each time.

### 🛝 Steps to reproduce

1. Delete an existing resource completion for some resource you have (or find a course with a resource you don't have a completion for)
2. Reload page with said resource
3. Very quickly complete, uncomplete and complete the resource
4. Observe 3 records in the table
 

### ✅ Definition of done

There are no duplicate resource completions for the same resource.

If you want to find duplicate resource completions you can group the table by the `email | unit resource record id` field.

### Suggested Priority Level

{"P2" => "Medium (Default)"}

PR #1699, opened 2025-11-26
Re-use `getResourceCompletions`

### LINKED ISSUE #1661: [Feature] Re-use `getResourceCompletions`
### 🧩 Summary

In https://github.com/bluedotimpact/bluedot/pull/1659 we created a new procedure, `getResourceCompletions`. The unit layout currently iterates through each resource and calls `getResourceCompletion` (singular) for each resource. This is slow both from a network perspective and from a DB perspective.

Re-use the new procedure to fetch all resource completions in the parent, then pass individual completion data to `ResourceListItem`.

### ✅ Definition of done

1. `getResourceCompletion` (singular) is removed
2. `getResourceCompletions` (plural) is used to fetch all resource completions
3. Those completions are passed to `ResourceListItem`

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P3" => "Low"}

joshestein:
# Description

Since I optimised the procedure as part of the upstream PR it makes sense to re-use it. At first I thought this would only be for efficiency, but because of the bug I've described in https://github.com/bluedotimpact/bluedot/issues/1698 this is also so that always get the same data, regardless of whether we are loading completions for the sidebar or resource display.

1. Make `ResourceDisplay` re-use `getResourceCompletions`
2. Pass `resourceCompletion` as prop
3. `createMockResource` and `createMockResourceCompletion` for tests

## Issue

Fixes #1661

## Developer checklist

## Screenshot

NA

### DIFF
--- a/apps/website/src/server/routers/resources.ts
+++ b/apps/website/src/server/routers/resources.ts
@@ -1,5 +1,5 @@
 import {
-  and, eq, inArray, resourceCompletionTable,
+  and, desc, eq, inArray, resourceCompletionTable,
 } from '@bluedot/db';
 import { RESOURCE_FEEDBACK } from '@bluedot/db/src/schema';
 import { z } from 'zod';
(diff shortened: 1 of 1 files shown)


joshestein:
@marn-in-prod your duplicate records are a menace to society 😅 no I'm joking you are 100% right - I've added ordering by descending `autoNumberId` because I figure we probably always want the latest completion.

Issue #1610, opened 2025-11-12
[Design] Allow facilitators to update their discussion time


anglilian:
### 🎨 Summary

This functionality is important for facilitators especially before december when they will need to move the discussions around to accomodate for christmas and new year's!!

We can just stick to the old design in the loom / similar to what we have for group switching for now and work on getting a better design up later!

https://www.loom.com/share/85154ef499db44cab271d3be2aa9447e?sid=4ce2fcce-48d3-410a-aeea-0c4e98a2dee2

### ✅ Definition of done

Similar to group switching functionality, this creates a record in Airtable which triggers the appropriate automations.

**Create record**
Base: Course Runner
Table: Change discussion time

**Required fields**

- Status = Requested
- Facilitator = Course registration record of person submitting request
(EXCEPT for if switch type = Update discussion facilitator, then this field is the facilitator record for the person who will be taking over the session)
- Group = Group record of group being changed

If Switch type = "Change for one unit"

- Discussion = Group discussion record being changed
- Updated Datetime = new datetime

If Switch type = "Change permanently"

- Updated datetime = new datetime for all future sessions, including the next upcoming session

If Switch type = "Cancel one unit"

- Anything else = "Caroline cancelled group name unit "

(This field is a log of what was cancelled, otherwise we won't know what was deleted.)

The most straightforward way to access:
List of facilitators on the course: Course Runner base course registrations filter for facilitators on the round
List of groups from the facilitator: From the facilitator's course registration record -> Group (facilitator) field -> gives the list of group records assigned to the facilitator 
List of upcoming sessions for the facilitator: From the facilitator's course registration record -> Expected discussions (facilitator) field -> gives list of group discussion records

Note that the bracket is important because there are identical fields for the participant!

### 📌 Notes / References

I'm not wed to this particular design and think this could be massively improved 

### Suggested Priority Level

{"P2" => "Medium (Default)"}

joshestein:
Thank you for the clarification 😄 

1. Yes, they are two different tables 👍 If the functionality is 'kind of like the group switching one but for facilitators' I'm asking why the columns are different (e.g. if a facilitator can change discussion for one unit, do we not need to track 'old discussion' and 'new discussion' like we do for group switching).

2 + 3: sounds good 🚀 

4: makes sense

[...]

PR #1707, opened 2025-11-28
Schema updates for facilitator discussion switch table

joshestein:
# Description

Upcoming feature to allow facilitators to change / switch / cancel discussion times. This is just to add the schema that corresponds to this new table: https://airtable.com/appPs3sb9BrYZN69z/tblYpyG4RjhlMTmSm/

Tested by running pg-sync-service connected to local postgres and observing data syncing to me successfully.

## Issue

Part of #1610 

## Developer checklist

NA

## Screenshot

NA

### DIFF
--- a/libraries/db/src/schema.ts
+++ b/libraries/db/src/schema.ts
@@ -636,6 +636,10 @@ export const facilitatorDiscussionSwitchingTable = pgAirtable('facilitator_discu
       pgColumn: numeric({ mode: 'number' }),
       airtableId: 'fld3ITxndtB6ZZtN5',
     },
+    createdAt: {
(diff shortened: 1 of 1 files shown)


libraries/db/src/schema.ts
```diff
@@ -612,6 +612,49 @@ export const meetCourseTable = pgAirtable('meet_course', {
   },
 });
 
+export const facilitatorDiscussionSwitchingTable = pgAirtable('change_discussion_time', {
```
marn-in-prod: Nit: if it's only for facilitator's to change time, then why not call the PG table something closer to that? i.e. `change_discussion_time` -> `facilitator_discussion_switching` or something similar

PR #1720, opened 2025-11-30
[Bug] Use isAdmin to check admin access rather than adminUsersTable

Will-Howard:
# Description

As discussed in Slack, using `adminUsersTable` was insecure as we don't require users to verify their emails, allowing someone in principle to guess an admin email and sign up with admin powers. Previously this wasn't an issue as admin powers only grant you access to the sync dashboard. Now we are about to add [the ability to impersonate other users](https://github.com/bluedotimpact/bluedot/pull/1719), so need to tighten this up.

I have set `isAdmin` to true in Airtable on all the users that correspond to entries in the previous `adminUsersTable`.

## Developer checklist

### DIFF
--- a/apps/website/src/server/trpc.ts
+++ b/apps/website/src/server/trpc.ts
@@ -1,6 +1,5 @@
 import {
-  AirtableTsError, ErrorType,
-  userTable,
+  AirtableTsError, ErrorType, userTable,
 } from '@bluedot/db';
 import { requestCounter } from '@bluedot/ui/src/utils/makeMakeApiRoute';
 import { initTRPC, TRPCError } from '@trpc/server';


libraries/db/src/schema.ts
```diff
@@ -42,10 +42,7 @@ export const syncMetadataTable = pgTable('sync_metadata', {
 });
 
 /**
- * Admin users table for pg-sync dashboard access control.
- * NOTE: This is a regular pgTable, NOT synced from Airtable.
- * Used solely for internal admin authentication. Currently we add
- * emails manually to this table.
+ * @deprecated Use `isAdmin` field on userTable instead.
```
Will-Howard: I'll delete this fully in a later PR (once this code is deployed)

Issue #1595, opened 2025-11-09
[Feature] Group switch modal v2


Will-Howard:
### 🧩 Summary

Implement "Switching Modal" in these designs:
- Updated UI: [Here](https://www.figma.com/design/m1sdGUyCu0OHY3GEeEVFYa/1.3----Group-Switching-v20?node-id=1-76710&t=rjKAM5hUQg0mmOBb-1)
- Updated prototype: [Here](https://www.figma.com/proto/m1sdGUyCu0OHY3GEeEVFYa/1.3----Group-Switching-v20?page-id=1%3A76710&node-id=1-79862&viewport=124%2C-818%2C0.35&t=nytBSJqZ4t33i3yz-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=1%3A79862&show-proto-sidebar=1)

### ✅ Definition of done

Key changes in `GroupSwitchModal`:
- [ ] "Switch for one unit"/"Switch permanently" is moved to the header
- [ ] Non-manual requests require 3 screens: "reason for change" -> "select group/discussion" -> "success"
- [ ] The "update availability" button in the manual request screen is replace with a link + checkbox

### Suggested Priority Level

{"P2" => "Medium (Default)"}

PR #1732, opened 2025-12-01
`GroupSwitchModal` extract custom `Select`

joshestein:
# Description

New facilitator switching modal is going to re-use the same `Select`.

I did a brief search and it seems like there are no downstream consumers of the `Select` exported from `@bluedot/ui`. I suggest that we replace that select with this one.

## Issue

Part of #1610

## Developer checklist

## Screenshot

NA

### DIFF
--- a/apps/website/src/components/courses/group-switching/Select.tsx
+++ b/apps/website/src/components/courses/group-switching/Select.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import { useState } from 'react';
+import { useState, type Key } from 'react';
 import {
   Button,
   Label,
@@ -27,7 +27,7 @@ const Select = ({
   const [isOpen, setIsOpen] = useState(false);
   const selectedOption = options.find((op) => op.value === value);
 
-  const handleSelectionChange = (key: React.Key | null) => {
+  const handleSelectionChange = (key: Key | null) => {
     if (key !== null) {
       onChange?.(key as string);
       setIsOpen(false);


joshestein:
By 'other Select' do you mean the one in `@bluedot/ui`? If you agree we should remove it I'd rather just replace it with this extracted component.

marn-in-prod (approved):
Looks good. ~~I'm also fine replacing the select given that the other one isn't in use~~

marn-in-prod:
I feel like this Select deviates quite a bit from what a normal select does.  it pushes content down instead of popping over it, which is unusual.

 Because of that, I think it makes sense to keep it separate from the UI library rather than replacing the existing one. I'd hold off on deleting @bluedot/ui Select too, removing it makes it harder to reuse later, and I don't think people are going to check git history to see if there was ever a Select before building a new one.

I wonder if this group dicussion select should be renamed so that it's clear that it deviates a bit from how normal selects work as well, or at the very least, documentation added that this is how it works.

Will-Howard:
> Because of that, I think it makes sense to keep it separate from the UI library rather than replacing the existing one. I'd hold off on deleting @bluedot/ui Select too, removing it makes it harder to reuse later, and I don't think people are going to check git history to see if there was ever a Select before building a new one.
>
> I wonder if this group dicussion select should be renamed so that it's clear that it deviates a bit from how normal selects work as well, or at the very least, documentation added that this is how it works.

This is a case where this applies:

> FYI in case you're not aware: There are [new designs for the group switch modal](https://github.com/bluedotimpact/bluedot/issues/1595), which will supersede this Select component.
>
> This is mainly a flag to not spend time optimising this as it will be replaced soon. I will try to update the facilitator modal as well when I finish implementing those designs (or flag if this is too big/doesn't seem worthwhile).

The new one is much more standard-looking and I think we can adopt it generally:

Screenshot 2025-12-03 at 06 44 30

And this should be added by EOW

joshestein:
To err on the side of caution I won't delete the old select here.

@Will-Howard I'll merge this then you (or I) can update the styling in the extracted component, then maybe another PR down the line to replace the existing Select with the updated component?

PR #1747, opened 2025-12-03
Facilitator switching: rename schema field

joshestein:
# Description

I misunderstood this field to be the time the record was updated. It is the new time that discussions are update to. We have automations setup that inspect this field and update the corresponding discussions.

## Issue

Part of #1610 

## Developer checklist

## Screenshot

NA


marn-in-prod (approved):
LGTM. The field isn't already being used so I think we can just directly rename it

PR #1749, opened 2025-12-03
Time picker

joshestein:
# Description

In https://github.com/bluedotimpact/bluedot/issues/1610 we need to allow users to pick a custom day and time. We don't have components to allow this.

Used this as reference: https://react-spectrum.adobe.com/react-aria/TimeField.html

@herrhaase this definitely needs your input. I'm creating this PR as something quick and dirty, I don't want to be blocked by not having UI components. The design can always be updated later.

## Issue

Part of https://github.com/bluedotimpact/bluedot/issues/1610

## Developer checklist

## Screenshot

image

### DIFF
--- a/libraries/ui/src/TimePicker.stories.tsx
+++ b/libraries/ui/src/TimePicker.stories.tsx
@@ -21,7 +21,7 @@ export const Default: Story = {
 
       return (
         <div className="flex flex-col gap-4">
-          <TimePicker label="Pick a time" value={time} onChange={setTime} />
+          <TimePicker label="Pick a time" timeValue={time} onTimeChange={setTime} />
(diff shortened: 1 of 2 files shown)


marn-in-prod (approved):
The Daypicker version of this is a lot better IMO because it does allow people to click an icon and get a dropdown menu. Similar to the other PR, would prefer to just use that instead of react-aria-component.

https://daypicker.dev/guides/timepicker

marn-in-prod:
@joshestein Ah, I didn't realize that. I thought react-day-picker had a dedicated time picker component as well since I just saw the UX. Let's stick with aria here since it'll have consistent browser styling over the native input.

marn-in-prod (approved):
See earlier comment but LGTM to keep react-aria-component

PR #1754, opened 2025-12-04
Return `null` on successful group switching mutation

joshestein:
# Description

We are getting a slack alert every time someone switches groups. This is because the mutation implicitly returns `undefined` which is then converted to null and alerted due to changes introduced in https://github.com/bluedotimpact/bluedot/pull/1708.

Prevent this by returning `null` on success. idk maybe we should rather return `{ 'success': true }` or something explicit 

## Issue

NA

## Developer checklist

## Screenshot

NA


marn-in-prod (approved):
LGTM.

Returning null when the frontend doesn't consume the result is better than returning the inserted row automatically. It's also better than { success: true } or similar. you can already assume success by the fact that it didn't error out.

PR #1818, opened 2025-12-14
Group discussions: unify timing logic, rename hasStarted to isTooLateToSwitchTo

### LINKED ISSUE #1690: [Feature] Unify group discussion logic: When they are counted as "soon", allowed actions, etc
### 🧩 Summary

There are various places where we handle group discussions differently depending on the relative start/end time:
- On the backend, in `apps/website/src/server/routers/group-discussions.ts`, to decide which discussion to return to display in `GroupDiscussionBanner`
- In `GroupDiscussionBanner`, to decide which actions to show to the user, and whether to display it as "live"
- In `CourseDetails` and `CourseListRow` on the course settings page, also to decide which actions to show to the user, and whether to display it as "live"

We currently define the rules for this separately in each file, and try to make them overall consistent. We should instead have one source of truth for when a discussion is counted as soon as it's live or ended, and the actions that are allowed depending on this.

I suggest doing something like this:
- In `website/lib` add a subfolder `group-discussions`, in this add two files:
  - utils.ts: constants and helper function relating to when a discussion is counted as soon/live/ended
  - useGroupDiscussion.tsx: (or hooks.tsx) which exports a `useGroupDiscussion` hook, the primary job of which is to return the list of allowed actions as a `IButtonOrMenuItem[]` array. As of #1688 this is duplicated between `CourseDetails` and `GroupDiscussionBanner`

### ✅ Definition of done

The time based rules for groups discussions are refactored into a single source of truth (there is no code outside of that which decides behaviour for a group discussion based on an explicit time)

### 📌 Notes / Designs

_No response_

### Suggested Priority Level

{"P2" => "Medium (Default)"}

Will-Howard:
# Description

On the timing logic: In #1690 I suggested also adding a hook `useGroupDiscussion` to determine the "actions" that are available for a given group discussion, as there is code in `GroupDiscussionBanner` and `CourseDetails` which uses similar rules for which actions to show. Upon trying to implement this I decided it's not worth it as the two cases aren't similar enough for this to really reduce complexity overall.

## Issue

Fixes #1690

## Developer checklist

### DIFF
--- a/apps/website/src/server/routers/group-switching.ts
+++ b/apps/website/src/server/routers/group-switching.ts
@@ -100,7 +100,7 @@ export function calculateGroupAvailability({
       // Update existing group data
       const existing = groupData[groupId];
 
-      // It's too late to switch into the group if it's too late to switch into *any* of the discussions
+      // It's too late to switch into the group if it's too late to switch into *all* of the discussions
       existing.isTooLateToSwitchTo = existing.isTooLateToSwitchTo && isTooLateToSwitchTo;
 
       if (!isTooLateToSwitchTo) {


apps/website/src/components/courses/GroupSwitchModal.stories.tsx
```diff
@@ -43,7 +43,7 @@ const mockAvailableGroupsAndDiscussions: DiscussionsAvailable = {
       }),
       userIsParticipant: true,
       spotsLeftIfKnown: 0,
-      allDiscussionsHaveStarted: false,
+      isTooLateToSwitchTo: false,
```
Will-Howard: Unrelated: I renamed `hasStarted` and `allDiscussionsHaveStarted` to `isTooLateToSwitchTo` for clarity

Issue #1821, opened 2025-12-16
Optimise the size of all images on the site


Will-Howard:
### 🧩 Summary

The main hero images are now optimised but we still have a lot of random images on the site that are very large. We should optimise all of these.

### ✅ Definition of done

For all images that are displayed on the site:
- Use imagemagick to convert them to webp and compress to an appropriate baseline level
- Consider also using next/image in code on a case by case basis. This requires setting a fixed width + height and is only really useful if the image is displayed at significantly different sizes
- Ensure fetchpriority=high is used for images that block Largest Contentful Paint

Stretch: Consider also adding some kind of CI step that asserts that images can't be too large. Though this may not be worthwhile if there isn't a simple way to do it.

Will-Howard:
Update on the image optimisation work and this lint rule PR (#1866):

I focused on optimising images first because it looked like the lowest-hanging fruit. The images were obviously oversized (some pages required >1MB of images before the main content appeared), and pagespeed.web.dev flagged it as the top priority recommendation.

I have now optimised all images across the site (#1864, #1868) to be under 200kB. From looking at the before and after in posthog (roughly speaking [this graph](https://eu.posthog.com/project/21762/web/web-vitals?date_from=2025-12-27T00%3A00%3A00&date_to=2025-12-31T23%3A59%3A59&interval=day&path_cleaning=true&filter_test_accounts=false&compare_filter=%7B%22compare%22%3Atrue%7D&percentile=p75&tile_visualizations=%7B%7D&filters=%5Bobject%20Object%5D%20) before and after 2025-12-29), I don't think this has made a critical difference. A difference of ~100ms it would take some time to show up, but we can rule out a difference of >500ms. 

I've also done more investigation of this geographical difference in latency (#1873), and I'm now convinced that _this_ is the biggest issue currently. Our server is in Amsterdam, and each page load requires ~2 round trips (one for the HTML, one for CSS/JS/assets). For users in the US or Asia, this alone can explain the ~1.25s difference we see between European and non-European users.

I put together this lint rule (#1866) as a way to maintain the image optimisation going forward. Given that I now don't think images were the critical bottleneck, I'm leaning towards closing that PR, and just relying on optimising images on an ad-hoc basis (until it actually becomes a bottleneck). I'll add a note to the README about image optimisation best practices instead.

The one-off optimisation will have done some good, but probably more like a few hundred ms saved on slow connections.

cc @marn-in-prod

PR #1859, opened 2025-12-27
Delete unused images

Will-Howard:
# Description

I'm working on optimising the size of all the images used on the site (#1821). For this PR I just used a script to go through and delete unused ones, to have a better starting point to work on optimising the others.

You can see the script I used in [this commit](https://github.com/bluedotimpact/bluedot/pull/1859/changes/4d89f197f34b2b860a28f7262548ea2c4d1a11f8). I deleted this in the next commit so it won't be merged to the codebase but will be here for reference if needed again in the future.

You can run the script like so:
```
cd apps/website
npx tsx scripts/find-unused-images.ts
```

## Issue

#1821

## Developer checklist


libraries/ui/src/CourseCard.stories.tsx
```diff
@@ -1,39 +0,0 @@
...
-
-export default meta;
-type Story = StoryObj<typeof meta>;
-
-export const Default: Story = {
-  args: {
-    title: 'Alignment',
-    description: 'AI systems are rapidly becoming more capable and more general. Despite AI’s potential to radically improve human society, there are still open questions about how we build AI systems that are controllable, aligned with our intentions and interpretable.',
-    courseType: 'In-depth course',
-    imageSrc: '/images/intro-course.png',
```
Will-Howard: This image was only used in this story, but the component itself is no longer used, I deleted it along with the image

PR #1865, opened 2025-12-28
[Tech debt] Delete deprecated images

### LINKED ISSUE #1821: Optimise the size of all images on the site

Will-Howard:
# Description

Delete images that are no longer used as a result of the optimisation done for #1821. The reason I'm not merging this immediately is that stale clients may still have references to the old images, for potentially up to a few days.

I will merge this about a week after I've finished everything in #1821.

## Issue

Fixes #1821


Will-Howard:
The optimised images has been out for two days now. I think it's safe to merge this now wrt breaking old clients, and better to do sooner rather than later to reduce the risk of conflicts arising. I'm just raising this for review to see if @marn-in-prod has any concerns, as it is somewhat risky to delete ~100 images.

PR #1964, opened 2026-01-15
PG Sync: don't throw error when delete doesn't return results

joshestein:
# Description

PG sync is currently not working. It's failing because there are records it is trying to delete (`{isDelete: true}`) that do not exist. I believe the generated SQL is something like:

```sql
DELETE FROM exercise_response WHERE id = '<>' RETURNING *;
```

If this returns empty we currently throw an error:

https://github.com/bluedotimpact/bluedot/blob/086727c3eb23723c253859265c33d8e6fa743c3e/libraries/db/src/lib/client.ts#L446

However, empty does not mean an error; it could mean that there is simply no record. Some ideas: the record may never be syncing to PG in the first place, the record was already deleted, there are stale notifications from previous sync.

I think that if `tx.delete()` fails it will throw, whereas returning an empty result is not an error in itself.

We can allow idempotent deletes by returning `undefined` when no delete results are found.

Tested by running locally and observing sync of 'exercise_response' table without errors.

## Issue

Fixes #

## Developer checklist

## Screenshot

NA

### DIFF
--- a/libraries/db/src/lib/client.ts
+++ b/libraries/db/src/lib/client.ts
@@ -381,7 +381,9 @@ export class PgAirtableDb {
   ): Promise<BasePgTableType<TTableName, TColumnsMap>['$inferSelect']> {
     const fullData = await this.airtableClient.insert(table.airtable, data);
 
+    // `ensureReplicated` only returns undefined for idempotent deletes; upserts always return a result
     const pgResult = await this.ensureReplicated({ table, id: fullData.id, fullData });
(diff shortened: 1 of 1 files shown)


Will-Howard (approved):
I haven't tested this locally, but I agree making deletes idempotent is the right way to go so this looks good 👍 .

I agree with the AIs that it would be better to explicitly allow undefined in the return type. If this is time-consuming to change, then it could be better to punt on this initially to fix the acute problem. I think it's worth at least trying to do this now though because if it results in downstream type errors that could be a sign that the result is used in a way that assumes it isn't `undefined`

PR #1966, opened 2026-01-16
Fine-tune facilitator DB filter

### LINKED ISSUE #1965: Facilitator cannot update discussion facilitator
### 🧩 Summary

I received this issue from one of the facilitators on the Technical AI Safety course

Image

### 🛝 Steps to reproduce

Perhaps try to do the change discussion facilitator and see if you encounter this.

### ✅ Definition of done

Facilitators can update the discussion facilitator.

### 📌 Notes / Designs

_No response_

joshestein:
# Description

To find a facilitator we:

1. Use the course slug to find the course,
2. Find the corresponding course registration
3. Use the course registration ID and facilitator email to lookup the facilitator in the `meetTable`

The problem is that there can be multiple similar records for a facilitator in `meetTable` where the course ID is the same. For example, if a facilitator application was previously denied and now accepted, or a facilitator facilitates multiple rounds, some of which are active, some are not. This means that `getFirst` might return an unexpected record.

I've improved the DB filter to check for 'decision: "Accept"' and 'roundStatus: "Active"'. There are 8 facilitators who still have two records returned when we search for them in the DB.

## Issue

Fixes #1965

## Developer checklist

## Screenshot

NA

### DIFF
--- a/apps/website/src/server/routers/facilitator-switching.ts
+++ b/apps/website/src/server/routers/facilitator-switching.ts
@@ -28,6 +28,8 @@ const getFacilitator = async (courseSlug: string, facilitatorEmail: string) => {
     filter: {
       email: facilitatorEmail,
       courseId: course.id,
+      roundStatus: 'Active',
+      decision: 'Accept',
     },
   });
   if (!courseRegistration) {


joshestein:
I'm going to merge this to unblock a bunch of facilitators who this would help; as noted in the upstream issue, it might make more sense to find facilitators using round ID. I will do some investigating and open a separate PR if that is possible.

PR #1983, opened 2026-01-20
[Feature] Add process for deprecating and deleting fields 

### LINKED ISSUE #1926: Add process for deprecating and deleting database fields
### 🧩 Summary

There's currently no safe way to delete a database field. This has caused issues recently, as I (and apparently Josh) was under the impression that it _was_ safe to delete a field once you've fully stopped using it in code:
- #1862 (images field on courseTable)
- #1858 (courseCadence field on courseTable)

The reason it is not safe is that when you remove a field from the schema and merge into master:
- pg-sync-service immediately drops the column when the PR is merged
- The production server doesn't deploy until you do a release, so in the window between the PR being merged and the release, it tries to select the column and this results in errors
  - The point is that even if you're not using the column in code, when the server tried to select a whole row from the table in Drizzle, it includes the column in the SELECT statement
- Less critical: other developers will also be affected by the same problem with their own copies of the code until they pull from master

I propose we add a `deprecated: true` flag to field definitions. When set:
- pg-sync-service sees the field and does NOT drop the column
- Drizzle excludes the field from SELECT statements (so queries don't break)
- TypeScript types exclude the field (so you get compile errors if you try to use it, and it feels deprecated to devs)

This lets you remove a field in two steps:
1. PR 1: Add `deprecated: true` to the field. Merge and release. The column remains, but code stops selecting it.
2. PR 2 (a few days later, once all running code has adopted the first PR): Delete the field definition entirely. Now pg-sync drops the column, which is safe because no running code references it.

Where to start:
- `libraries/db/src/lib/typeUtils.ts`: Add `deprecated?: true` to `PgAirtableColumnInput`.
- `libraries/db/src/lib/db-core.ts`: When building `this.pg` in the constructor, skip deprecated columns entirely so Drizzle never sees them.
- `apps/pg-sync-service/src/lib/schema-sync.ts`: In `cleanupRemovedColumns`, include deprecated columns in the "expected" set so they don't get dropped.

### ✅ Definition of done

When the `deprecated: true` flag is set on a field in `schema.ts`:
- pg-sync-service sees the field and does NOT drop the column
- Drizzle excludes the field from SELECT statements (so queries don't break)
- TypeScript types exclude the field (so you get compile errors if you try to use it, and it feels deprecated to devs)

### 📌 Notes / Designs

_No response_

Will-Howard:
# Description

Adds a `deprecatedColumns` option to `pgAirtable()` for safely phasing out database columns.

Summary of problem from #1926: When removing a column from the schema, pg-sync-service drops the column immediately on merge, but the website doesn't deploy until release. In this window, the old website code tries to SELECT the dropped column and errors.

With this change, when a column is added to `deprecatedColumns` it will stop being SELECT-ed but won't be dropped from the db. This lets us safely remove columns in two deploys (see updated DEVELOPMENT_HANDBOOK.md).

I also deprecated the image column on `courseTable` as an example (we already don't use it):
```typescript
  deprecatedColumns: {
    image: {
      pgColumn: text(),
      airtableId: 'fldh90A6x8HwQSkMy',
      deprecated: true,
    },
  },
```
I suggest we leave this here rather than deleting it in the second deploy so that devs have a reference for how to do this. It is messy to leave deprecatedColumns there for no reason but it doesn't actually do any harm.

### Why this approach differs from the issue description

The issue suggested adding `deprecated: true` inline within `columns`. I tried this but ran into two problems:

1. TypeScript complexity: Filtering deprecated fields from the same object while preserving type inference was messy. Since we care less about deprecated fields anyway, moving them to a separate `deprecatedColumns` object sidesteps this.

2. Drizzle interactive prompt bug: I wanted to keep the deprecatedColumns as just a list of strings, and filter these out from `DROP COLUMN` statements that `pushSchema` generates. However, there's a [known Drizzle bug](https://github.com/drizzle-team/drizzle-orm/issues/4651) where `pushSchema` prompts for interactive input when columns are added and removed simultaneously, which hangs in CI. I couldn't find a way to work around this without creating a fully-fledged `pgWithDeprecatedColumns` version of each table, so I ended up doing this.

## Issue

Fixes #1926

## Developer checklist

## Screenshot

N/A, backend only

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,6 +14,8 @@ Fixes #
 - [ ] Considered having snapshot tests and/or happy path functional tests
 - [ ] Considered adding Storybook stories
 
+<!-- If adding/removing db schema, see DEVELOPMENT_HANDBOOK.md for the two-PR migration process -->
+
 <!-- You might also want to check the tests locally with `npm run test`, although CI will check this for you -->
 
 ## Screenshot


Will-Howard:
I'll wait for review on this one so everyone sees the new process. Lmk if you have any questions, I can update the DEVELOPMENT_HANDBOOK if things are unclear

Will-Howard:
> I think we should add a comment to .github/pull_request_template.md about the two-PR process for schema changes.

Yep sounds good! I will add that and merge this now

PR #1992, opened 2026-01-23
Sidebar acts as accordion, clicking header expands unit details

### LINKED ISSUE #1950: Side panel doesn't expand unit on mobile
### 🧩 Summary

On mobile and web, when you tap the title of the unit or the drop down arrow, it will open the page instead of just expanding the section as expected.

Image

The impact of this is that it's harder to browse through the curriculum from titles.

### 🛝 Steps to reproduce

1. Go to https://bluedot.org/courses/agi-strategy/4/1 on mobile

### ✅ Definition of done

When tapping the title or arrow on mobile/desktop, the side panel expands the options instead of bringing you to the page.

The red area should only expand/collapse rather than navigate.

Image

### 📌 Notes / Designs

_No response_

joshestein:
# Description

We want clicking unit headers to expand the unit section rather than navigating to the clicked unit. To facilitate we make the following changes:

1. When a unit layout loads we fetch the unit with all basic chunk information for quick display (i.e. with all chunk titles). This allows for fast expansion/collapse of units
2. When a unit is expanded, we lazily fetch core completion data for that unit
3. Only when a chunk within a unit is clicked do we navigate to that chunk

Minor improvements:

1. Filtering core resources on the server, not the client
2. Allow entire mobile course unit header to be clickable, not just the arrow

## Issue

Fixes #1950

## Developer checklist

## Screenshot

| 📸 | Before | After |
|---------|---|---|
| 📱  | mobile_course_accordion_before | mobile_course_accordion |
| 🖥️ | desktop_course_accordion_before | desktop_course_accordion |

### DIFF
--- a/apps/website/src/components/icons/ChunkIcon.tsx
+++ b/apps/website/src/components/icons/ChunkIcon.tsx
@@ -29,7 +29,7 @@ export const ChunkIcon: React.FC<ChunkIconProps> = ({
   }
 
   return (
-    <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <svg width={size} height={size} viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
       <path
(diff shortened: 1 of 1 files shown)


marn-in-prod (approved):
Looks good! Appreciate pulling others in to share knowledge on this given the scope of the changes. Tested locally and the accordion works well. 

I also agree that the AI comments dont matter too much here IMO

Issue #1794, opened 2025-12-11
[Design] Facilitators can view their participants' exercise responses


anglilian:
### 🎨 Summary

This is the old "design". Here's a [Loom](https://www.loom.com/share/27af5e11cbed47528a7c772490e49449) of what the features + UI was

<img width="1456" height="1308" alt="Image" src="https://github.com/user-attachments/assets/29f0773c-e381-4917-902f-39fe05253bca" />

TBH I think this looks incredibly ugly and would be keen for a much better design implementation. 

### ✅ Definition of done

- Allow facilitators to review their group's exercise responses before the session
- Making the responses visible by default led to slow page loading times on the old course hub, if this isn't a problem we can make it show by default if not, toggling is usually useful
- I don't allow facilitators to submit responses, mostly to avoid making this view feel very cluttered

### Suggested Priority Level

{"P2" => "Medium (Default)"}

anglilian:
@marn-in-prod could you assign someone to this? Or other engineers feel free to take it on 💪

marn-in-prod:
Toggle was meant to switch between the two different UX's, you see one OR the other.

So we're not supposed to "stack" and "combine" both of the modes.

This also means that the engineer should probably change the text on the toggle to accurately reflect this.

anglilian:
Looks great!! very happy with this :)

PR #2001, opened 2026-01-31
[Feature] Add ability for facilitators to view participant responses

Will-Howard:
# Description

Facilitators see their group's exercise responses on the course unit page, with a toggle to switch back to their own editable response.

Things to note:
- Title, description, and the completion checkbox were previously duplicated across FreeTextResponse and MultipleChoice. I moved them up into Exercise.tsx so the facilitator view could be added as a third render path without more duplication.
- I removed the "Correct! Quiz completed" banner from MultipleChoice. The green border on the correct option already communicates success, and the banner used z-index/negative-margin hacks and added a surprising amount of complexity (fyi @anglilian)
- Only completed responses are shown to facilitators. The designs show a timestamp but we don't currently track completion time so I omitted that in this PR. I created #2002 as a followup to implement this
- I changed `<ul>`/`<li>` to `<div>` in ResourceDisplay.tsx. This added coupling between components (because things that could appear in ResourceDisplay technically needed to be `<li>`s). This already wasn't consistently applied and I'd be surprised if screen readers interpreted it correctly anyway (because each `<li>` is a complex form) so I decided to cut it
- There were pre-existing import cycles between Exercise/GroupResponses/MarkdownExtendedRenderer. This PR adds some more. The underlying problem is deeper than is worth addressing for this feature IMO

## Issue

#1794

## Developer checklist

## Screenshot

| 📸 | Before | After |
|---------|---|---|
| 📱  | Screenshot 2026-01-31 at 23 10 16 | Screenshot 2026-01-31 at 23 11 17 |
| | | Screenshot 2026-01-31 at 23 11 07 |
| | | Multiple groups available, from storybook: Screenshot 2026-01-31 at 23 11 59 |
| 🖥️ | Screenshot 2026-01-31 at 23 09 00 | Screenshot 2026-01-31 at 23 06 20 |
| | | Screenshot 2026-01-31 at 23 06 47 |
| | | Multiple groups available, from storybook: Screenshot 2026-01-31 at 23 07 47 |

### DIFF
--- a/apps/website/src/components/courses/exercises/Exercise.test.tsx
+++ b/apps/website/src/components/courses/exercises/Exercise.test.tsx
@@ -1,5 +1,6 @@
 import '@testing-library/jest-dom';
-import { render, screen } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import {
   afterEach, beforeEach, describe, expect, test, vi,
 } from 'vitest';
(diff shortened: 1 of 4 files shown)


marn-in-prod (commented):
Functionally, this is good enough to submit (I'll take another look at the code shortly and send comments).

One thing to highlight, the "show more" toggle appears to be based entirely on word count rather than also considering the number of newlines, or capping at a max height before showing the button (which I think might be the best solution). 

Def not a blocker to submit though

Image

marn-in-prod (approved):
LGTM. Nothing blocking. See also my earlier comment with the image

PR #2024, opened 2026-02-04
[fix] Remove partner details from AGI strategy content, and fix tests

dewierwan:
# Description

## Screenshot

Before and After
CleanShot 2026-02-04 at 17 25 10@2x


dewierwan:
@Will-Howard still keen to remove the partner details, though you can ignore the test updates! Could you please own removing the "this was built with XYZ" from the agi strategy course landing page? Thank you!

PR #2034, opened 2026-02-06
Fix deprecated `completed` in favour of `completedAt`

joshestein:
# Description

`completed` was deprecated in favour of `completedAt`. There are still some usages of `completed` that were not updated, breaking CD.

## Issue

NA

## Developer checklist

### DIFF
--- a/apps/website/src/components/courses/courseSidebarStoryFixtures.ts
+++ b/apps/website/src/components/courses/courseSidebarStoryFixtures.ts
@@ -95,7 +95,7 @@ export const someProgressHandlers: RequestHandler[] = [
     makeResourceCompletion('rc-2', 'res-2', 2),
   ]),
   trpcStorybookMsw.exercises.getExerciseCompletions.query(() => [
-    { exerciseId: 'ex-1', completed: true },
+    { exerciseId: 'ex-1', completedAt: new Date().toISOString() },
   ]),
 ];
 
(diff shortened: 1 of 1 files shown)


joshestein:
@Will-Howard just ran this on prod:

```sql
  SELECT count(*)
  FROM "exercise_response"                                                                                  
  WHERE "completed" = true AND "completedAt" IS NULL;
```

and see a count of 1024. Is this any cause for concern?

Also, I see we haven't gotten any notifications about pg-sync since Thurs, when I think the original changes were merged. Also wondering if that's expected as a result of these changes not yet been merged, or if this could be related to something else.

Will-Howard:
Merging this now, overall what happened here was:
  - I merged #2007 and ran a backfill script to convert `completed` to `completedAt` in Airtable
  - There was a semantic conflict with #2026 (Track progress in sidebar), which merged between the schema change and code switch, causing CD to fail. Not noticing this was my mistake
  - The old code using completed has been running on prod since then
  - ~1000 new exercises were marked completed in that time, creating the `completed = true AND completedAt IS NULL` cases Josh found (the original ~65k were already backfilled)
  - This shouldn't have affected users, they were just getting the old code path that still reads `completed`

So overall to get us into the state we want:
- [x] Run the backfill script again to cover the new cases
- [ ] Merge and deploy this PR
- [ ] Run backfill one more time to catch any that slip through during the transition

Issue #120, opened 2025-01-29
Update `eslint` to `v9`


janraasch:
> ESLint v8 is end-of-life as of October 5, 2024. 

See https://turbo.build/repo/docs/guides/tools/eslint

domdomegg:
I've updated eslint-config-domdomegg, which is the core of the main config used across the repo, to support eslint 9 - so this should be a fairly easy and quick transition.

It took about 30 minutes to update the similarly structured and probably higher complexity raise repository in [this PR](https://github.com/raisenational/raise/pull/404). NB: because many other eslint packages still don't support eslint 9 (e.g. the airbnb preset), some other underlying presets changed which changed an indenting rule which resulted in a lot of changes - almost almost all auto migrated with `npm run lint:fix`

anglilian:
@Will-Howard is the decision to not work on this or for it to be deprioritised? Also wondering if it's still relevant. If not, can we close it?

domdomegg:
fwiw I think this won't be _too_ difficult, the changes probably look like the config+package.json in https://github.com/domdomegg/aws-ses-v2-local/commit/9add1d7a5b5af59c1deb2c673818c3a64961bc99 then running `npm run lint:fix` everywhere

I think the reason we didn't do this before was 1. not having a very clear upgrade path like the above, and 2. having many people devving and a big 'change all the formatting everywhere' change would create a tonne of merge conflicts.

PR #2042, opened 2026-02-09
Upgrade eslint to v9: Create v9 config and migrate `apps/availability`

Will-Howard:
# Description

eslint v8 reached end-of-life in October 2024. This PR begins migrating the whole to eslint 9 with the new "flat config" format (`eslint.config.mjs`), starting with `apps/availability` as a straightforward case. After that my plan is to:
1. Migrate the rest of the codebase apart from `apps/website` (PR 2)
2. Migrate `apps/website`, and delete the old config (PR 3)

This PR introduces a new shared config package (`@bluedot/eslint-config-v9`) that lives alongside the existing `@bluedot/eslint-config`. Individual workspaces can be migrated one at a time by switching their dependency and adding an `eslint.config.mjs`.

### New shared config: what changed and why

The previous config dependency chain was `@bluedot/eslint-config` -> `eslint-config-domdomegg` v1 -> `airbnb-typescript`. domdomegg released v2 for eslint v9, but v2 switches the base from `airbnb-typescript` to `eslint-config-xo-typescript`. This introduced a large number of new style defaults that didn't match our existing conventions, e.g preferring tabs to spaces. I found myself overriding enough rules that it made more sense to vendor domdomegg's ~130 lines into our own config and control everything directly.

I actually first migrated `apps/website`, which is where I ran into all of these problems. I decided to put up this simpler case first just to get some review and split it into more manageable chunks. Because I've already done `apps/website`, I'm reasonably confident this is all of the surprises we'll run into. Also: the commit history here includes migrating `apps/website` and then undoing this. This will all get squashed when this is merged so nbd. 

### Style overrides (XO defaults we override to preserve existing conventions)

- `@stylistic/indent: 2`: XO defaults to tabs (~31k errors if we switched)
- `@stylistic/object-curly-spacing: 'always'`: XO defaults to no spaces (~5k errors)
- `@stylistic/jsx-quotes: 'prefer-double'`: XO defaults to single (~2.8k errors)
- `@stylistic/block-spacing: 'always'`: must match object-curly-spacing to avoid a circular auto-fix loop where the two rules fight each other
- `@stylistic/indent-binary-ops: 2`: must match indent

Smaller XO style changes (curly braces on single-line blocks, function-paren-newline, etc.) I accepted, these produce auto-fixable diffs but the XO defaults are reasonable.

### Issues encountered and workarounds

- `eslint-plugin-tailwindcss`: Not compatible with ESLint 9 + Tailwind v4. This is the biggest loss, but unfortunately there is no way to get it to work. This set these rules: TODO list. There is an alternative [eslint-plugin-better-tailwindcss](https://github.com/schoero/eslint-plugin-better-tailwindcss) that we could try
- @typescript-eslint/promise-function-async: Disabled. React 18+ component return types include Promise, so this rule incorrectly adds async to components, breaking builds.
- no-implicit-coercion: Disabled. Auto-fixes !!x -> Boolean(x), but Boolean() doesn't narrow types in TypeScript.
- non-nullable-type-assertion-style: Kept enabled. Auto-fixes as HTMLElement -> ! which widens the type; workaround is querySelector<HTMLElement>(...)!.
- Circular block-spacing / object-curly-spacing auto-fix: Set both to 'always' so they don't fight each other.

## Issue

#120

## Developer checklist

### DIFF
--- a/libraries/eslint-config-v9/src/index.mjs
+++ b/libraries/eslint-config-v9/src/index.mjs
@@ -14,30 +14,8 @@ import importPlugin from 'eslint-plugin-import';
 import customPlugin from '@bluedot/eslint-plugin-custom';
 
 // Process xo-typescript configs:
-const removedRules = [
-  '@typescript-eslint/naming-convention',
-  '@typescript-eslint/consistent-type-assertions',
-  '@typescript-eslint/no-restricted-types',
-  '@typescript-eslint/prefer-nullish-coalescing',
-  '@typescript-eslint/no-unsafe-argument',
-  '@typescript-eslint/no-unsafe-assignment',
-  '@typescript-eslint/no-unsafe-call',
-  '@typescript-eslint/no-unsafe-declaration-merging',
-  '@typescript-eslint/no-unsafe-enum-comparison',
-  '@typescript-eslint/no-unsafe-function-type',
-  '@typescript-eslint/no-unsafe-member-access',
-  '@typescript-eslint/no-unsafe-return',
-  '@typescript-eslint/no-unsafe-type-assertion',
-  '@typescript-eslint/promise-function-async',
-];
-
 const processedXoConfig = xoConfig.map((c) => {
-  const rules = { ...c.rules };
-  for (const rule of removedRules) {
-    delete rules[rule];
-  }
-
-  const result = { ...c, rules };
+  const result = { ...c };
 
   if (c.plugins?.['@typescript-eslint']) {
     result.files = ['**/*.ts', '**/*.tsx', '**/*.mts', '**/*.cts'];
@@ -124,6 +102,7 @@ export default [
       'no-warning-comments': 'off',
       'max-params': 'off',
       'no-negated-condition': 'off',
+      curly: ['error', 'multi-line', 'consistent'],
       'no-implicit-coercion': 'off',
       'no-console': 'warn',
       'prefer-template': 'error',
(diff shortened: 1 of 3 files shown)


libraries/eslint-config-v9/src/index.mjs
```diff
@@ -0,0 +1,297 @@
+/* eslint-disable import/no-extraneous-dependencies, import/no-unresolved */
+// This file's deps are all devDependencies. The IDE doesn't know that because
+// eslint.useFlatConfig is disabled globally (for workspaces still on ESLint 8).
+// Once everything is using ESLint 9 we can flip this to true.
```
joshestein: Appreciate the explanation 🙌

libraries/eslint-config-v9/src/index.mjs
```diff
@@ -0,0 +1,297 @@
...
+  },
+
+  // ── TypeScript-only overrides ─────────────────────────────────────────
+  // Must use files pattern matching xo-typescript's TS plugin scope,
+  // otherwise ESLint throws "Could not find plugin" for non-TS files
+  {
+    files: ['**/*.ts', '**/*.tsx', '**/*.mts', '**/*.cts'],
+    rules: {
+      '@typescript-eslint/ban-ts-comment': 'off',
+      '@typescript-eslint/no-unused-vars': 'error',
```
joshestein: Can we allow unused vars with leading underscores?
```suggestion
      '@typescript-eslint/no-unused-vars': ['error', {
        argsIgnorePattern: '^_',
        varsIgnorePattern: '^_',
        caughtErrorsIgnorePattern: '^_',
      }],
```

libraries/eslint-config-v9/src/index.mjs
```diff
@@ -0,0 +1,297 @@
...
+import globals from 'globals';
+import importPlugin from 'eslint-plugin-import';
+import customPlugin from '@bluedot/eslint-plugin-custom';
+
+// Process xo-typescript configs:
+const removedRules = [
+  '@typescript-eslint/naming-convention',
+  '@typescript-eslint/consistent-type-assertions',
+  '@typescript-eslint/no-restricted-types',
+  '@typescript-eslint/prefer-nullish-coalescing',
```
joshestein: I kinda like this one 👉👈

libraries/eslint-config-v9/src/index.mjs
```diff
@@ -0,0 +1,297 @@
...
+import reactPlugin from 'eslint-plugin-react';
+import reactHooksPlugin from 'eslint-plugin-react-hooks';
+import jsxA11y from 'eslint-plugin-jsx-a11y';
+import turbo from 'eslint-config-turbo/flat';
+import globals from 'globals';
+import importPlugin from 'eslint-plugin-import';
+import customPlugin from '@bluedot/eslint-plugin-custom';
+
+// Process xo-typescript configs:
+const removedRules = [
```
joshestein: Is there any difference to be aware of between removing rules and setting them to 'off' as you've done later (L103-129 below)?

marn-in-prod (commented):
Thanks for getting this done! 

+1 to moving things from external repositories to internal ones. I'd be down to do that for `airtable-ts` potentially as well

Certain lints that I think would be generally good:

- Turning on `import/no-cycle`. We previously had this turned on (since I remember us needed to add it in a few places) and I think it catches real bugs
- turning on `no-floating-promises`. I don't think we lose very much but could be good. 
- +1 to adding `eslint-plugin-better-tailwindcss` or similar
- +1 to allowing _ for unused variables

PR #2067, opened 2026-02-16
Extract `--bluedot-navy` colour

### LINKED ISSUE #2036: Extract #13132E as a custom tailwind colour
### 🧩 Summary

#131312E has been chosen as our default grey for secondary text. It is hard-coded throughout our code and continues to be used in new designs.

Much as we extracted `bluedot-normal` we should do the same with this colour.

### ✅ Definition of done

1. #13132E has been extracted as a custom colour
2. All existing instances make use of new tailwind variable

### 📌 Notes / Designs

_No response_

joshestein:
# Description

This colour (#13132E or rgb(19, 19, 46)) is used extensively. It's time we extracted it.

1. Create a new `--bluedot-navy` colour
2. Replace all instances of existing colours with new variable
3. Prefer `text-bluedot-navy/20` to `text-bluedot-navy opacity-20` wherever possible
4. Prefer using semantic `<P>` instead of `<div>` for things like FAQ answer, course information description
5. Remove dead `CourseSearchCard`

We have an existing `bluedot-black` which is used by the `bluedot-base` class. I wanted to change the colour of bluedot-black, but was worried it would lead to too many unintended downstream consequences.

However, I do think we should change this: it would allow us to remove all these `text-bluedot-navy` calls everywhere, since the text elements would inherit the correct colour.

## Issue

Fixes #2036

## Developer checklist

## Screenshot

NA

### DIFF
--- a/apps/website/src/components/lander/TestimonialCarousel.tsx
+++ b/apps/website/src/components/lander/TestimonialCarousel.tsx
@@ -253,6 +253,7 @@ const TestimonialCarousel = ({
               </h2>
             ) : (
               <H2 className={clsx(headerSizeClasses, 'font-semibold leading-[125%] text-bluedot-navy tracking-[-0.01em]')}>
+                {/* eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing */}
                 {title || defaultTitle}
               </H2>
             )}


joshestein:
@coderabbitai review

Issue #1840, opened 2025-12-20
[SEO] robots.txt/ai.txt/llms.txt


alyssia-x:
### 🧩 Summary

A quick win in low-lift SEO optimization would be adding a robots.txt/ai.txt/llms.txt of sorts!

**What it is:** Text files that tell search engines and AI crawlers how to understand our site.

**Why it matters:**
- AI chatbots might confuse BlueDot (the nonprofit) with BlueDot (the Canadian disease tracking company), BlueDot (the SF-based AI note-taking company) or other similarly-named organizations
- When someone asks an AI assistant "What AI safety courses exist?", the AI would accurately describe BlueDot with correct details
- Points Google directly to our sitemap for faster indexing!

### ✅ Definition of done

Likely one or two files implemented in /apps/website/public/ (robots.txt/ai.txt/llms.txt) according to best practices for a quick win in low-lift SEO optimization

### 📌 Notes / Designs

There are a bunch of growing best practices for these in modern robots.txt/ai.txt files, e.g:
- Being specific: Tell models exactly how to describe your company
- Clarifying any confusions that people might already have with your brand/site, e.g explicitly state what you're NOT
- Prioritizing content: Direct models to your best, highest-priority, most accurate pages
- Keeping updated: Dating instructions + regular updates

Can research additional best practices and iterate/add to it as we go.

PR #2107, opened 2026-02-22
[feat/seo]: static llm.txt and robots.txt

alyssia-x:
# Description
Adds static `robots.txt` and `llm.txt` files based on current best practices for structure and content.

**Scope:** For this PR specifically, only course lander content is included (not full course content) since landers change infrequently, making them a stable starting point.

**Why static (not dynamic):** I considered dynamic from the get-go but deprioritized since course content lives inside `.tsx` files, which would require a larger refactor to parse cleanly.

**Potential follow-up PRs** (@marn-in-prod would love to get your thoughts):
- Include full course content (not just landers)
- Auto-generate these files to avoid multiple sources of truth

## Issue
#1840

### DIFF
--- a/apps/website/public/robots.txt
+++ b/apps/website/public/robots.txt
@@ -1,0 +1,22 @@
+# robots.txt for https://bluedot.org
+# Last updated: 2026-02-22
+
+# Allow all crawlers
+# BlueDot Impact is an educational nonprofit with a mission of building a AI safety workforce
+User-agent: *
+Allow: /
+
+# Authenticated user routes — no public SEO value, content is behind login
+Disallow: /profile
+Disallow: /settings/
+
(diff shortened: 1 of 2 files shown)


apps/website/public/robots.txt
```diff
@@ -0,0 +1,22 @@
...
+# Last updated: 2026-02-22
+
+# Allow all crawlers
+# BlueDot Impact is an educational nonprofit with a mission of building a AI safety workforce
+User-agent: *
+Allow: /
+
+# Authenticated user routes — no public SEO value, content is behind login
+Disallow: /profile
+Disallow: /settings/
```
marn-in-prod: Fix inconsistent slash patterns

PR #2113, opened 2026-02-23
[Feat] Rewrite copy on about and join-us pages

dewierwan:
# Description
The copy on the about and join us pages were stale. 

About: new hero subtitle, rewrite Who We Are section, simplify Core Beliefs and Values to text-only, update history timeline, update team member names/roles, remove JoinUsCta and Contact sections.

Join Us: rename to "Work with us", add "Why us" section combining previous Why Now and How We Work, rename jobs to "Open roles", sort Exceptional Talent role last.


dewierwan:
@claude do a final review of everything.

PR #2132, opened 2026-02-26
Support using the database in tests

### LINKED ISSUE #1987: Upgrade drizzle + drizzle-kit to a version that supports pg-lite for testing
### 🧩 Summary

[pg-lite](https://github.com/electric-sql/pglite) is a WASM build of PostgreSQL that can be used in-process, without having a separate database running, similar to SQLite. This lets you write tests that interact with a database without having a separate Docker database running or otherwise jerry-rigging the setup.

I have been starting to use this in other projects and have found it to be _incredibly_ useful. For us it would mean we could add a lot of meaningful test coverage to pg-sync-service, and add more tests on the integration to end-to-end level to website/

A version I know works is ^1.0.0-beta.3-36e9b9e

### ✅ Definition of done

1. Drizzle and Drizzle Kit are upgraded to a version that supports pg-lite
2. We have at least one test that uses this (as a demo)

### 📌 Notes / Designs

_No response_

Will-Howard:
# Description

This is the last in my bag of tricks that I have found to significantly improve developer experience. This PR sets up `pg-lite` to be used in tests, this is a WASM build of postgres that lets it run in-process without rather than needing a separate Docker container for tests.

I have also added a few example tests, to work out the kinks and provide demos of the types of cases we can cover now:
1. Standalone tests of tRPC endpoints: `exercises.test.ts`, and `group-switching.test.ts` (which was previously notably complicated and notably lacking tests)
2. Tests of react components, using the real tPRC calls: `ProfileNameEditor.test.tsx`
3. Tests of DevOps database stuff (pg-sync-service): `sync-manager.test.ts`

The core code changes are:
- `libraries/db/src/lib/client.ts`: added `isTest` mode that uses PGlite in-process and skips Airtable calls in `insert`/`update`/`remove`
- Where the consumers create the database instance (`apps/website/src/lib/api/db/index.ts` and `apps/pg-sync-service/src/lib/db.ts`): pass `isTest` to the constructor
- `libraries/db/src/lib/test-db.ts`: provides `pushTestSchema`, `resetTestDb`, and `TestPgAirtableDb` type
- The rest is testing infra and specific tests

This required bumping the `drizzle-orm` and `drizzle-kit` versions. These were minor versions and there were just some slight changes to types I had to fix in a couple of other cases. I clicked around the site and ran the tests and couldn't find any other issues.

## Issue

Fixes #1987

## Developer checklist

### DIFF
--- a/libraries/db/src/lib/client.ts
+++ b/libraries/db/src/lib/client.ts
@@ -3,9 +3,7 @@ import {
 } from 'drizzle-orm';
 import { type PgInsertValue, type PgUpdateSetSource, type PgColumn } from 'drizzle-orm/pg-core';
 import { drizzle } from 'drizzle-orm/node-postgres';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import { PGlite } from '@electric-sql/pglite';
-import { drizzle as pgLiteDrizzle } from 'drizzle-orm/pglite';
+import { type drizzle as pgLiteDrizzle } from 'drizzle-orm/pglite';
 import {
   AirtableTs, AirtableTsError, type AirtableTsOptions,
 } from 'airtable-ts';
@@ -102,14 +100,8 @@ type RestrictedPgDatabase = Omit<PgDatabase, 'insert' | 'update' | 'delete'> & {
   delete: PgDatabase['delete'];
 };
 
-function generateTestRecordId(): string {
-  return `rec${Math.random().toString(36).slice(2, 16).padEnd(14, 'a')}`;
-}
-
 export class PgAirtableDb {
   private pgUnrestricted: PgDatabase;
-
-  private isTest: boolean;
 
   public pg: RestrictedPgDatabase;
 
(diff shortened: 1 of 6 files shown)


libraries/db/src/lib/client.ts
```diff
@@ -394,7 +416,15 @@ export class PgAirtableDb {
...
     data: Partial<Omit<AirtableItemFromColumnsMap<TColumnsMap>, 'id'>>,
   ): Promise<BasePgTableType<TTableName, TColumnsMap>['$inferSelect']> {
-    const fullData = await this.airtableClient.insert(table.airtable, data);
+    let fullData: AirtableItemFromColumnsMap<TColumnsMap>;
+    if (this.isTest) {
+      // isTest => skip airtable, but replicate what it would return
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const id = (data as any).id ?? generateTestRecordId();
+      fullData = { ...data, id } as AirtableItemFromColumnsMap<TColumnsMap>;
+    } else {
```
Will-Howard: @marn-in-prod per your suggestion I removed MockAirtableTs and just handled the lack of airtable inline in insert/update/remove

marn-in-prod:
> I'm generally a fan of composition over inheritance

I generally prefer this as well. 

> can we just use an in-memory store to mock Airtable and remove the mock's dependence on PG entirely? That way it's clear we're not testing any integration with Airtable itself?

> (2) is trickier. IMO adding back MockAirtableTs would improve this. Also: I've realised that we can get rid of the double-write, if we adopt the flow in the latest version which just SELECTS the existing data rather than eagerly updating it[1]


Sounds like there's a path to making the compostion/mock path simpler. I think we're generally aligned and that @Will-Howard should move forward with that as well

PR #2184, opened 2026-03-19
Fix drizzle-kit bundling breaking Next.js API routes

dewierwan:
## Summary
- Moved `drizzle-kit/api` from a static top-level import to a dynamic `import()` inside `pushTestSchema()` in `@bluedot/db`
- This prevents drizzle-kit (which uses `require('fs')`) from being bundled into the Next.js server runtime, where dynamic requires aren't supported
- Root cause: `@bluedot/db` re-exports test utilities from its main entry point, and `test-db.ts` statically imported `drizzle-kit/api`

## Test plan
- [x] `next build` succeeds (was failing with `Dynamic require of "fs" is not supported`)
- [ ] Dev server serves API routes without 500 errors
- [ ] Tests that use `pushTestSchema()` still work (dynamic import is functionally equivalent)

### DIFF
--- a/apps/website/src/components/homepage/CourseSection.tsx
+++ b/apps/website/src/components/homepage/CourseSection.tsx
@@ -431,8 +431,8 @@ const CourseCardsGrid = ({
   otherCourses: Course[];
 }) => (
   <div className="hidden lg:flex flex-col items-center gap-8 w-full max-w-screen-xl mx-auto">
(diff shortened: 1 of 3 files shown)


marn-in-prod (approved):
Works without the fix for drizzle-kit. Just run `npm install` on your machine

PR #2218, opened 2026-03-31
Decouple .notNull() from airtable-ts return types

### LINKED ISSUE #2081: Ban `notNull()` in db schema
### 🧩 Summary

Enforcing not null has been a continuous source of problems, and as Martin points out we can't really guarantee that Airtable will return a non-null value anyway. Even in cases where `airtable-ts` can coerce null to e.g. `false`, I [ran into a problem when trying to drop such a field](https://github.com/bluedotimpact/bluedot/pull/2007#discussion_r2773919573).

### ✅ Definition of done

- `notNull()` is disallowed at the level of the `libraries/db` package
- All current schema fields that are not null are converted to handle null explicitly (this will involve meaningful code changes in some cases)

### 📌 Notes / Designs

_No response_

Will-Howard:
# Description

Decouples `.notNull()` from airtable-ts behaviour. Instead of airtable-ts changing what it returns based on notNull status, we now set per-type defaults:
- Strings, booleans, arrays: always non-null (coerced to `''`, `false`, `[]`)
- Numbers: always nullable (need to distinguish `null` vs `0`)

Small change (3 files) but broad effect since it feeds into schema generation for every table.

Tested: all CI passes, Playwright spot-checks of courses page, FoAI landing, home page look fine. One code fix needed in `course-rounds.ts` (`!== null` -> `!!`).

Risk: code relying on getting `null` for a string/boolean/array would now get the zero value instead. Unlikely to be user-visible since most code guards with `!!`.

Because of the breadth of this change, I think there is some chance of it causing bugs or slight changes in behaviour. However, it's easy to revert (one full sync of the db will migrate to or from the new behaviour), so it's not a big risk to merge.

## Issue

Fixes #2081

## Developer checklist

### DIFF
--- a/apps/website/src/server/routers/course-rounds.ts
+++ b/apps/website/src/server/routers/course-rounds.ts
@@ -131,7 +131,7 @@ export type CourseRoundsData = Awaited<ReturnType<typeof getCourseRoundsData>>;
 export function getSoonestDeadline(rounds: CourseRoundsData): string | null {
   const allRounds = [...rounds.intense, ...rounds.partTime];
 
-  const roundsWithDeadlines = allRounds.filter((r): r is typeof r & { applicationDeadlineRaw: string } => r.applicationDeadlineRaw !== null);
+  const roundsWithDeadlines = allRounds.filter((r): r is typeof r & { applicationDeadlineRaw: string } => !!r.applicationDeadlineRaw);
 
   if (roundsWithDeadlines.length === 0) {
     return null;
--- a/libraries/db/src/lib/typeUtils.test.ts
+++ b/libraries/db/src/lib/typeUtils.test.ts
@@ -37,18 +37,20 @@ const booleanNull: BooleanType = null;
 // @ts-expect-error - should not accept strings
 const booleanInvalid: BooleanType = 'test';
 
-// Test that array types are correct
 const textArrayValue: TextArrayType = ['test', 'test2'];
+// @ts-expect-error - should not accept null for arrays
 const textArrayNull: TextArrayType = null;
 // @ts-expect-error - should not accept single strings
 const textArrayInvalid: TextArrayType = 'test';
 
 const numericArrayValue: NumericArrayType = [1, 2, 3];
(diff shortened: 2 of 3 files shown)


Will-Howard:
Re: DrizzleColumnToTsType not being updated: I've intentionally left this as-is. DrizzleColumnToTsType is only used for the write side (db.insert() / db.update() input types) and the airtable-ts table schema. The read types come from drizzle's $inferSelect on the pg columns, which is separate.

Keeping | null on the write type means callers can still pass null to clear a field in Airtable, which is valid. Making it stricter would require changing ~10 write sites to use undefined instead of null for no runtime benefit.

marn-in-prod (approved):
> Re: DrizzleColumnToTsType not being updated: I've intentionally left this as-is. DrizzleColumnToTsType is only used for the write side (db.insert() / db.update() input types) and the airtable-ts table schema. The read types come from drizzle's $inferSelect on the pg columns, which is separate.

> Keeping | null on the write type means callers can still pass null to clear a field in Airtable, which is valid. Making it stricter would require changing ~10 write sites to use undefined instead of null for no runtime benefit.

 didn't see this comment until afterwards, I'll need a second to look at this.

marn-in-prod (changes requested):
Okay, I looked through the code around your comment. `drizzleColumnToTsTypeString` builds the typing for `airtable-ts`. 

With these changes, `text().array()` and `text().array().notNull()` will both produce `string[]` as the type. 

`nameMapper` in `airtable-ts` throws an error if you null out an array not labelled as nullable. 

So we should modify the TypeScript type so we don't pass nulls into arrays or we modify `airtable-ts`. This check in nameMapper only exists for arrays for whatever reason

Will-Howard:
> So we should modify the TypeScript type so we don't pass nulls into arrays or we modify airtable-ts.

Fixed! I did the first option here.

Will-Howard:
(I will merge this in a moment, I'm just waiting for a time when I can monitor it after it goes out)

PR #2250, opened 2026-04-09
Use emptyDir for MCP token storage (delete stuck PVCs, unblock CD)

domdomegg:
CD has been red since #2247 — the three PVCs are stuck `Pending` (originally Vultr quota; now CSI exponential backoff after a day of failures).

This swaps the three MCP volumes to `emptyDir` and removes the PVC resources, which makes Pulumi delete the stuck PVCs and lets the services actually come up. Tokens are lost on pod restart while on emptyDir — fine, no one's using it yet.

Follow-up PR adds the PVCs back at 5Gi with `skipAwait`.


domdomegg:
Merge this if the CI failure on main is blocking; will mean that tokens get reset on each restart but we can revert once vultr gives us more capacity. Have put in a support ticket, waiting on them to raise account limits.

PR #2268, opened 2026-04-12
[codex] Fix course lander crash on client navigation

somepersonnamedjosh:
## Summary
Fix a client-side crash when navigating into certain course landers, including Technical AI Safety Project, from another page.

## Root Cause
The lander CTA URL path assumed the application URL was always defined during client navigation. In practice, the value can be briefly missing while the page is transitioning. When that happened, the query-param helper tried to call `.split()` on an undefined URL and the page crashed.

A full refresh worked because the lander then rendered with complete data.

## What Changed
- made `addQueryParam` tolerate `null` and `undefined` input in `libraries/ui/src/utils/addQueryParam.ts`
- made `CourseLander` normalize a missing `baseApplicationUrl` before appending UTM parameters in `apps/website/src/components/lander/CourseLander.tsx`
- added regression coverage for both the helper and the lander crash case

## Impact
- client navigation to `/courses/technical-ai-safety-project` no longer crashes if the application URL is temporarily unavailable
- the shared URL helper is more defensive against similar transient states elsewhere

## Validation
- `npm exec --workspace @bluedot/ui vitest run src/utils/addQueryParam.test.ts`
- `npm exec --workspace apps/website vitest run src/components/lander/AgiStrategyLander.test.tsx`
- `npm exec --workspace apps/website tsc --noEmit`
- `npm exec --workspace apps/website eslint src/components/lander/CourseLander.tsx src/components/lander/AgiStrategyLander.test.tsx`
- `npm exec --workspace @bluedot/ui eslint src/utils/addQueryParam.ts src/utils/addQueryParam.test.ts`

### DIFF
--- a/apps/website/src/components/lander/AgiStrategyLander.test.tsx
+++ b/apps/website/src/components/lander/AgiStrategyLander.test.tsx
@@ -10,6 +10,7 @@ import { server, trpcMsw } from '../../__tests__/trpcMswSetup';
 
 // Test URL - in production this comes from the database
 const TEST_APPLICATION_URL = 'https://web.miniextensions.com/test';
+const mockLatestUtmParams: { utm_source: string | undefined } = { utm_source: undefined };
 
 // Mock Next.js Head component
 vi.mock('next/head', () => ({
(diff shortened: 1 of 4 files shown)


somepersonnamedjosh:
@claude review pls

PR #2274, opened 2026-04-13
Clean up defensive workarounds from #2268

### LINKED ISSUE #2271: Fix types that led to https://github.com/bluedotimpact/bluedot/pull/2268
### 🧩 Summary

`latestUtmParams` is typed as  `Record<string, string>`. In truth, it is `Record<string, string | undefined>`. The PR in the title fixed  the symptoms but didn't fix the underlying type dishonesty.

The cause of the problem is that `latestUtmParams` [is initialised to an empty object ](https://github.com/bluedotimpact/bluedot/blob/6a09129984c6233977bfabc97ace64ac98bf5f5d/libraries/ui/src/hooks/useLatestUtmParams.tsx#L36), which results in `utm_source` being a missing key, and so  `latestUtmParams.utm_source` is undefined.

### ✅ Definition of done

Change the type to `Partial<Record<string, string>>`, update callers as appropriate.

Make sure to update `courses/index.tsx`, since it is currently susceptible to the same bug.

### 📌 Notes / Designs

_No response_

joshestein:
Closes #2271

## Summary

PR #2268 fixed a crash in `CourseLander` caused by `baseApplicationUrl` being `undefined` at runtime during client navigation. As part of that fix, `addQueryParam` was made to accept `null | undefined`, and a double guard (`utm_source && baseApplicationUrl`) was added at call sites. This PR removes those workarounds in favour of a cleaner approach.

## What changed

- Reverted `addQueryParam` to `url: string` - callers normalise nullable URLs upfront with `?? ''` before calling, which is clearer about intent
- Replaced `|| ''` with `?? ''` in `courses/index.tsx`, removing the `eslint-disable` comment
- Removed the `null`/`undefined` URL tests from `addQueryParam.test.ts` since those inputs are now type errors

## Notes

The issue title mentions fixing the `latestUtmParams` type (`Record<string, string>` implying all keys always exist). In practice `noUncheckedIndexedAccess` is already enabled, so index access on `Record<string, string>` already returns `string | undefined` - a `Partial` wrapper would be redundant. The original crash was not caused by this type gap; it was caused by `baseApplicationUrl` being typed as a required `string` prop when it could be transiently `undefined` during client navigation, which was fixed in #2268.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/website/src/components/lander/CourseLander.tsx
+++ b/apps/website/src/components/lander/CourseLander.tsx
@@ -81,7 +81,7 @@ const CourseLander = ({
 }: CourseLanderProps) => {
   const { latestUtmParams } = useLatestUtmParams();
   const applyUrl = baseApplicationUrl ?? '';
-  const applicationUrlWithUtm = appendPosthogSessionIdPrefill(latestUtmParams.utm_source
+  const applicationUrlWithUtm = appendPosthogSessionIdPrefill(latestUtmParams.utm_source && applyUrl
     ? addQueryParam(applyUrl, 'prefill_Source', latestUtmParams.utm_source)
     : applyUrl);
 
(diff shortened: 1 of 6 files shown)


joshestein:
@greptileai review

joshestein:
@greptileai review please!

PR #2326, opened 2026-04-26
[refactor] website: standardise canvas to pure white

dewierwan:
## Summary

- The bluedot.org site was rendering some pages on cream (#FCFBF9) and others on pure white. The mismatch was between pages, not within them — homepage and `/courses` were white via per-element `bg-white` overrides; `/programs` was white via a `mainBgWhite` opt-in flag in `_app.tsx`; `/about`, `/blog`, `/join-us`, `/agi-strategy` were cream because they didn't override anything.
- This PR standardises on **pure white**. One token override at the website-app level (`--bluedot-cream-normal: #ffffff` in `apps/website/src/globals.css`) flips the canvas; every now-redundant per-page `bg-white` is removed, and the `mainBgWhite` escape hatch in `_app.tsx` is deleted.
- Library-side `--bluedot-cream-normal` is left at `#FCFBF9` so the other monorepo apps that still want cream are unaffected.

## Screenshots

`/about` is where the change is most visible: it switches from cream to white for the first time. Home / courses / programs look near-identical because they were already white via overrides — this PR just routes them through the canvas instead.

### `/about`

|            | Before          | After               |
| ---------- | --------------- | ------------------- |
| 📱 Mobile  |  |  |
| 🖥️ Desktop |  |  |

### ` / ` (home)

|            | Before          | After               |
| ---------- | --------------- | ------------------- |
| 📱 Mobile  |  |  |
| 🖥️ Desktop |  |  |

### `/courses`

|            | Before          | After               |
| ---------- | --------------- | ------------------- |
| 📱 Mobile  |  |  |
| 🖥️ Desktop |  |  |

### `/programs`

|            | Before          | After               |
| ---------- | --------------- | ------------------- |
| 📱 Mobile  |  |  |
| 🖥️ Desktop |  |  |

## Test plan

- [x] `npm test` in `apps/website` (530/530 passing; 2 snapshots updated for the homepage sections that lost their `bg-white` class)
- [x] `npm run lint` clean
- [x] Walked `/`, `/courses`, `/programs`, `/about` in dev — canvas reads consistently white across pages
- [ ] Reviewer to spot-check elevated UI (modals, dropdowns, cards on `/about`) — borders + shadows now do the elevation work without canvas contrast


dewierwan:
@claude review

dewierwan:
Good catch — both Greptile and @claude flagged the same thing. The old `mainBgWhite` was doing two things: setting the white background AND, via the `||` in `_app.tsx:32`, forcing `mainShrinkToContent` to `true` which set `minHeight: 'auto'` to override `bluedot-base`'s `min-h-screen`. I removed the flag and the `||` together, which dropped the second behaviour by accident.

Practical impact is small (`/programs` content is ~1300–1700px on desktop, only short of `min-h-screen` on viewports > ~1500px tall) but real — would have shown a blank slab between the newsletter and the footer on large monitors.

Fixed in e6e1513 by re-declaring `ProgramsPage.mainShrinkToContent = true` directly on the page. Restores the prior layout contract without resurrecting the canvas coupling.

PR #2338, opened 2026-04-26
[test] website: cover course-registrations tRPC router

dewierwan:
## Summary
- 14 tests across the 4 procedures in \`src/server/routers/course-registrations.ts\` (\`getByCourseId\`, \`getAll\`, \`getRoundStartDates\`, \`ensureExists\`).
- Reuses the in-memory test DB pattern from \`admin.test.ts\` / \`grants.test.ts\`.

## Why
Part of a focused testing pass on tRPC routers. \`course-registrations.ts\` is on the critical path for course access (FOAI auto-registration, registration lookup) and was completely untested.

## What's covered
- \`getByCourseId\`: rejects logged-out callers, returns null when no Accept registration, returns the right row, ignores non-Accept decisions, isolates by email.
- \`getAll\`: rejects logged-out callers, returns non-Withdrawn rows for the auth email (including null-decision rows), excludes Withdrawn and other users' rows.
- \`getRoundStartDates\`: rejects logged-out callers, short-circuits on empty input, returns roundId → firstDiscussionDate map (including null entries), ignores rounds not requested.
- \`ensureExists\`: rejects logged-out callers, returns existing registration without creating a duplicate, returns null for non-FOAI courses, throws \`NOT_FOUND\` when FOAI applications_course config row is missing.

## What's not covered (intentional, flagged in source)
The "creates a new FOAI registration" branch can't be exercised cleanly. The router's insert omits the \`courseId\` column, but \`courseId\` is \`notNull()\` in the pg schema, so the test fails with a constraint violation. This either reveals a real router bug or a divergence between the test pg schema and Airtable's runtime (likely a formula populating courseId from courseApplicationsBaseId). Worth a separate look — a comment in the test file points at the relevant lines.

## Test plan
- [x] \`npm run lint\` clean
- [x] \`npm run typecheck\` clean
- [x] \`npx vitest run src/server/routers/course-registrations.test.ts\` — 14/14 pass

### DIFF
--- a/apps/website/src/server/routers/course-registrations.test.ts
+++ b/apps/website/src/server/routers/course-registrations.test.ts
@@ -1,0 +1,155 @@
+import {
+  applicationsRoundTable, courseRegistrationTable,
+} from '@bluedot/db';
+import { describe, expect, test } from 'vitest';
+import {
+  createCaller, setupTestDb, testAuthContextLoggedIn, testAuthContextLoggedOut, testDb,
+} from '../../__tests__/dbTestUtils';
+import { FOAI_COURSE_ID } from '../../lib/constants';
+
+setupTestDb();
+
+const otherCourseId = 'recOtherCourseId12345';
+
+describe('courseRegistrations.getByCourseId', () => {
+  test('rejects unauthenticated callers', async () => {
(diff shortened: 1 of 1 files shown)


dewierwan:
@claude review

dewierwan:
@claude review again with the latest commit

PR #2329, opened 2026-04-26
[perf] ci: parallelise typecheck/lint/test, cache deps + .next + .turbo

dewierwan:
## Summary

CI/CD on master push currently takes **9-12 min**, PR runs **4-7 min**. This PR is the no-infra quick-win bundle:

- **Splits `ci` into parallel jobs** (`setup` → `typecheck` / `lint` / `test`) sharing a `node_modules` cache. Aggregator job kept named `ci` so branch protection (`required_status_checks.contexts: ["ci"]`) still passes.
- **Caches `node_modules` in `cd`** and `website_deploy_production` (was reinstalling from scratch every run, ~44s).
- **Caches `apps/website/.next/cache`** between deploys for faster `next build`.
- **Caches `.turbo`** so unchanged tasks short-circuit even without remote cache.
- **`concurrency: cancel-in-progress`** on PR runs so stacked pushes don't all run to completion.
- **`npm ci --prefer-offline`** wherever we install fresh.

## Baseline (run [`24964029299`](https://github.com/bluedotimpact/bluedot/actions/runs/24964029299), master push)

| Job | Step | Time |
|---|---|---|
| `ci` (4m19s, sequential) | Tests | 2m41s |
|  | Lint | 0m50s |
|  | Typecheck | 0m15s |
|  | Setup + npm cache | ~30s |
| `cd` (4m55s) | `npm ci` (no cache) | 0m44s |
|  | `next build` + docker + push + rollout | 4m01s |

## Expected impact

- **PR runs:** 4-7m → ~2-3m (CI critical path = max(typecheck, lint, test) ≈ test ~3m, vs. serial 4m19s today; cancel-in-progress kills stacked runs).
- **Master push:** 9-12m → ~5-6m (cd starts 1m earlier, then warm `node_modules` saves 40s, warm `.next/cache` saves 30-90s on `next build`).
- **Steady-state cache hits** (no app code changed): turbo's local cache lets typecheck/lint/test no-op; saved further.

## Branch protection

Verified `gh api repos/bluedotimpact/bluedot/branches/master/protection` requires the check named `ci`. Aggregator job preserves this. **No branch-protection changes needed.**

## Out of scope (deliberate, follow-up)

- Turbo remote cache (Vercel free tier or self-hosted on R2/S3) — biggest remaining lever.
- Decoupling `cd` from `test` (run deploy in parallel with tests on master). Requires confirming PR-level test-gate is sufficient.
- `vitest --shard` test parallelisation across runners.
[...]

PR #2348, opened 2026-04-27
[fix] ci/cd: retry + fail-loud on flaky GitHub API responses

dewierwan:
## Context

Today's deploys of PR #2347 hit two separate failure modes — both rooted in the same flaky GitHub Actions API behaviour:

- **`/actions/workflows/{id}/runs?status=success`** intermittently returns an empty page even when thousands of matching runs exist.
- **`listWorkflowRuns({head_sha})`** has the same intermittent-empty-page problem.

Both gates trusted a single API call with no retries.

## Failure modes seen today

1. **Silent staging skip.** PR #2347 merged, `ci_cd` ran green, but the `cd` job was skipped because `affected_deployable_apps == '[]'`. Setup logged `Successful parent commits:` (empty) and `0 packages changed` even though 11 files were committed. Root cause: the API returned `{workflow_runs: []}` for `?status=success`, the parent walk found nothing, and the empty result silently flowed through to the deploy gate.
2. **Loud production failure.** `website_deploy_production` for `website/v2.24.18` (and again for `v2.24.19`) failed instantly with `CI/CD has not passed for <sha> (status: not found, conclusion: N/A)`. The `Verify CI/CD has passed` step reads only the first run via `workflow_runs: [run]` destructuring, so any empty/stale response = whole-pipeline failure.

Re-running both workflows worked as soon as the API recovered, confirming it's transient — but a transient API blip should not silently skip a deploy or block a release.

### `libraries/affected-packages/src/lib/getPackagesWithChanges.ts`

- Wrap the success-runs API call in a 3-attempt retry with exponential backoff (1s/2s/4s) via a new `fetchSuccessfulCommitShas` helper.
- Bump `per_page` from the implicit 30 → 100, so the immediate-parent successful run is far less likely to fall off the first page during burst-PR periods.
- **Throw** if every attempt returns empty, instead of silently returning `[]`. Better to fail a deploy loudly than skip it silently.
- Also **throw** if the parent walk yields no successful ancestor despite a non-empty SHA list — closes the second silent-skip path (e.g. relevant ancestor not in the page).

### `.github/workflows/website_deploy_production.yaml`

- Add `retries: 3` to the `Verify CI/CD has passed` `actions/github-script` step (transport-level retries on 5xx).
- Replace the single-call, first-run-only check with a 5-attempt retry loop that **searches every returned run** for a `conclusion === 'success'` entry. Avoids both the empty-page and the stale-first-run failure modes.
- On final failure, surface what the API actually returned (`status/conclusion` for every run) instead of just `not found`, to make next time's debugging quicker.

### `libraries/affected-packages/src/lib/getPackagesWithChanges.test.ts`

- Update the mocked `gh api` URL to match the new `&per_page=100` query string.

## Notes

- The bug is in code that predates PR #2329 (the recent CI parallelisation). #2329 didn't introduce or amplify the silent-skip; it just sat next to it.
- Scope intentionally tight: only the prod-deploy verify gate is patched, not other workflows. If this lands cleanly, easy follow-up to apply the same retry pattern anywhere else `listWorkflowRuns` is used as a gate.

## Test plan

- [x] `libraries/affected-packages` vitest suite passes (mock URL updated for new query param).
- [x] `tsc --noEmit` clean.
- [x] `eslint` clean.
- [ ] Reviewer: spot-check the YAML script block by reading the diff (no local actionlint).
- [ ] After merge: confirm next master push runs `cd` and deploys to staging (would have skipped today otherwise).

### DIFF
--- a/.github/workflows/website_deploy_production.yaml
+++ b/.github/workflows/website_deploy_production.yaml
@@ -20,7 +20,6 @@ jobs:
         if: ${{ !inputs.skip_ci_check }}
         uses: actions/github-script@v7
         with:
-          retries: 3
           script: |
             // GitHub's listWorkflowRuns endpoint occasionally returns an
             // empty page or only stale entries during transient API blips,
--- a/.github/workflows/website_deploy_production.yaml
+++ b/.github/workflows/website_deploy_production.yaml
@@ -20,14 +20,32 @@ jobs:
         if: ${{ !inputs.skip_ci_check }}
         uses: actions/github-script@v7
         with:
+          retries: 3
           script: |
-            const { data: { workflow_runs: [run] } } = await github.rest.actions.listWorkflowRuns({
-              owner: context.repo.owner, repo: context.repo.repo,
-              workflow_id: 'ci_cd.yaml', head_sha: context.sha,
-            });
-            if (!run || run.status !== 'completed' || run.conclusion !== 'success') {
-              core.setFailed(`CI/CD has not passed for ${context.sha} (status: ${run?.status ?? 'not found'}, conclusion: ${run?.conclusion ?? 'N/A'})`);
+            // GitHub's listWorkflowRuns endpoint occasionally returns an
+            // empty page or only stale entries during transient API blips,
+            // so retry a few times and search every run for the SHA rather
+            // than trusting the first one returned.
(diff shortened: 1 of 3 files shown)


dewierwan:
@claude review

dewierwan:
@claude review the latest commits and all the code diffs here

PR #2294, opened 2026-04-22
Migrate icons to use `IconProps`

joshestein:
# Description

Migrate custom icons to new `IconProps` interface.

## Issue

Closes #2291.

## Developer checklist

- [x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [x] Considered having snapshot tests and/or happy path functional tests
- [x] Considered adding Storybook stories

## Screenshot

NA

PR #2374, opened 2026-04-29
[chore] website/icons: add barrel export, document icon-library policy, sweep call sites

dewierwan:
## Summary

The 27 custom icons in `apps/website/src/components/icons/` had no barrel — every consumer imported per-file (`../icons/CheckIcon`, `../icons/ClockIcon`, …), so a file using 5 icons had 5 separate import lines and the catalog wasn't discoverable from one spot.

- Add `components/icons/index.ts` re-exporting every icon plus the `IconProps` type, with a header comment documenting the icon-library policy: prefer `react-icons/fa6` (Font Awesome 6 — the team standard per Will/Josh in the design-system Slack thread) or `react-icons/pi` (Phosphor — already used in lander course-content blocks). Only add a custom file when no library icon matches and the design is genuinely bespoke. New custom icons must implement `IconProps` for consistency with Josh's #2294 work.
- Sweep the 23 consumer files to import from `'../icons'` (or `'../../icons'` etc.) instead of `'../icons/<Name>'`, collapsing 19 extra import lines into combined imports. Imports are sorted within each combined group.

This unblocks part of Mallorie/Cyrus' design-system question (which icons exist? which library to use for new ones?) — the barrel comment is the first answer; the gallery at https://storybook.k8s.bluedot.org/?path=/story/website-icons-gallery--all-icons remains the visual catalog.

## Visual

Byte-equivalent at runtime — same exports, same components. The barrel just gives one place to look at the catalog and one path for consumers.

## Test plan

- [x] `npx vitest run` (apps/website): 103 files / 629 passed / 1 skipped — no snapshot updates needed
- [x] `tsc --noEmit` clean — types resolve through the barrel correctly


dewierwan:
@claude review

dewierwan:
Pushed `b1679d4d` addressing the Claude bot's P2.

**P2 — `Icons.stories.tsx` still using 27 individual imports** — fixed. Switched to a single barrel import via `from '.'`. The stories file was overlooked in the original sweep because my script targeted external consumers (different directories); this brings the in-folder gallery in line with the documented policy.

**Stylistic — multi-line barrel imports in some files** — left as-is. That's prettier wrap behaviour when the combined import line exceeds the 120-col limit; functionally identical to single-line. Not a meaningful problem to fix.

🤖 Reply drafted by Claude Code

dewierwan:
@claude review

PR #2375, opened 2026-04-29
[chore] save-status: replace #1641D9 with bluedot-normal token

dewierwan:
## Summary

The save-status indicator and rich-text editor both used a one-off blue hex `#1641D9` (rgb 22, 65, 217) — visually close to `--bluedot-normal` (#1144CC, rgb 17, 68, 204) but a slightly different shade, so the "saving" spinner and the editor focus border were drifting from the canonical brand blue used everywhere else.

Two swaps:

- `SaveStatusIndicator.tsx`: `style={{ color: '#1641D9' }}` → `className="… text-bluedot-normal"`. Drops the inline style for a className that goes through the design-system token.
- `RichTextAutoSaveEditor.tsx`: `focus-within:border-[#1641D9]` → `focus-within:border-bluedot-normal` (still keeping the 1.25px width + box-shadow halo, which are bespoke focus-state effects).

The `rgba(34, 68, 187, 0.3)` halo shadow elsewhere on the editor is left alone — it's a derived alpha + slightly different blue and would need either a new token or an inline `var(--bluedot-normal)` with `color-mix(...)` to be tokenised cleanly. Separate PR.

## Test plan

- [x] `npx vitest run` (apps/website): 103 files / 629 passed / 1 skipped (3 snapshots updated to reflect class swaps)
- [x] No new visual regressions — colour shift is rgb(22, 65, 217) → rgb(17, 68, 204), imperceptible at the spinner / 1px border level

### DIFF
--- a/apps/website/src/components/courses/exercises/RichTextAutoSaveEditor.tsx
+++ b/apps/website/src/components/courses/exercises/RichTextAutoSaveEditor.tsx
@@ -247,7 +247,7 @@ const RichTextAutoSaveEditor: React.FC<RichTextAutoSaveEditorProps> = ({
     'resize-y overflow-auto relative cursor-text',
     'box-border w-full bg-white rounded-md z-[1] p-4',
     'border-[0.5px] border-bluedot-navy/25',
-    'focus-within:border-[1.25px] focus-within:border-[#1641D9] focus-within:shadow-[0px_0px_10px_rgba(34,68,187,0.3)]',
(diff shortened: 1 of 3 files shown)


dewierwan:
@claude review

dewierwan:
Bot review summary: Greptile 5/5, Claude bot ✅ approved. No findings to address.

🤖 Reply drafted by Claude Code

PR #2378, opened 2026-04-29
[fix] website_deploy_production: poll ci_cd until terminal

dewierwan:
## Summary

- The deploy gate failed run [25119610438](https://github.com/bluedotimpact/bluedot/actions/runs/25119610438) (v2.25.3) not because CI failed — it was still running. ci_cd needed ~12 min; the gate gave up after ~26 s with `in_progress/N/A`.
- Root cause: the loop only retried 5 times with exponential backoff (1+2+4+8 s) and treated `in_progress` as a failure, conflating "API blip / no run visible yet" with "run is still going".
- Fix: state-aware poll. Tolerate `no run found` for 60 s (API blips), keep polling on `queued`/`in_progress` up to a 30 min ceiling, fail fast on any non-`success` conclusion. Bump the `deploy` job's `timeout-minutes` 20 → 45 so the wait fits alongside the deploy itself.

## Test plan

- [ ] Logic check: tag-pushed deploy where ci_cd is already green completes in one poll.
- [ ] On the next legitimate `website/v*` tag, push the master commit and the tag close together to deliberately recreate the original race; gate should log `status=in_progress` then `succeeded` and proceed.
- [ ] If ci_cd actually fails for the SHA, gate should fail fast with the conclusion, not wait 30 min.
- [ ] If the tag is pushed without a corresponding master push, gate fails at 60 s with "No ci_cd run found".
- [ ] `skip_ci_check` workflow_dispatch input still bypasses the gate.

### DIFF
--- a/.github/workflows/website_deploy_production.yaml
+++ b/.github/workflows/website_deploy_production.yaml
@@ -14,37 +14,60 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 45
     steps:
-      - name: Verify CI/CD has passed
+      - name: Wait for CI/CD to pass
         if: ${{ !inputs.skip_ci_check }}
         uses: actions/github-script@v7
         with:
           script: |
-            // GitHub's listWorkflowRuns endpoint occasionally returns an
-            // empty page or only stale entries during transient API blips,
-            // so retry a few times and search every run for the SHA rather
-            // than trusting the first one returned.
+            // The matching ci_cd run for this SHA may still be running when the
+            // tag push fires, so poll listWorkflowRuns until it reaches a
+            // terminal state. Distinguishes three cases:
+            //   - no run visible yet: API blip, tolerate up to 60s
+            //   - run queued / in_progress: keep polling up to the 30 min ceiling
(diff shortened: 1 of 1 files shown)


dewierwan:
@claude review

dewierwan:
Pushed 508e7ed6 addressing both Greptile P2s.

- **runEverSeen flag** (Greptile's apiBlipDeadline reset concern): once we observe a run, a later empty response is treated as a transient API result, not a missing run. The 60 s guard now only fires before the first sighting.
- **timeout-minutes 45 → 60**: gives the deploy step ~30 min headroom even when polling consumes the full 30 min ceiling.

Skipping Claude bot's two minor notes:
- `per_page: 100` is filtered by `head_sha`, so realistically 1–3 runs per SHA — overflow isn't a real risk.
- The sort is intentional, not assuming one run; we pick the most recent attempt and treat re-runs correctly.

@claude review

dewierwan:
Pushed 90e1a4f6 with CodeRabbit's fix: added \`event: 'push'\` to the \`listWorkflowRuns\` query so only the master-push CI run can satisfy the gate. Catches the rebase-and-merge edge case where a PR run could share a SHA with the master commit and finish first.

PR #2432, opened 2026-05-06
[feat] db: add linkPreviewImage column to job_posting

dewierwan:
## Summary

- Adds `linkPreviewImage: text()` to `jobPostingTable` (`libraries/db/src/schema.ts:728`), mapped to Airtable field `[*] Image` (`fldXua1PI4iXLKCKm`).
- That Airtable field is a formula that outputs a permanent miniextensions-hosted URL for whatever attachment is in the `Link preview image` field. Same pattern as `imageAttachmentUrls` on `teamMemberTable` (`schema.ts:1480`).
- Schema-first PR. Once merged, pg-sync will materialise the column in staging and prod automatically. A follow-up PR will read `job.linkPreviewImage` in `apps/website/src/pages/join-us/[slug].tsx` and use it as the og:image source.

## Why this column

Today every role page falls back to BlueDot's generic logo link-preview unless someone hand-commits a `public/images/jobs/link-preview/<slug>.png`. Only `biosecurity.png` exists. Wiring this through Airtable lets ops upload a per-role share image without a code change.

## Test plan

- [x] `npm test` in `libraries/db` (4 files, 22 tests pass)
- [x] `npm run lint` in `libraries/db`
- [ ] Reviewer: confirm field name + Airtable id are correct (visible in [the field URL](https://airtable.com/app63L1YChHfS6RJF/tblGv8yisIfJMjT6K/viwk9RrUJNeeOjq1F/fldXua1PI4iXLKCKm))
- [ ] After merge: confirm pg-sync-service auto-backfills `link_preview_image` on next sync (per [reference_pg_sync_auto_backfill](https://github.com/bluedotimpact/bluedot)) before opening the consumer PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2433, opened 2026-05-06
[feat] join-us: use Airtable image as og:image

dewierwan:
## Summary

- Uses the new `job.linkPreviewImage` column (added in #2432) as the og:image source for `/join-us/<slug>` pages.
- The Airtable `[*] Image` formula always returns a miniextensions URL — including for jobs that have no `Link preview image` attachment uploaded, where the URL 500s. Added a HEAD-probe (`isUrlReachable`, 3s timeout) before trusting the URL, so missing-image jobs fall through to the existing static-PNG path and then to the generic logo.
- New util `apps/website/src/utils/isUrlReachable.ts`, mirroring the `fileExists.ts` pattern.

## Stacked PR

Base is `dewi/job-posting-link-preview-image-schema` (#2432) — once that merges, I'll rebase this onto `master`.

⚠️ Do not merge until #2432 is merged AND pg-sync has materialised the `link_preview_image` column. Otherwise `getStaticProps` errors and every `/join-us/<slug>` page 500s.

## Why HEAD-probe

The miniextensions URL `https://web.miniextensions.com/api/public-attachments/.../recXXX/1` is a deterministic function of the record id, not the attachment, so it exists for every record. With no attachment behind it, it returns 500. Without a probe, every job without a link-preview image would get a broken og:image instead of falling back to the BlueDot logo. Future-proofs against ops uploading images for some roles but not others.

The HEAD request runs server-side inside ISR (5min revalidate), so it's at most one network call per slug per 5min per server.

## Followups (not in this PR)

- Suggest changing the Airtable formula `[*] Image` to gate on `{Link preview image}` so it returns empty when there's no attachment. Would let us drop the HEAD probe.
- Migrate the existing `biosecurity.png` static asset to Airtable and remove `apps/website/public/images/jobs/link-preview/`.

## Test plan

- [x] `npm test` in `apps/website` (104 files, 631 tests pass)
- [x] `npm run lint` in `apps/website`
- [ ] After #2432 merges + pg-sync materialises column: rebase, push, verify locally with `curl -s http://localhost:8000/join-us/course-ops-lead | grep og:image` shows the miniextensions URL
- [ ] Same check on a slug without an image (e.g. `head-of-operations`) shows the logo fallback, not a broken URL

### DIFF
--- a/apps/website/src/__tests__/pages/join-us/[slug].test.tsx
+++ b/apps/website/src/__tests__/pages/join-us/[slug].test.tsx
@@ -45,6 +45,7 @@ const mockJob: JobPosting = {
   publishedAt: 1609459200,
   publicationStatus: 'Published',
   category: null,
+  linkPreviewImage: null,
 };
 
 describe('JobPostingPage SSR/SEO', () => {
--- a/apps/website/src/components/join-us/JobsListSection.test.tsx
+++ b/apps/website/src/components/join-us/JobsListSection.test.tsx
@@ -14,6 +14,7 @@ const mockCmsJobs = [
     publicationStatus: 'Published' as const,
     publishedAt: Date.now() / 1000,
     category: null,
+    linkPreviewImage: null,
   },
   {
     id: '2',
@@ -24,6 +25,7 @@ const mockCmsJobs = [
     publicationStatus: 'Published' as const,
     publishedAt: Date.now() / 1000,
     category: null,
+    linkPreviewImage: null,
   },
 ];
 
--- a/apps/website/src/pages/join-us/[slug].tsx
+++ b/apps/website/src/pages/join-us/[slug].tsx
@@ -14,6 +14,7 @@ import { ONE_MINUTE_SECONDS } from '../../lib/constants';
 import MarkdownExtendedRenderer from '../../components/courses/MarkdownExtendedRenderer';
 import db from '../../lib/api/db';
 import { fileExists } from '../../utils/fileExists';
+import { isUrlReachable } from '../../utils/isUrlReachable';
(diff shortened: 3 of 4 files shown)


dewierwan:
@claude review

dewierwan:
Rebased onto master now that #2432 is in. Two changes since the previous review:

**Dropped the HEAD probe entirely.** The Airtable `[*] Image` formula now gates on `{Link preview image}` and returns empty when there's no attachment, so `job.linkPreviewImage` being truthy is sufficient. Deleted `apps/website/src/utils/isUrlReachable.ts` (single-caller, now dead).

**Dropped three og:image:* meta tags** (`og:image:type`, `og:image:width`, `og:image:height`). They were hardcoded to PNG / 1200×630 — fine for the static-PNG path but a lie for arbitrary-size Airtable uploads. Both reviewers flagged this (Greptile P2 + Claude bot suggestion). Platforms infer these from the actual response.

**Verified locally:**
- biosecurity → static-PNG fallback (unchanged) ✓
- head-of-operations (no attachment) → generic logo fallback ✓
- course-ops-lead → currently `linkPreviewImage` is `NULL` in pg (column DDL applied, row backfill pending pg-sync's next cycle) — falls through to logo for now. Will switch to the miniextensions URL automatically once pg-sync re-syncs that row.

**Bot suggestions intentionally NOT taken:**
- Greptile's `redirect: 'manual'` SSRF guard on `isUrlReachable` — moot, file deleted.
- Claude bot's `content-type` check — moot, probe deleted.
- Claude bot's error logging — would be noise (probe is gone anyway).

@claude review

Issue #2439, opened 2026-05-07
Deprecate deleted Airtable fields: courseTable.level, unitTable.learningOutcomes, unitTable.unitPodcastUrl, unitTable.autoNumberId


Will-Howard:
### 🧩 Summary

We are getting some backend alerts like this:
```
Course Alerts  [8:03 PM]
pg-sync-service: [WEBHOOK] Removed 1 invalid field IDs from base appnJbsG1eWbAdEvf: fldc7bNIkEyrMsQ4w. These fields may have been deleted in Airtable.
[8:03 PM]pg-sync-service: [WEBHOOK] Removed 8 invalid field IDs from base appbiNKDcn1sGPGOG: fldHxekJ6BioQMF3e, fldkL7aWITGCPqzxc, fldEY7ZHZtXrBL3nv, fldF9hjDhZpLbBIUV, fldpJLWVPh0IXHfmm, fld9vAMgn0Fm7x6Xf, fldwByN7lbmcjc3Fj, fld1rl39p5fSOiFya. These fields may have been deleted in Airtable.
```

All of these fields have actually been deleted. The ones flagged in the title (`courseTable.level`, `unitTable.learningOutcomes`, `unitTable.unitPodcastUrl`, `unitTable.autoNumberId`) are genuinely unused anywhere in the code, so they just need to be deprecated + deleted from Postgres. `courseTable.durationDescription` is also safe to delete, it's only used as a fallback so can be deleted straight away.

I'll create a separate issue for the others, where I first need to investigate if/how they are used.

### 🛝 Steps to reproduce

Search "invalid field IDs" in `update_tech-prod` Slack channel

### ✅ Definition of done

These fields have been deleted from Postgres

PR #2440, opened 2026-05-07
[chore] db: deprecate fields deleted from Airtable

Will-Howard:
# Description

Move `courseTable.level`, `courseTable.durationDescription`, `unitTable.learningOutcomes`, `unitTable.unitPodcastUrl`, `unitTable.autoNumberId` to `deprecatedColumns`. These fields were already deleted from Airtable, which is why pg-sync-service has been logging "Removed N invalid field IDs..." alerts.

`durationDescription` is technically still referenced as a fallback when `durationHours` is missing, but all 3 active courses with blank `durationHours` have `displayOnCourseHubIndex=false`, and the only call site filters on that flag, so the fallback never fires today.

## Issue

#2439

## Developer checklist


apps/website/src/server/routers/courses.ts
```diff
@@ -194,19 +194,11 @@ export const coursesRouter = router({
...
-      const unit = await db.getFirst(unitTable, {
-        filter: {
-          id: unitId,
-          courseSlug,
-          unitStatus: 'Active',
-        },
+      return db.get(unitTable, {
+        id: unitId,
+        courseSlug,
+        unitStatus: 'Active',
```
Will-Howard, line 200: `getFirst` was implicitly using `autoNumberId`, which was null for almost all rows. I switched to a basic `get` because we're fetching by id anyway, which should be genuinely unique, so changing to a different `sortBy` here would just be a false sense of security

PR #2461, opened 2026-05-11
[feat] db: add priority column to job_posting table

dewierwan:
## Summary

Adds the Airtable "Prio" singleSelect field (`fld819hEjDV1DI0je`) to `jobPostingTable` so the website careers page can sort open roles by priority.

The Airtable field has three options: `1 Top`, `2 Med`, `3 Low`. The numeric prefix means a plain string sort produces the right order.

This PR is schema-only, following the [`linkPreviewImage` precedent (#2432)](https://github.com/bluedotimpact/bluedot/pull/2432) so `pg-sync-service` materialises the column in staging Postgres before the consuming website code lands.

**Follow-up PR:** website router + JobsListSection changes that consume `priority` for sorting will open as a draft against master, ready to rebase once this merges and pg-sync auto-runs a full sync.

## Test plan

- [x] `npx tsc --noEmit` in `libraries/db`, `apps/website`, `apps/editor` — all pass
- [x] `npm test` in `apps/website` — 633 tests pass
- [ ] After merge: watch `pg-sync-service` redeploy, confirm `priority` column populated in staging

PR #2462, opened 2026-05-11
[feat] website: sort careers page roles by Airtable Prio

dewierwan:
## Summary

Sorts the `/join-us` open roles list by the new Airtable "Prio" singleSelect (`1 Top` → `2 Med` → `3 Low`), with title as the tiebreaker. Roles without a Prio fall to the bottom. Applies to both the main "Open roles" list and the contractor "Support our mission" list.

Also drops the hardcoded `talent`-slug push-to-bottom in `JobsListSection.tsx` — Prio is now the single source of truth for ordering.

**Depends on #2461** (schema-only PR adding the `priority` column). This PR is draft and based on `dewi/joinus-priority-schema`; once #2461 merges and pg-sync materialises the column in staging, GitHub will auto-retarget this PR to master.

## Why split

Following the [`linkPreviewImage` precedent (#2432)](https://github.com/bluedotimpact/bluedot/pull/2432): website code that reads a not-yet-materialised column 500s on staging until pg-sync redeploys. Shipping the schema separately means staging never breaks.

## Test plan

- [x] `npx tsc --noEmit` in `libraries/db`, `apps/website`, `apps/editor` — all pass
- [x] `cd apps/website && npm test` — 633 tests pass
- [ ] After #2461 merges + pg-sync redeploys: locally hit `/join-us` and confirm order matches Airtable Prio; sweep viewports 320/480/720/1024/1440
- [ ] After merge: confirm prod ordering matches Airtable after the manual website deploy

### DIFF
--- a/apps/website/src/__tests__/pages/join-us/[slug].test.tsx
+++ b/apps/website/src/__tests__/pages/join-us/[slug].test.tsx
@@ -46,6 +46,7 @@ const mockJob: JobPosting = {
   publicationStatus: 'Published',
   category: null,
   linkPreviewImage: null,
+  priority: null,
 };
 
 describe('JobPostingPage SSR/SEO', () => {
--- a/apps/website/src/components/join-us/JobsListSection.test.tsx
+++ b/apps/website/src/components/join-us/JobsListSection.test.tsx
@@ -15,6 +15,7 @@ const mockCmsJobs = [
     publishedAt: Date.now() / 1000,
     category: null,
     linkPreviewImage: null,
+    priority: null,
   },
   {
     id: '2',
@@ -26,6 +27,7 @@ const mockCmsJobs = [
     publishedAt: Date.now() / 1000,
     category: null,
     linkPreviewImage: null,
+    priority: null,
   },
 ];
 
--- a/apps/website/src/components/join-us/JobsListSection.tsx
+++ b/apps/website/src/components/join-us/JobsListSection.tsx
@@ -7,11 +7,9 @@ import { PageListGroup, PageListRow, pageSectionHeadingClass } from '../PageList
 type JobsListSectionProps = inferRouterOutputs<AppRouter>['jobs']['getAll'];
 
 const JobsListSection = ({ jobs }: { jobs: JobsListSectionProps }) => {
-  // Split jobs into regular and contractor positions
-  // job.category is a single string value from Airtable
-  const regularJobs = jobs
-    .filter((job) => job.category !== 'Contractor')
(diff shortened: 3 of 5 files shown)


dewierwan:
@claude review

dewierwan:
Claude bot approved with one minor: explain the `'￿'` sentinel. Pushed dea3b55c — added a comment that this is U+FFFF, the max BMP code point, so null priorities sort after every real value. Left the implementation otherwise as-is.

Greptile + CodeRabbit didn't review (draft). Will flip out of draft once #2461 merges and pg-sync materialises the `priority` column on staging, so the local /join-us page can render without 500ing and I can do the viewport sweep.

dewierwan:
Marking ready for review. Rebased onto master after #2461 merged.

**Verified locally** (staging Postgres, pg-sync completed backfill):
- ✅ `npx tsc --noEmit` (libraries/db, apps/website, apps/editor)
- ✅ `npm test` — 633 pass
- ✅ tRPC `/api/trpc/jobs.getAll` returns the 9 published roles in correct order (1 Top → 2 Med → 3 Low, title-alphabetical tiebreaker within each tier)
- ✅ `/join-us` viewport sweep at 320/480/720/1024/1440 — no horizontal overflow, render order matches Airtable Prio
- ✅ Talent slug no longer pinned to bottom — "Exceptional Talent" sits last only because its Prio is `3 Low`. You can promote it any time by changing the Airtable field.

Re-triggering Claude/Greptile/CodeRabbit on the now-non-draft PR.

dewierwan:
@claude review

dewierwan:
Pass over the three reviews:

**Greptile** flagged "comparator has no unit test" — fair, addressed in 3e6451d2 with `apps/website/src/server/routers/jobs.test.ts`. The test scrambles input order across all three Prio tiers + null and asserts the exact output sequence, so a future comparator edit can't silently regress the order.

Greptile also flagged the U+FFFF sentinel as "unspecified collation behaviour." Pushing back on this one: U+FFFF is the last code point in the BMP, and `String.prototype.localeCompare` consistently orders it after `"1 Top"` / `"2 Med"` / `"3 Low"` in every ICU collation. The new unit test pins the behaviour either way, so even if some future locale config got pathological, CI would catch it.

**Claude bot** approved both passes. Its only suggestion (JSDoc on the comparator) — skipping, agree with its own caveat that the inline comments are already clear.

**CodeRabbit** rate-limited, nothing actionable.

dewierwan:
@claude review

PR #2512, opened 2026-05-20
[fix] CTG: sum all approved grants + correct check-in cadence

anglilian:
## Summary
- Removes the redundant `evaluationStatus === 'Agreement signed'` filter from `getCareerTransitionGrantStats`. The Airtable source view is already pre-filtered to Approved + Agreement signed, so the website should trust the sync — this makes the stats include the 3 currently-Approved grants too. **Stat changes: $481,000 → $591,000.**
- Moves `evaluationStatus` to `deprecatedColumns` since nothing reads it anymore. A PR 3 will delete it entirely once this ships to production.
- Corrects the CTG page check-in cadence copy: actual cadence is monthly, not quarterly. Updates the chip from `Quarterly` to `Monthly` and the body from "Every three months" to "Every month".

## Test plan
- [x] `npm run typecheck` (clean)
- [x] `npm run lint` (clean)
- [x] `npx vitest run` — 679 tests pass, twice in a row (added 2 new tests for `getCareerTransitionGrantStats`)

## Follow-up
- PR 3 (after this ships to production): delete `evaluationStatus` from `deprecatedColumns` entirely, per `@bluedot/db` 2-PR removal workflow.

### DIFF
--- a/apps/website/src/components/career-transition-grant/ExpectationsSection.tsx
+++ b/apps/website/src/components/career-transition-grant/ExpectationsSection.tsx
@@ -13,9 +13,9 @@ const EXPECTATIONS = [
     body: 'Short async updates to your BlueDot point of contact on what you did, who you talked to, what you learned, and how your thinking is evolving.',
   },
   {
-    cadence: 'Quarterly',
+    cadence: 'Monthly',
     title: 'Check-in',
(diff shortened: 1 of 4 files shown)


anglilian:
@claude review

PR #2511, opened 2026-05-19
[fix] Point CTG schema at the new synced tables

anglilian:
## Summary
- Re-target `careerTransitionGrantTable` to the live `firstName` / `lastName` / `profileUrl` field IDs in the Career transition grantees table after the CTG data migration. The previous IDs no longer exist in Airtable, so the GranteesSection on `/programs/career-transition-grant` has been rendering empty cards.
- Re-target `careerTransitionGrantApplicationTable` from the now-emptied `tblTBVFKyJNlvhfA6` (old CTG table) to the new synced `tblYkZrxApKMUe7uf` (pre-filtered to Approved + Agreement signed).

## Test plan
- [x] `npm run typecheck` clean (website)
- [x] `npm run lint` clean
- [x] `npx vitest run` — 677 tests pass, twice in a row (111 test files)
- [x] `libraries/db` tests pass (22 tests)

## Follow-up
This is a schema-only PR per `@bluedot/db` schema-first workflow. Once pg-sync materialises the new data, a follow-up will:
- Remove the now-redundant `evaluationStatus === 'Agreement signed'` filter from `getCareerTransitionGrantStats` (sync handles it)
- Move `evaluationStatus` to `deprecatedColumns`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2513, opened 2026-05-20
[fix] Hotfix: hardcode CTG stats for orphaned pg-sync rows

anglilian:
## Summary
**Production is broken** — `/programs/career-transition-grant` shows `Grants made: 164` / `Funding awarded: \$8,101,032` instead of the correct `10` / `\$591,000`.

**Root cause:** PR #2511 re-pointed `careerTransitionGrantApplicationTable` to the new synced view `tblYkZrxApKMUe7uf`, but pg-sync has no orphan-row cleanup when a `tableId` changes. The Postgres `career_transition_grant_application` table still holds ~153 stale rows from the old master CTG (`tblTBVFKyJNlvhfA6`, since deleted in Airtable) alongside the 10 fresh synced rows. PR #2512's filter-free aggregate then sums all 164 rows.

**Stopgap:** Hardcode the stats in `grants.ts`. Site immediately shows the correct number on the next release. Constant must be updated manually when new CTGs are awarded.

**Proper follow-up (separate PR):** Rename the pg-airtable identifier from `career_transition_grant_application` to `…_v2` so pg-sync creates a fresh table with no orphans; then revert this hardcoding.

## Test plan
- [x] `npm run lint` clean
- [x] `npx vitest run` — 678 tests pass (1 fewer than #2512: collapsed two data-driven tests into one hardcoded check)

### DIFF
--- a/apps/website/src/server/routers/grants.test.ts
+++ b/apps/website/src/server/routers/grants.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'vitest';
-import { careerTransitionGrantApplicationTable, rapidGrantTable } from '@bluedot/db';
+import { rapidGrantTable } from '@bluedot/db';
 import { createCaller, setupTestDb, testDb } from '../../__tests__/dbTestUtils';
 
 setupTestDb();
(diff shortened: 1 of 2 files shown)


anglilian:
@claude review

PR #2522, opened 2026-05-20
[chore] db: add decidedAt to rapidGrantApplicationTable

dewierwan:
## Summary

- Adds `decidedAt` (Airtable `lastModifiedTime` watching `Grant decision`, field id `fldYSUC9oNUMrRiKr`) to the `rapid_grant_application` Postgres sync.
- Unlocks a follow-up PR that will compute live decision-time stats for the `/programs/rapid-grants` marketing page (currently hardcoded as `~5 working days`; real median is ~14h).
- `pg-sync-service` auto-detects the schema diff at startup and backfills — no manual trigger needed.

## Test plan

- [x] `npm test` in `libraries/db` (22/22 pass)
- [x] `npx tsc --noEmit` in `libraries/db` (clean)
- [x] `npm run lint` in `libraries/db` (clean)
- [ ] Confirm `decided_at` column populated in staging after merge before opening the consumer PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2520, opened 2026-05-20
[feat] CTG page: 3 live stats (grants made, funding awarded, avg days to decision)

anglilian:
## Summary
Un-hardcodes `getCareerTransitionGrantStats` now that pg has the master CTG data correctly synced (PRs #2516–#2518). Reads from `careerTransitionGrantApplicationTable`, filters count + funding-awarded to Approved + Agreement signed records, and averages `timeToDecisionDays` over all decided rows.

Adds a third item — **"Avg days to decision"** — to the `GrantStatsStrip` on `/programs/career-transition-grant`, rendered as `~N` to signal an approximate average.

`GrantStats` is extended via a new `CareerTransitionGrantStats` subtype so the rapid-grants endpoint isn't forced to carry the avg-days field.

## Data flow
- pg-sync materialises `career_transition_grant_application` from the master CTG table (`tblh5zr4jRdrndKnC`) — ~162 rows of all statuses.
- `getCareerTransitionGrantStats` runs in JS:
  - `count` + `totalAmountUsd`: filter to `status IN ['Approved', 'Agreement signed']`
  - `averageDaysToDecision`: average `timeToDecisionDays` over rows where it's > 0 (excludes TODO/unscheduled rows and same-day decisions); `null` if no decided rows

## Test plan
- [x] `npm run typecheck` clean
- [x] `npm run lint` clean
- [x] `npx vitest run` — 680 tests pass (+1 test file new coverage for filter and avg behaviour)

## Verification before release
Staging will auto-deploy on merge. Eyeball https://website-staging.k8s.bluedot.org/programs/career-transition-grant — expecting **3 stats**, with values pulled live from pg (likely Grants made ≈ 10-12, Funding awarded ≈ $591–691k, Avg days to decision ≈ small single-digit number `~N`).

Once staging looks right, cut `website/v2.26.39` to ship to production.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2524, opened 2026-05-21
[feat] rapid-grants page: live avg decision time + p90 tail

dewierwan:
## Summary

- Stats strip on `/programs/rapid-grants` now shows a **live trimmed-mean decision time** instead of the hardcoded `~5 working days`. Label renamed to "Avg decision time".
- "How it works → Get a decision" body now reads "On average we reply in N days, and 9 in 10 applicants hear back within a week" — the p90 tail sets honest expectations for accepts (which run longer than rejects).
- 10% trim drops the fastest and slowest 10% of decided rows, so one 60-day outlier can't dominate the headline. For small samples (n < 10) the trim degrades gracefully to the arithmetic mean.
- Mirrors the live-stats pattern shipped for the CTG page in #2520; reads from `rapid_grant_application.decidedAt` synced via #2522.

## Test plan

- [x] `npx vitest run src/server/routers/grants.test.ts` — 8/8 pass (5 new tests covering accept-count + trim behaviour + null/empty cases, 3 existing CTG tests preserved)
- [x] Full `npm test` in `apps/website` — 683 pass + 1 unrelated skipped
- [x] `npx tsc --noEmit` clean
- [x] `npm run lint` clean
- [x] Viewport sweep at 320/480/600/720/1024/1440 — no horizontal overflow, decision-card copy wraps cleanly at every width
- [x] Visually verified both changes at mobile (iPhone 14) + desktop (1280 DPR=2) — screenshots below

### Stats strip

|            | Before                                                                                              | After                                                                                              |
| ---------- | --------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
| 📱 Mobile  |   |   |
| 🖥️ Desktop |   |   |

### How it works → Get a decision

|            | Before                                                                                              | After                                                                                              |
| ---------- | --------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
| 📱 Mobile  |   |   |
| 🖥️ Desktop |   |   |


dewierwan:
@claude review

dewierwan:
Thanks for the review.

**Findings addressed:** none — all three are defensive nice-to-haves rather than functional issues, and I think each makes the code worse on net:

1. **Division-by-zero in `trimmedMean`** — the only call site already guards with `decisionHours.length ? trimmedMean(...) : null`, so the helper is unreachable on an empty array. Returning `0` from "mean of nothing" would be misleading; NaN is more honest.
2. **Magic-number program-launch date** — pre-dates this PR (it was already hardcoded in `getRapidGrantStats`). Out of scope here.
3. **Server-side caching** — n≈1000, the table doesn't grow fast, and tRPC + React Query already caches on the client. Premature for this scale.

Greptile + CodeRabbit haven't landed yet — if either flags something material I'll address it then.

dewierwan:
Thanks Greptile — also non-blocking, also skipping both with reasoning:

1. **`trimmedMean` NaN footgun** — same as Claude bot's flag. Call site guards (`decisionHours.length ? trimmedMean(...) : null`), so the helper is unreachable on an empty array. Greptile itself notes "cannot happen at trimPct=0.1." Defensive `if (inner.length === 0)` would force a misleading return value (NaN is more honest than 0).
2. **Fallback copy inconsistency between strip and body during loading** — intentional graceful degradation. The strip's em-dash matches the existing convention for `grantsMadeLabel` / `fundingGivenOutLabel`; the body's fallback sentence is conservative and matches what live values produce in practice. Acceptable for the ~100ms loading window and the (rare) DB-unreachable case.

Ready to merge.

Issue #2493, opened 2026-05-14
Scope proposal to move resource completions and exercise responses into Postgres-only


Will-Howard:
### 🧩 Summary

Following on from #2418: These tables will push us into the hard limit of rows in Airtable within ~2 months, so we need to come up with a permanent solution for moving them out of Airtable.

Currently, our Postgres instance is within the same node as the server, and is treated as ephemeral due to being a pure replica of Airtable (no backups, and gets re-synced from scratch frequently). Moving these tables out will require solving:
- Making Postgres durable to normal production-app standards
- Any issues with cross-database joins and consistency within the code
- Replacing any uses of these tables within Airtable automations, interfaces, workflows of the BlueDot team
- ...possibly more

### ✅ Definition of done

The engineers and BlueDot team are aligned on a proposal to move these tables, and we are confident we will complete it before we run into the hard limits

Will-Howard:
# Next steps for investigating

There are two main issues to figure out:
1. Our current postgres instance is within-node, which makes the single query latency very fast (microseconds vs milliseconds for a same-region standalone instance). The simplest way to get a more durable instance is to use a standalone provider (Supabase or Neon), which would be outside-node. This will increase latency, which could be a small problem for the website itself, and could be a _big_ problem for pg-sync-service; if, for instance, it makes a full sync take forever, or generally causes race condition issues by making the queue pile up
2. There are some lookup/aggregate fields derived from exercise responses and resource completions. We'll need to either drop these or work out a way to push them back into Airtable

Apart from these issues, it appears that we can just flip the tables from `pgAirtable` to `pgTable` and things will keep working. On (2), none of these aggregate fields are actually used in code, so handling them isn't an issue (purely an admin-UI and airtable-automation issue).

## 1. Latency issue

- [x] Test the latency hypothesis. In both the plain website and pg-sync-service
  - [x] Website: Sufficient to test on a Render preview environment, since this is already far from the db but within-Europe
  - [x] pg-sync-service: IMO we should deploy a second "shadow" sync service, make it point to an out-of-node postgres, and measure how it behaves in terms of full-sync time and issues with the queue piling up
  - [x] It's possible the _interaction_ of the two will be an issue, but I only expect this to be a problem if pg-sync-service itself handles it badly. So I think there's no need to investigate this per se unless we discover an additional reason to think the interaction might be a problem
- If both are fine, just set up a plain out-of-node postgres instance (Neon or Supabase) and flip the tables to pgTable
[...]

PR #2533, opened 2026-05-22
[chore] db: deprecate unitResource.avgRating

Will-Howard:
# Description

The `Avg rating` field (`fldOWWeymJQTwlfaY`) on the Course Builder "Chunk resource" table (`unitResourceTable` in code) was deleted from Airtable by Dewi during cleanup for #2493, so it no longer receives sync updates.

This moves `avgRating` from `columns` to `deprecatedColumns` in `libraries/db/src/schema.ts`. Per the two-PR column-removal process in the handbook, this:

- keeps the existing Postgres column (via `pgWithDeprecatedColumns`) so no data is dropped,
- stops `pg-sync-service` trying to read a field that no longer exists in Airtable,
- removes it from the `UnitResource` type, so the now-dead `avgRating: null` value in the `createMockResource` test util is dropped too.

The field was never read in application code (only the schema entry + the test mock referenced it), so there's no consumer change.

Full removal from the schema is a **follow-up PR** once this has merged and deployed to production.

## Issue

Related to #2493

## Developer checklist

- [x] Considered having snapshot tests and/or happy path functional tests (db tests pass 22/22, incl. `validate-schema`; website typecheck clean)
- [ ] Front-end code follows Component Accessibility Checklist — n/a
- [ ] Considered adding Storybook stories — n/a

## Screenshot

n/a — schema-only change.

### DIFF
--- a/apps/website/src/__tests__/testUtils.ts
+++ b/apps/website/src/__tests__/testUtils.ts
@@ -266,7 +266,6 @@ export const createMockResource = (overrides: Partial<UnitResource> = {}): UnitR
   coreFurtherMaybe: 'Core',
   readingOrder: null,
   unitId: 'unit-1',
-  avgRating: null,
   syncedAudioUrl: null,
   year: 2024,
   autoNumberId: 1,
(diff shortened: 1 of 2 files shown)


Will-Howard:
@claude review

PR #2413, opened 2026-05-05
settings/courses to My Courses migration: Participant view

Will-Howard:
# Description

_Issue #2400 is to implement [this fairly big redesign of the `settings/courses` page](https://www.figma.com/design/CYf4tCWNcbhh9cYIMhTKtL/Settings-courses?node-id=179-87548&m=dev), splitting it into "My Courses" (participant view), and "Facilitated Courses" (facilitator view)._

_Plan for review + deploy: As this is a complete redesign, I've decided to aim to merge this ~all-in-one-go into master, leaning on [preview environments](https://bluedot-preview-pr-2413.onrender.com/) to do detailed review. This PR covers the whole participant view, I plan to make one other PR against this one for the facilitator view (which will be a lot smaller). For debugging, this PR also has the legacy version of the page under "[Courses (legacy)](https://bluedot-preview-pr-2413.onrender.com/legacy/settings/courses)" in the menu, I'll remove this before merging._

Areas for reviewers to focus:
- States of CourseListRow (see [story](https://bluedot-storybook-preview-pr-2413.onrender.com/?path=/docs/website-my-courses-courselistrow--docs)): There are various rules that we have deliberately added over time (e.g. show "Submit feedback" before allowing them to view their cert), which need to render nicely in the new version
- States of DiscussionListRow (see [story](https://bluedot-storybook-preview-pr-2413.onrender.com/?path=/story/website-my-courses-discussionlistrow--all-states)): As above
- States of NextDiscussionCard (see [story](https://bluedot-storybook-preview-pr-2413.onrender.com/?path=/docs/website-my-courses-nextdiscussioncard--docs)): As above
- Empty states

## Issue

#2400

## Developer checklist

- [x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist)
- [x] Considered having snapshot tests and/or happy path functional tests
- [x] Considered adding Storybook stories

## Screenshot

_Due to the complexity, I suggest relying on the [preview environment](https://bluedot-preview-pr-2413.onrender.com/) rather than screenshots for review._

| 📸 | Before | After |
|---------|---|---|
| 📱  | Screenshot 2026-05-19 at 11 25 21 | Screenshot 2026-05-19 at 12 06 41 |
| 🖥️ | Screenshot 2026-05-08 at 18 00 46 | Screenshot 2026-05-19 at 12 06 25 |

PR #2541, opened 2026-05-22
CLAUDE.md proposal: allow named palette utilities (e.g. text-bluedot-navy), preserve the ban on raw hex

Will-Howard:
# Description

**Proposal, opening for discussion (also raising in Slack).**

Our docs currently steer assistants/humans toward semantic colour tokens (`text-color-secondary-text`, `bg-color-primary-accent`) and frame the named `bluedot-*` palette utilities, e.g. `bg-bluedot-normal`, `text-bluedot-navy/62`, as "raw values to avoid".

In practice the `bluedot-*` utilities are used pervasively (~109 files in `apps/website`). I also personally find them more dev-friendly, because I can keep in my mind the main colours (`bluedot-normal` and `bluedot-navy`) and picture gradations like text-bluedot-navy/60, whereas I find it hard to remember `bluedot-secondary`, `color-divider` etc.

This PR reframes the docs so:
- The real enemy is raw hex values, and excessive use of raw pixel values
- Named palette utilities (`text-bluedot-navy`, incl. opacity like `/60`) and semantic tokens (`bg-color-primary-accent`, `border-color-divider`) are **both acceptable**

The immediate motivation for this was getting a bunch of misfiring "use design token instead of raw colour class" comments on #2413. The main effect of this change will be to stop AI assistants from nudging in this semantic-token direction.

**The case against**

I think there are reasonable arguments against this:
- Under a strict design system (with rigidly followed rule _and_ stock components), semantic tokens help to uphold the strictness
- If we ever support dark mode, it may make it somewhat easier to define non-exact inverses of semantic colour tokens (though, this is a long way off, and I'm not sure if it would actually be easier because it just creates more tokens).

## Issue

N/A

## Developer checklist

N/A

## Screenshot

N/A

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,7 +31,7 @@ We're fans of [boring technology](https://boringtechnology.club/) — don't intr
 ## Before you start a task
 
 - Look for existing components/utilities first. Most marketing primitives already live in `@bluedot/ui` or `apps/website/src/components/`. See `apps/website/README.md` → "Reusing existing components first".
-- Use design tokens (`text-size-md`, `bg-color-primary-accent`, `border-color-divider`, …), not raw Tailwind values like `text-[16px]` or `bg-bluedot-normal`. Tokens are documented in `apps/storybook/src/GettingStarted.mdx` and defined in `apps/website/src/globals.css`.
+- Use the project's named design tokens / utilities (`text-size-md`, `text-bluedot-navy`, `bg-color-primary-accent`, `border-color-divider`, …), not raw or arbitrary values like `text-[#0037ff]`. Exact pixel values like `gap-[3px]` are sometimes ok. Tokens are documented in `apps/storybook/src/GettingStarted.mdx` and defined in `apps/website/src/globals.css`.
 - Check existing components in Storybook (`apps/storybook/` or [storybook.k8s.bluedot.org](https://storybook.k8s.bluedot.org)) before building new ones. If a new component would benefit from visual documentation (e.g. for design review or reuse by non-technical team members), ask the user if they'd like a Storybook story.
 - For non-trivial work, sketch a plan before editing.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,7 +31,7 @@ We're fans of [boring technology](https://boringtechnology.club/) — don't intr
 ## Before you start a task
 
 - Look for existing components/utilities first. Most marketing primitives already live in `@bluedot/ui` or `apps/website/src/components/`. See `apps/website/README.md` → "Reusing existing components first".
-- Use the project's named design tokens / utilities (`text-size-md`, `text-bluedot-navy`, `bg-color-primary-accent`, `border-color-divider`, …), not raw or arbitrary values like `text-[#0037ff]`. Exact pixel values like `gap-[3px]` are sometimes ok. Tokens are documented in `apps/storybook/src/GettingStarted.mdx` and defined in `apps/website/src/globals.css`.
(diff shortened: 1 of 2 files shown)


joshestein:
I definitely prefer `text-bluedot-navy/60` to `text-color-secondary-text`, +1 on the proposal

marn-in-prod:
I think this PR should ban semantic colour tokens and say to use the bluedot palette utilities instead. We should still ban raw hex values.

I'd actually want to make this more strict than what you're suggesting, saying both are fine will just lead to larger inconsistency. I took a look at the numbers and we use the bluedot utilities like 12x more than the semantic tokens, and I agree that they're easier to read IMO. Since we already have so many more `bluedot-*` usages, we should just enforce those and ban the semantic colour tokens.

Will-Howard:
> I think this PR should ban semantic colour tokens and say to use the bluedot palette utilities instead

👍 seems reasonable. I've updated it and I'll merge this now

cyrus-hhc:
Quick follow-up on the colour-token direction -- one case for keeping a thin semantic layer on top of the bluedot-* palette, then I'll defer to where the team lands. 🙏

I'm not arguing to replace bluedot-* or revive the old color-* tokens. Keep the palette, keep the raw-hex ban. The old semantic layer wasn't working and Martin's right that "both are fine, use whichever" just breeds inconsistency. I'm proposing something narrower.

# What
We will have a two layers
- primitive: This is the colour palette set we reference for different components. 
- semantic: This is what we will use and always references primitive.
image


# Why 
The core point — a palette class (bluedot-*) names a colour; a semantic token names a decision.

text-bluedot-navy/70 tells you what the colour is (navy, 70% opacity). It tells you nothing about what it's for or whether 70 was deliberate. So the next person can't reuse that decision — because the name doesn't contain one. They just make their own: /60, /65, /72. Every one of them looks reasonable, every one is "correct," and that's exactly how we ended up with ~18 different navy opacities in the codebase.

text-secondary is different in kind. It doesn't describe a colour — it points at a decision the team made once: "this is our supporting-text style, and it has one blessed value." When three of us need dimmed body text:

- Palette world: I type /70, you type /60, someone types /65 — three greys ship, all defensible.
- Semantic world: all three of us type text-secondary and get the identical pixel. The name forced the agreement.

That's the crux: with the palette, consistency depends on everyone remembering the right number every single time — consistency by discipline. With a semantic token, consistency is structural — there's one value behind the name, so you can't drift even if you try. Palette-only doesn't deliver the consistency we want; it just relocates the inconsistency from hex codes to opacity values.


(Dark mode also gets much easier later — but that's far off and not my main argument.)

If the team still prefers palette-only, genuinely fine — but then I'd push that we at least lock the opacity steps to a fixed set and make it clear in the documentation that when we would use a particular opacity step, since the open /NN knob is the real inconsistency, whatever we name things.

marn-in-prod:
Hi @cyrus-hhc , I think there's a misunderstanding from looking at the PR comment. It's important to note that the PR discussion happened before we intended to have a UXer do a design system. I linked to it because it had more examples of the bluedot naming convention.

Quick points so that we can all get aligned:
- With the new design system, we don't want to keep using different opacities. The hope is that there are enough semantic tokens + descriptions that we can always use something specific.
- We should definitely have both Primitives and semantic tokens that exist, and components should be using semantic tokens.
- Me saying "I think this PR should ban semantic colour tokens and say to use the bluedot palette utilities instead." was because we weren't going to have enough semantic tokens to cover most situations without the engineers creating a design system, so it was the correct decision at the time, but not for after a design system is in place that was created by a UXer.
- As much as possible, the semantic tokens should be plentiful enough (or at least have enough "use for" descriptions) so that it's clear when to use which. I.e. try not to use something like "text/secondary" or "text/primary", because it very quickly devolves into questions like: do placeholders, captions, disabled text, and helper text all count as "secondary", or should I be using something else instead? Or if you are going to use them, then the decision process for choosing should be very clear and repeatable across people (and AI).

Hopefully that mostly clears things up!

Issue #2543, opened 2026-05-22
Moving resource/exercise responses into Postgres-only: Scope pushing aggregate fields back into Airtable


Will-Howard:
### 🧩 Summary

Things we want to be pushed back into Airtable:
- Resource table: avg rating, num completions
- Exercise table: num completions

Explicitly out of scope: Providing a simple way to look at individual answers in the database (by exercise, by resource, or by user). We likely will want this eventually, but not as a blocker to switching over. We also used to have num completion/avg rating fields spread across resource/exercise/resource completion/user. Again we may want these again in future, but it's sufficient to just cover Resource and Exercise to establish the pattern, and extend to more in future.

The specific task of this issue is: We will need to push these values back in some concrete way, via a cron job and/or live updates when resource/exercise completions are written. We should plan how to do this in a fairly generic way, to support these and future postgres-to-airtable aggregate fields

### ✅ Definition of done

We have a plan for pushing these values back into Airtable (a new Issue to implement it)

Will-Howard:
# Proposal using `.invalidate` semantics, similar to tRPC

## Framing the problem

1. Denormalising fields generally has some pitfalls:
  a. Data can get out of sync
  b. It can add bloat at write sites if you have to do manual logic to incrementally update the normalised fields
  c. It can thrash the database (e.g. if you are overzealous on recalculating _every_ value on every write)
  d. ...many more besides these
2. In my experience, some good ingredients in any solution to denormalisation (or any kind of non-source-of-truth data store) are:
  a. Make the rule for deriving the denormalised data _declarative_, e.g. `numCompletions = count(completions)` rather than `numCompletions = prevNumCompletions + 1`. This makes it so recalculation always clears any existing data issue by rebuilding from scratch
  b. Have a cron job that periodically refreshes the data so that any discrepancies eventually get back in sync
3. The remaining problem is choosing how to trigger recalculation. I.e. whether to do it manually or implicitly, and balancing being more eager to avoid out-of-sync issues, vs being more conservative to avoid thrashing the database.

On the "declarative" point, this is what the Airtable Lookup/Rollup fields we're replacing already do.

## Proposed solution

On how to trigger recalculation (point 3 above), this is my proposal to achieve a good middle ground:
- Manually trigger recalculation on the server using an `.invalidate()` function, which follows the same semantics as the `tRPC/react-query` invalidate function. For the current fields, we'll need to call this whenever someone edits an exercise or ticks a resource
[...]

PR #2550, opened 2026-05-26
Demo: Mini rollup library for pushing fields back into Airtable

Will-Howard:
# Description

TODO intended as a proof-of-concept, outline usage

## Issue

#2543


apps/pg-sync-service/src/lib/rollup-sync.ts
```diff
@@ -0,0 +1,6 @@
+import { bindRollups, rollupDefinitions } from '@bluedot/rollups';
+import { db } from './db';
+
+const rollups = bindRollups(db, rollupDefinitions);
```
Will-Howard, line 4: Note: I don't like this `bindRollups` function, this is something I would try to remove in a non-demo version

apps/pg-sync-service/src/lib/rollup-sync.ts
```diff
@@ -0,0 +1,6 @@
...
+
+export const recomputeRollups = () => rollups.invalidate();
```
Will-Howard, line 6: Starting from scratch with the current fields we need, this would do ~thousands of writes to Airtable, taking 10-30 mins, so similar to a "full sync" of postgres. Once the fields are populated, each incremental run it would be much faster because for most rows the Airtable write would be skipped because the field already contains the correct value

Will-Howard:
@coderabbitai @greptileai review

Will-Howard:
Closing this as it has done its demo-ing work. I'll re-create the real version using parts of this

Issue #2535, opened 2026-05-22
Deprecate `unitTable.content`


joshestein:
### 🧩 Summary

Looks like this field was deleted from Airtable 😮‍💨:

pg-sync-service: Airtable validation warning encountered, attempting to proceed by setting the affected fields to undefined. Warning message: Failed to map record from Airtable format for table 'unit' (tblsDKJ8VCyO619nk) and record rec0eWIhAXVcBSyBY: Field 'fldF9hjDhZpLbBIUV' does not exist in the table definition. This error should not happen in normal operation. (Table: tblsDKJ8VCyO619nk) (Field: fldF9hjDhZpLbBIUV)

### 🛝 Steps to reproduce

View prod logs

### ✅ Definition of done

Field is deprecated then removed from schema. Update usages accordingly

PR #2552, opened 2026-05-26
[refactor] Deprecate `unitTable.content`

joshestein:
Closes part of #2535.

## User-visible impact

Course unit pages render the same. `unit.content` Markdown field was previously a fallback when a chunk had no content of its own; that fallback has been dead since the Airtable `Content` field was deleted upstream, so removing it is a no-op for users and silences a recurring pg-sync-service warning in production logs.

## Summary

- Drop `chunk?.chunkContent || unit.content` fallback in `UnitLayout.tsx`. Render only `chunk.chunkContent`.
- Move `unitTable.content` to `deprecatedColumns` in `libraries/db/src/schema.ts`. Already nullable, no constraint change.
- Strip `content:` from test/story fixtures and from `createMockUnit` in `testUtils.ts`.

## Follow-up

Per 2-PR deprecation workflow: after this lands and ships to **production** (`website/vX.Y.Z` tag), a follow-up PR deletes `content` from `deprecatedColumns` and drops the postgres column.

## Test plan

- [x] `npm test` in `apps/website` (one unrelated FreeTextResponse timer flake; passes on rerun)
- [x] `npx tsc --noEmit` in `apps/website`
- [x] `npm run lint` in `apps/website`
- [ ] Manual: load a course unit page in staging, confirm chunk content still renders

### DIFF
--- a/apps/website/src/__tests__/pages/courses/[courseSlug]/units/[unitNumber].test.tsx
+++ b/apps/website/src/__tests__/pages/courses/[courseSlug]/units/[unitNumber].test.tsx
@@ -65,10 +65,10 @@ describe('CourseUnitPage', () => {
     } as unknown as NextRouter);
 
     const mockUnits = [
-      createMockUnit({ unitNumber: '0', title: 'Icebreaker', content: 'Welcome to the course' }),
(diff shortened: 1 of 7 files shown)


joshestein:
@claude review

Issue #2544, opened 2026-05-22
Moving resource/exercise responses into Postgres-only: Measure latency effect of outside-node postgres on website, pg-sync-service


Will-Howard:
### 🧩 Summary

Per [this comment](https://github.com/bluedotimpact/bluedot/issues/2493#issuecomment-4509195785):

> Our current postgres instance is within-node, which makes the single query latency very fast (microseconds vs milliseconds for a same-region standalone instance). The simplest way to get a more durable instance is to use a standalone provider (Supabase or Neon), which would be outside-node. This will increase latency, which could be a small problem for the website itself, and could be a big problem for pg-sync-service; if, for instance, it makes a full sync take forever, or generally causes race condition issues by making the queue pile up
>
> ...
> - [x] Test the latency hypothesis. In both the plain website and pg-sync-service
>   - [x] Website: Sufficient to test on a Render preview environment, since this is already far from the db but within-Europe
>   - [x] pg-sync-service: IMO we should deploy a second "shadow" sync service, make it point to an out-of-node postgres, and measure how it behaves in terms of full-sync time and issues with the queue piling up
>   - [x] It's possible the _interaction_ of the two will be an issue, but I only expect this to be a problem if pg-sync-service itself handles it badly. So I think there's no need to investigate this per se unless we discover an additional reason to think the interaction might be a problem

### ✅ Definition of done

1. We have a baseline estimate of the effect of moving to outside-node postgres
2. We have established metrics for the behaviour of pg-sync-service which can be used as a guide to refactoring (e.g. something like time-to-full-sync, which we can estimate locally and try to bring down through refactoring)

Will-Howard:
**Results for the website**

I realised that I could just "use AI" to exactly analyse the time-complexity of every tRPC endpoint, so I did that rather than doing a live measurement. Results assuming going from <1ms/query round-trip-time to ~10ms (very pessimistic, within-region could keep it in single digit ms):
- The ~worst case scenarios are around 6-8 sequential queries (`facilitators.getFeedbackFormData`, `courses.getCourseProgress`, `exercises.getGroupExerciseResponses` etc). This is fine, it will add at most ~80ms, and likely more like ~40ms
- There are no real N+1 issues

This accords with the Render preview environments generally being fast. So, as expected, it seems the website is safe and `pg-sync-service` is the likely problem.

Will-Howard:
# Results for pg-sync-service

[...]

PR #2571, opened 2026-05-28
[refactor] db: drop redundant transaction wrap in ensureReplicated

Will-Howard:
# Description

The `pgUnrestricted.transaction(...)` wrap in `ensureReplicated` was redundant: each branch (delete, or upsert) executes exactly one SQL statement. The explicit BEGIN/COMMIT added 2 PG round trips per row with no atomicity benefit.

Discovered as part of [#2544](https://github.com/bluedotimpact/bluedot/issues/2544#issuecomment-4564020063)

## Issue

Related to #2544.

## Developer checklist

### DIFF
--- a/libraries/db/src/lib/client.ts
+++ b/libraries/db/src/lib/client.ts
@@ -464,37 +464,34 @@ export class PgAirtableDb {
     /** Must be passed explicitly for a record to be deleted. */
     isDelete?: boolean;
   }): Promise<BasePgTableType<TTableName, TColumnsMap>['$inferSelect'] | undefined> {
-    return this.pgUnrestricted.transaction(async (tx) => {
-      if (isDelete) {
-        const deletedResults = await tx.delete(table.pg).where(eq(table.pg.id, id)).returning();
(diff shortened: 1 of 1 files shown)


libraries/db/src/lib/client.ts
```diff
@@ -464,37 +464,34 @@ export class PgAirtableDb {
...
-      const data = fullData ?? await this.airtableClient.get(table.airtable, id);
+      return deletedResult;
+    }
 
-      if (!data) {
-        throw new Error('No data found for upsert operation');
-      }
+    const data = fullData ?? await this.airtableClient.get(table.airtable, id);
 
-      // TODO For updates, check if there are changes with SELECT first to avoid acquiring a lock
```
Will-Howard, line 485: This comment was misguided. A `SELECT` pre-check without `SELECT FOR UPDATE` would introduce a check-then-act race condition. If we did do `SELECT FOR UPDATE` then that would still introduce a lock, so overall the `onConflictDoUpdate` is below is already the best we can do in terms of minimum queries/minimum locking

joshestein (approved):
LGTM 💃

PR #2604, opened 2026-06-02
[refactor] Move availability utils + offset into @bluedot/utils

### LINKED ISSUE #2603: Move `weekly-availabilities` utils + offset into `@bluedot/utils`
### 🧩 Summary

The new 'quick apply' makes use of the extracted time availability grid (`TimeAvailabilityGrid`, which lives in `@bluedot/ui`).

It also needs the pure conversion/offset logic: `weeklyTimeAvToIntervals`, `intervalsToWeeklyTimeAv`, `shiftIntervals`, the timezone `offsets` list + parse/format, etc.

This logic is **framework-agnostic domain code**, not UI. It should live in `@bluedot/utils` (React-free, already depended on by `@bluedot/ui`, `apps/availability`, and `apps/website` incl. server) rather than in `@bluedot/ui`. That way the quick-apply **server-side** mutation can import it without pulling in the React UI library.

Split:
- Pure logic (conversions, helpers, offsets, `MINUTES_IN_UNIT`/`MINUTES_IN_WEEK`, `TimeAvailabilityMap` type) → `@bluedot/utils`.
- `TimeAvailabilityGrid` component stays in `@bluedot/ui`, importing the shared const/type from `@bluedot/utils`.

> Note: originally scoped as "into `libraries/ui`"; changed to `@bluedot/utils` for the layering reason above.

### ✅ Definition of done

Two encapsulating helpers:

- `gridToUtcIntervalString(map, timezone): string`
- `utcIntervalStringToGrid(utc, timezone): TimeAvailabilityMap`

are exported from `@bluedot/utils` and re-used by both `apps/availability` and `apps/website` (after quick-apply implemented).

### 📌 Notes / Designs

_No response_

joshestein:
Relocate the framework-agnostic weekly-availabilities conversion logic and timezone offset helpers out of apps/availability into @bluedot/utils, so server-side code can consume them without pulling in React. Adds `gridToUtcIntervalString` / `utcIntervalStringToGrid` helpers. TimeAvailabilityGrid stays in @bluedot/ui, now importing MINUTES_IN_UNIT + TimeAvailabilityMap from @bluedot/utils.

Closes #2603


joshestein:
Thanks bots. Addressed in 0363b8b.

**Greptile P2 (internal signatures should use `TimeAvailabilityMap`)** — applied. `weeklyTimeAvToIntervals` now takes `TimeAvailabilityMap` and `intervalsToWeeklyTimeAv` returns it, which let me drop every `as Record<wa.WeeklyTime, boolean>` cast (in `intervalsToWeeklyTimeAv`, `gridToUtcIntervalString`, and the call site).

**Greptile P1 (missing cast in `longEnoughInterval`)** — this isn't actually a compile error: `wa.WeeklyTime` is `Brand<number, 'WeeklyTime'>`, so `Record<number, boolean>` is assignable to `Record<wa.WeeklyTime, boolean>` (the branded-number key collapses to a `number` index signature). Confirmed with `tsc --noEmit` on both `@bluedot/utils` and `apps/availability` (exit 0, before and after). The P2 fix removes the inconsistency anyway — no call site casts now.

**CodeRabbit (subpath exports for the `@bluedot/utils` barrel)** — declining. Bundlers tree-shake unused exports, and `validateEnv` / `slackAlert` aren't referenced from `TimeAvailabilityGrid` so they won't land in the client bundle. CodeRabbit's own analysis confirms both are client-safe regardless. Subpath exports would be a broader package restructure, out of scope here.

Typecheck + the 7 `timeAvailability` tests pass after the change.

Issue #2582, opened 2026-05-31
Moving resource/exercise responses into Postgres-only: Create framework for pushing aggregate fields back into Airtable


Will-Howard:
### 🧩 Summary

Implementation of #2543 (and #2550, demo version): Create a new library for defining aggregate fields, and add a cron in `pg-sync-service` to keep these up to date

### ✅ Definition of done

1. The above library is implemented
2. We have this set up to populate `numCompletions` for resource/exercise responses (populated back onto Resource and Exercise), and `averageRating` for Resources

PR #2605, opened 2026-06-03
Add computed Airtable fields to db schema

Will-Howard:
## Description
- add Resource table mapping for computed Airtable fields only
- add computed response count field to Exercise
- export Resource/resourceTable and RESOURCE_FEEDBACK from @bluedot/db

## Issue

#2582

### DIFF
--- a/libraries/db/src/schema.ts
+++ b/libraries/db/src/schema.ts
@@ -1227,6 +1227,21 @@ export const unitResourceTable = pgAirtable('unit_resource', {
   },
 });
 
+export const resourceTable = pgAirtable('resource', {
+  baseId: COURSE_BUILDER_BASE_ID,
+  tableId: 'tblmcJcHQp1uKOK4q',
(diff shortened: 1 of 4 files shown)


libraries/db/src/schema.ts
```diff
@@ -1227,6 +1227,21 @@ export const unitResourceTable = pgAirtable('unit_resource', {
   },
 });
 
+export const resourceTable = pgAirtable('resource', {
+  baseId: COURSE_BUILDER_BASE_ID,
+  tableId: 'tblmcJcHQp1uKOK4q',
+  columns: {
+    computedNumCompletions: {
```
Will-Howard, line 1234: @marn-in-prod @joshestein Thoughts on "computed" as a naming convention to identify fields that are populated by the cron job (ultimately borrowed from [django-computedfields](https://pypi.org/project/django-computedfields/))

PR #2614, opened 2026-06-05
Implement computed-airtable-fields (framework for pushing aggregate fields back into Airtable)

### LINKED ISSUE #2582: Moving resource/exercise responses into Postgres-only: Create framework for pushing aggregate fields back into Airtable

Will-Howard:
# Description

Creates a new library, `computed-airtable-fields`:
- Computed fields are defined in `libraries/computed-airtable-fields/src/definitions.ts`, analogous to `schema.ts`
- A new cron job, `recomputeComputedAirtableFieldsCron` in pg-sync-service runs every 2 hours to recalculate the fields
- I suggest looking at the tests (libraries/computed-airtable-fields/src/core.test.ts) to get an idea of the properties of this recalculation logic, e.g. only writing changed rows
- I ran this locally and measured the time taken. It takes ~400ms/row, which is around 12 min for a full refresh with the 3 fields I have added here, with most of this being the single `db.update` calls for changed rows

## Issue

Fixes #2582

## Developer checklist


marn-in-prod (commented):
Mostly good, mostly changes on edge cases. I'll add them in a second

libraries/computed-airtable-fields/src/definitions.ts
```diff
@@ -0,0 +1,88 @@
...
+        const rows = await db.pg
+          .select({
+            rid,
+            avg: sql<string>`avg(${resourceCompletionTable.pg.resourceFeedback})`,
+          })
+          .from(resourceCompletionTable.pg)
+          .where(and(
+            arrayOverlaps(resourceCompletionTable.pg.resourceId, ids),
+            // != NO_RESPONSE also excludes NULL (NULL != 0 → UNKNOWN, filtered out).
+            ne(resourceCompletionTable.pg.resourceFeedback, RESOURCE_FEEDBACK.NO_RESPONSE),
```
marn-in-prod, line 76: `computedNumCompletions` filters on `isCompleted = true` but this average doesn't, so a row with `isCompleted: false` and an actual rating still moves the average while not counting as a completion.

This is currently possible in the UI, rate a resource and then uncomplete it. 

The data state already exists, but since this field is new we should probably consciously decide what we intend here and add a test to make the decision obvious (or a comment)

marn-in-prod (approved):
Looks good. Just the edge case that we should probably make a decision on

libraries/computed-airtable-fields/src/definitions.ts
Will-Howard: Seems reasonable 👍. I think I'll go with `isCompleted = true` everywhere to keep things simple (so we can keep the mental model that there's not much different between a row having `isCompleted = false` vs simply not existing)

Issue #2618, opened 2026-06-07
VKE cluster stuck on EOL k8s 1.33 — empty version metadata blocks upgrade (Vultr support)


joshestein:
## Summary

Our VKE cluster (`bluedot-prod`) is stuck on **end-of-life Kubernetes 1.33** and **cannot be upgraded by any method**. Root cause appears to be a Vultr-side bug: the cluster's `version` metadata is empty, which breaks upgrade-path validation. This is a blocker that requires **Vultr support** to resolve.

Prod is fully healthy on 1.33 in the meantime — no user impact. The constraint is the EOL auto-upgrade deadline (~**2026-07-04**), which is now also Vultr's problem (their forced auto-upgrade likely can't proceed either while the metadata is broken).

## Cluster

- ID: `83dad468-420e-4bdb-a5da-5b5395b22bdc`
- Label: `bluedot-prod`, region `ams`, single node pool (`vhf-4c-16gb`, 1 node)
- Control plane running `v1.33.0` (`kubectl` confirms); node is the original (~276d old)

## Evidence (Vultr-side bug)

- `GET /v2/kubernetes/clusters/{id}` returns **`"version": ""`** (empty), despite the control plane running v1.33.0.
- `available-upgrades` lists `v1.34.8+1`, `v1.35.5+1`, `v1.36.1+1`, but **no upgrade actually applies**:
  - Pulumi/provider upgrade to `v1.35.5+1` → `422 "Validation error: invalid upgrade version"` (local), and earlier `500 "Cluster upgrade failed:"` (CI, 4 retries).
  - Dashboard "Start Upgrade" flashes "latest version available," then re-offers the upgrade — never applies.

The empty `version` field appears to break the upgrade-path validation across every path (provider API, raw API, dashboard). Likely fallout from a prior forced auto-upgrade drift (see #1267).

## What we tried

- `pulumi up` locally and via CI → both fail (422 / 500).
- Raw `available-upgrades` query → valid targets listed, but upgrade action rejected.
- Vultr dashboard "Start Upgrade" → silent no-op / reverts.

## Action items

- [x] **File Vultr support ticket** — ask them to repair the cluster's `version` metadata (ticket text below).
- [x] **Revert master `vke.ts` → `v1.33.0+3`** to unblock infra CD (no-op vs live; stops the doomed retry). PR: _link to follow_.
[...]

PR #2628, opened 2026-06-08
[fix] Temporarily ignore VKE version to unblock infra CD

joshestein:
## What

Adds `ignoreChanges: ['version']` to the `vke-cluster` resource, as a **temporary** measure to unblock infra deploys.

## Why

The cluster can't be upgraded: Vultr reports its `version` as an empty string, which breaks upgrade-path validation and **re-poisons our Pulumi state on every read** (the provider reads the empty version back after each operation). The result is that *every* infra deploy — including the revert in #2619 — fails with `422 "invalid upgrade version"` while trying to reconcile the phantom version diff.

State edits don't hold (the read-back undoes them). `ignoreChanges` on the version field stops Pulumi from attempting the rejected upgrade, so unrelated infra changes can deploy again — e.g. #2627 (shadow pg-sync service).

## Why this and not the alternatives

- **vs editing state:** state edits get re-poisoned by the provider's read-back on the next run.
- **vs `pulumi up --exclude` in `deploy:cd`:** that skips the *whole* cluster resource and only fixes CI (a local `pulumi up` would still 422). `ignoreChanges` is scoped to just the `version` field and behaves consistently locally and in CI, and lives next to `version` where it's discoverable.

## Verification

`pulumi preview` against prod: `130 unchanged`, no error, and the cluster no longer appears in the plan at all (no version diff attempted).

## Temporary — remove when Vultr fixes it

This masks version drift on the cluster, so it must be removed once Vultr repairs the metadata. At that point the state self-heals and we resume the upgrade to `v1.35.5+1`. Tracked in #2618.

### DIFF
--- a/apps/infra/src/vultr/vke.ts
+++ b/apps/infra/src/vultr/vke.ts
@@ -13,6 +13,12 @@ export const k8sCluster = new vultr.Kubernetes('vke-cluster', {
   //   plan: 'vc2-1c-2gb',
   //   autoScaler: false,
   // },
+}, {
+  // TEMPORARY: Vultr reports this cluster's version as empty, which breaks upgrade
+  // validation and poisons our state on every read, failing all infra deploys.
+  // Ignoring version stops Pulumi attempting the (rejected) upgrade. Remove once
+  // Vultr repairs the metadata, then resume the upgrade. See #2618.
+  ignoreChanges: ['version'],
 });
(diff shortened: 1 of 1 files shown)


Will-Howard (approved):
LGTM! And seems relatively low risk because it will either work or not when we merge it

PR #2635, opened 2026-06-09
[feat] Sort speed-review session-complete list by outcome

dewierwan:
## What

On the speed-review **session complete** screen, the Accept and Reject columns now order applications by review outcome strength instead of the order they happened to be rated in:

- **Accept column:** strong-yes → weak-yes → neutral
- **Reject column:** neutral → weak-no → most-negative (moved-to-AGISC last)

So the strongest yeses sit at the top and the most-negative at the bottom, making it easy to scan the best/worst at a glance.

## How

Added a `RATING_RANK` map in `SessionComplete.tsx` and sorted the `accepted` / `rejected` lists by it. The sort keys off `effectiveRating` (not the raw rating), so if you re-rate a row in the summary, it re-sorts immediately.

## Scope note

Earlier iterations of this branch also explored a pre-round settings panel (configurable timer + expand-responses toggle); those were dropped at the author's request. This PR is **only** the sort change — a single-file, ~13-line diff with no layout/CSS changes.

## Testing

- `npm run typecheck`, `npm run lint`, `npm test` all pass.
- Verified locally against the real round/application data: rated across the full scale, concluded, and confirmed both columns order strongest-first and that re-rating a row re-sorts it.

## Screenshots

Not included. The session-complete screen sits behind Keycloak auth **and** is only reachable after submitting real review decisions to Airtable, so it can't be captured in a headless/throwaway run without writing to production data. The change is pure list-ordering with no visual restyling, so there's nothing new to show beyond the existing column layout.

### DIFF
--- a/apps/speed-review/src/components/SessionComplete.tsx
+++ b/apps/speed-review/src/components/SessionComplete.tsx
@@ -22,6 +22,16 @@ const RATING_OPTIONS: { value: RatingValue; humanOpinion: string; decision: stri
   { value: 'no', humanOpinion: 'Weak no', decision: 'Reject' },
 ];
 
+// Strongest yes (top) → most-negative (bottom), used to sort the result lists.
+const RATING_RANK: Record<RatingValue, number> = {
+  'strong-yes': 0,
+  yes: 1,
+  'neutral-accept': 2,
+  'neutral-reject': 3,
+  no: 4,
+  'moved-to-agisc': 5,
+};
+
 const sendOpinion = (id: string, rating: RatingValue) => {
   authFetch('/api/decisions', {
     method: 'POST',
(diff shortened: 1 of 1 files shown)


dewierwan:
@claude review

dewierwan:
Thanks both — both passes are clean (Greptile 5/5, Claude bot LGTM), no P1/P2 findings.

The one suggestion (memoizing `byStrength` with `useCallback`) I'm intentionally not applying: `.sort()` is called inline on each render regardless, so memoizing the comparator doesn't avoid any work, and the sorted lists are a single review session's worth of rows. It'd add a dependency-array footgun for no measurable gain. Leaving as-is.

No code changes, so no re-trigger needed.

PR #2636, opened 2026-06-10
[chore] Bump app Docker base images node 20 -> 22 (off EOL)

joshestein:
## What

Bumps all 10 node-based app Dockerfiles from `node:20-alpine` → `node:22-alpine` (digest-pinned). Node 20 hit EOL ~Apr 2026, so **production has been running an EOL runtime**. Node 22 EOL is ~Apr 2027 (~10mo runway).

This is distinct from #2594 (GitHub Actions runtime). That covers CI action versions; this is the node our containers actually run.

Apps: website, meet, availability, editor, course-demos, speed-review, room, pg-sync-service, app-template, frontend-example. (`login`/`posthog-proxy`/`storybook`/`website-proxy` are not node-based and untouched.)

## Why 22 (not 24)

`engines: ^22` already, and CI builds/tests on node 22, so app code is de-facto validated. 22 is the smallest safe hop off EOL with zero engines/CI churn. 24 (newest LTS) is a reasonable fast-follow once this lands.

## Validation

No `sharp`/native deps in any package.json, so runtime risk is low. Built + boot-checked the 3 representative apps locally (Colima, docker 28.4):

- **pg-sync-service** — the only app running `npm ci` in-image; resolved cleanly on node 22; runtime reports `v22.22.3`.
- **website** — Next.js standalone; image builds, server boots to `v22.22.3` (only 500s on missing local secrets, unrelated to the bump).
- **meet** — Zoom Web SDK; builds, runtime `v22.22.3`.

The other 7 are COPY-and-run clones of the same base; covered by the CI build/deploy matrix.

## Rollout

- Merge auto-deploys **9 apps straight to prod** (their `deploy:cd` = single env). Plan: monitor the `cd_*` matrix + app health on merge; revert is a one-line digest swap.
- **website** deploys to staging on merge only; prod needs a `website/vX.Y.Z` release tag. I'll cut that release as the final step. **EOL fix is not complete for website until that release ships.**

Closes #2622

PR #2642, opened 2026-06-10
[fix] Disable buildkit provenance/sbom so ghcr push doesn't fail

joshestein:
## Problem

After the node 20→22 base bump (#2636) merged, CD started failing on `docker push` to ghcr with `unknown blob` for several apps (course-demos, editor, frontend-example, pg-sync-service). All image **layers** push fine ("Pushed" / "Layer already exists"); the failure is at the **manifest** push.

## Root cause

Modern buildkit attaches **provenance** and **SBOM** attestations to builds by default. That turns the output into an **OCI image index** referencing two manifests — the real image manifest plus an attestation manifest — instead of a single image manifest. ghcr rejects the attestation-bearing index on push with `unknown blob`.

The node bump didn't introduce the buildkit behaviour, but it forced a **fresh manifest push for every app** (previously, unchanged apps were no-op pushes that found everything already in ghcr), which exposed this latent issue.

Note: pinning a platform-specific base digest would **not** fix this — the attestation manifest is attached to the *built* image regardless of base image type.

## Fix

Add `--provenance=false --sbom=false` to `docker build` in the three deploy scripts that push to ghcr (`deploy.sh`, `multistage/deploy-staging.sh`, `multistage/deploy-production.sh`). The build then emits a single clean manifest. `test.sh`/`start.sh` don't push, so they're untouched.

## Validation

Local (Colima, docker 28.4):
- **Default build** → `exporting attestation manifest` + `exporting manifest list` (the OCI index ghcr rejects).
- **`--provenance=false --sbom=false`** → single `exporting manifest`, no index.

Attestations are supply-chain metadata, not required to run the image; disabling them is the standard remedy for registries that don't accept attestation manifests.

### DIFF
--- a/libraries/docker-scripts/src/deploy.sh
+++ b/libraries/docker-scripts/src/deploy.sh
@@ -7,8 +7,10 @@ IMAGE_NAME="bluedot-$APP_NAME"
 VERSION_TAG="$(TZ=UTC date +'%Y%m%d.%H%M%S').$(git rev-parse --short HEAD)"
 
 # Build
+# --provenance=false --sbom=false: otherwise buildkit emits an OCI image index with
+# attestation manifests, which ghcr rejects on push with "unknown blob".
 APP_NAME=$APP_NAME VERSION_TAG=$VERSION_TAG npm run build --if-present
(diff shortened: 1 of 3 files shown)


joshestein:
@claude review

joshestein:
This is nonsense; it was a transient error.

PR #2649, opened 2026-06-11
[feat] add temporary capacity notice banner to grant + advising pages

dewierwan:
## What

Adds a temporary "note on timing" banner to `/programs/career-transition-grant` and `/programs/advising`, telling applicants that application reviews will be slower than normal until early July while team members relocate to San Francisco. Applications stay open as usual.

## How

- Reuses the existing `AnnouncementBanner` component, placed directly below each page's `GrantStatsStrip` (so it sits next to the "Avg days to decision" stat it qualifies).
- Adds a `dismissible` prop to `AnnouncementBanner` (default `true`, so the global `_app.tsx` banner is unchanged). The capacity notice passes `dismissible={false}`: no Dismiss button, and non-dismissible banners ignore previously persisted dismissals in localStorage.
- Updates the advising page's fallback description to include "and biosecurity" (the live value comes from the Programs Airtable table and is being updated separately).

The banner is intentionally not auto-expiring; it will be removed manually in early July.

## Testing

- `cd apps/website && npm test` — 117 files, 879 passed, 1 skipped.
- `npx tsc --noEmit` — clean. `eslint` clean on changed files.
- Viewport sweep at 320/480/600/720/1024/1440/1920 (heights 700–1800, retina DPR) on both pages: banner visible below the stats strip at every width, no Dismiss button, no horizontal overflow introduced. (Note: there is a pre-existing 64px nav overflow at exactly 1024px wide that also reproduces on production — unrelated to this change.)
- Verified non-dismissible behaviour by injecting a persisted dismissal key into localStorage and confirming the banner still renders.

### `/programs/career-transition-grant`

|            | Before          | After               |
| ---------- | --------------- | ------------------- |
| 📱 Mobile  |  |  |
| 🖥️ Desktop |  |  |

### `/programs/advising`

|            | Before          | After               |
| ---------- | --------------- | ------------------- |
| 📱 Mobile  |  |  |
| 🖥️ Desktop |  |  |


dewierwan:
@claude review

dewierwan:
Addressed Greptile's feedback in commit `[test] cover dismissible={false} banner behaviour + storybook variant`:

- Added two tests to `AnnouncementBanner.test.tsx`: non-dismissible banners render no Dismiss button, and they render even when a dismissal for the same content key is persisted in the store (16 tests passing).
- Added a `NonDismissible` story to `AnnouncementBanner.stories.tsx`.

Claude's review found no issues; nothing else to address. CodeRabbit was still processing at time of writing.

PR #2660, opened 2026-06-14
[chore] infra: remove temporary VKE upgrade node pool

joshestein:
## What

Post-upgrade cleanup for the VKE `v1.33.0+1 → v1.35.5+1` upgrade.

- **Remove `vke-node-pool-upgrade`.** It was a temporary second node pool added so the single-node cluster had a drain target during the rolling upgrade. The node has already been **manually drained** (`kubectl drain ... --ignore-daemonsets --delete-emptydir-data`) and workloads rescheduled onto the main node, so this just deletes an empty, cordoned VM.
- **Keep `enablePDB: false` permanently** on the three single-instance CNPG clusters (keycloak-pg, grafana-pg, airtable-sync-pg). A `minAvailable: 1` PDB on a one-replica cluster provides no availability (no replica to fail over to) and only blocks node drains — it was the original upgrade preflight blocker. Comments updated to explain; re-enable only if these go multi-instance.

## Sequencing (already done manually before merge)

Deleting a Vultr node pool hard-deletes the VMs without a graceful drain, so the node was drained by hand first. The pg pods were on the main node and stayed put — no DB disruption. On merge, CI's `pulumi up` removes the now-empty pool.

## Verification

- `npx tsc --noEmit` in `apps/infra` — passes.
- Cluster post-drain: 2 nodes on `v1.35.5` (one cordoned/empty awaiting this deletion), all 3 pg clusters `Cluster in healthy state`, website/login/ingress recovered.

## Follow-ups (not in this PR)

Today's upgrade caused a multi-minute outage because every deployment is single-replica and Vultr replaced both nodes at once. Worth separate PRs:
- 2 replicas on critical deployments (website, login, ingress) so a node roll drains gracefully instead of taking the site down.
- GH issue #2661

### DIFF
--- a/apps/infra/src/k8s/postgres.ts
+++ b/apps/infra/src/k8s/postgres.ts
@@ -22,9 +22,10 @@ export const keycloakPg = new k8s.apiextensions.CustomResource('keycloak-pg', {
   },
   spec: {
     instances: 1,
-    // Temporarily disabled for VKE k8s upgrade: single-instance CNPG creates a
-    // primary PDB with minAvailable 1, which blocks node drain entirely. Revert
-    // to default (remove this line) once the upgrade completes.
+    // Disabled deliberately. With instances: 1 there's no replica to fail over to,
+    // so the default primary PDB (minAvailable 1) provides no availability and only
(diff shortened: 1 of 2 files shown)


joshestein:
@claude review

PR #2662, opened 2026-06-14
[docs] infra: document VKE upgrade preflight + drain-target procedure

joshestein:
## What

Extends the "Upgrading the Kubernetes version" section of `apps/infra/README.md` with the preflight + teardown procedure learned during the `v1.33.0+1 → v1.35.5+1` upgrade.

Adds:
- The two preflight blockers (single-node cluster, single-instance CNPG `-primary` PDBs) and how to clear them.
- Adding a temporary node pool as a drain target — as a **new** pool, not a `nodeQuantity` bump (Vultr 422s in-place node-count changes when `autoScaler: false`) — and draining it before deletion.
- Why `enablePDB: false` is kept permanently on the single-instance pg clusters.
- The expected full cold-start when Vultr replaces both nodes at once.

Docs only.


joshestein:
Addressed Greptile's note: the `kubectl drain` example now includes `--ignore-daemonsets --delete-emptydir-data` so it works copy-pasted. CodeRabbit was rate-limited, no findings.

PR #2685, opened 2026-06-19
Attempt to fix pg-sync-service outage: Remove AbortSignal

Will-Howard:
# Description

There is an outage in `pg-sync-service` currently, with errors like `Expected signal to be an instanceof AbortSignal` originating from deep within the official `airtable` package. This appears to be ultimately due to a polyfill version of Abort signal failing this check: `proto.constructor.name === "AbortSignal"`, because it uses the name `AbortSignal3` instead.

Hypothesis: Adding a use of AbortSignal in code caused the bundler to add this polyfill, so removing it will fix it. We don't actually need this AbortSignal, it was just nice-to-have.

## Issue

N/A

## Developer checklist

### DIFF
--- a/libraries/computed-posthog-events/src/core.ts
+++ b/libraries/computed-posthog-events/src/core.ts
@@ -162,7 +162,6 @@ export async function forwardEventTypeToPostHog({
           historical_migration: historicalMigration,
           batch: prepared.map((p) => p.event),
         }),
-        signal: AbortSignal.timeout(60_000), // don't let a hung request hold the cron's reentry guard
       });
       if (!res.ok) {
         throw new Error(`PostHog /batch failed: ${res.status} ${res.statusText} ${await res.text().catch(() => '')}`);


Will-Howard:
This did resolve the issue, crossposting the write-up from Slack which has a more correct description of the problem (minor: the point is that the [airtable package](https://www.npmjs.com/package/airtable) already used its own polyfill, so adding a new usage of AbortSignal caused the bundler to rename it to AbortSignal3):
>**Summary**
> [This PR](https://github.com/bluedotimpact/bluedot/pull/2675) added a usage of AbortSignal in pg-sync-service. The usage itself worked, but it caused the bundler to change the name of the AbortSignal polyfill that the airtable package uses, and this hit an error condition deep within node-fetch: proto.constructor.name === "AbortSignal".
>
> As a result, requests to airtable failed and syncing was broken for ~24h. This was fixed by reverting the addition of AbortSignal ([second PR](https://github.com/bluedotimpact/bluedot/pull/2685)), and doing a full sync to catch up.
>
>**Timeline**
> 2026-06-18 10:52am UTC (not 11pm as I said above): PR 1 deploys, pg-sync-service stops syncing
> 2026-06-19 10:46am UTC: PR 2 deploys, syncing starts again for newly updated rows. Full sync completes at 10:57am
>
>**Impact**
>The website and meet app mostly kept working because their writes replicate to Postgres synchronously, independent of pg-sync. Features that degraded due to changes in Airtable not being picked up:
> - Group switching: The new group wouldn’t appear in the UI for the user, though they would still receive new calendar events
> - Meeting attendance: A user who switched into group during the outage window wouldn’t see their name come up in the attendance list
> - Facilitators assigning substitutes and changing discussion times: Similar to the above, though no meetings were actually affected during the window
>
>There were 15 users who tried to switch to a different discussion that started inside the outage window. Of these, it seems most (11) were able to find their way in (presumably via the calendar event). There were 4 who were recorded as absent.

PR #2689, opened 2026-06-22
Show participants from all groups in facilitator feedback form

### LINKED ISSUE #2688: Facilitator feedback form: Handle the case where the user facilitated multiple groups in a round
### 🧩 Summary

We received a bug report from a user who facilitated two groups, saying they were unable to submit feedback for the second group. This is the consequence of a wrong assumption in the form, which is that one meetPerson can only facilitate one group. In fact there are many cases of a meetPerson facilitating multiple groups within the same round. The logic is:
- One application -> one meetPerson -> potentially facilitating several groups within the same round, but never across different rounds

Since the on-site version of the form launched, this has affected about 20 facilitators, this is just the first bug report we have received.

Prior to the on-site form launching, the norm was one form covering all groups the user facilitated within that round. So for the change here I suggest just making a narrow edit to show all participants for all groups within the one form, without adding any "Group N" indicator or anything (we could add that later, but it's not needed to fix the bug).

### 🛝 Steps to reproduce

1. Find a user who is facilitated multiple groups within the same round
2. Impersonate them and go to https://bluedot.org/facilitated-courses?tab=pastCourses
3. Click through "Submit feedback"/"Edit feedback" on one of the affected groups
4. The form will show participants from one of the groups, but not the other

### ✅ Definition of done

The facilitator feedback form shows all participants for all groups the meetPerson facilitated.

### 📌 Notes / Designs

_No response_

Will-Howard:
# Description

Previously it would only show the first group associated with the facilitators meetPerson, see #2688 for a full description of the issue.

## Issue

Fixes #2688

## Developer checklist


Will-Howard:
@greptileai review

Issue #2700, opened 2026-06-24
Finish migrating resource_completion from isCompleted to completedAt


Will-Howard:
### 🧩 Summary

I started migrating resource_completion from using boolean `isCompleted` to using a `completedAt` timestamp. This was for parity with exercises, and to allow submitting `resource_completed` events to PostHog with a correct timestamp.

The current state (following #2684 and #2691) is that both `isCompleted` and `completedAt` are set, and `isCompleted` is still used on read. The state we want to get to is:
- `completedAt` is used exclusively on both read and write
- `isCompleted` is dropped

### ✅ Definition of done

- `completedAt` is used exclusively on both read and write
- `isCompleted` is dropped

PR #2703, opened 2026-06-24
[refactor] Read resource completion status from completedAt, not isCompleted

Will-Howard:
# Description

Switches all reads of resource_completion to derive completion from completedAt (IS NOT NULL) instead of the isCompleted boolean: the course-progress query, the computed Airtable fields (num completions / average rating), and the ResourceListItem display + optimistic cache. The write path still sets both columns.

## Issue

#2700

## Developer checklist


Will-Howard:
> a backfill is needed first.

The backfill has been done locally! So this is safe to deploy now

PR #2712, opened 2026-06-30
[feat] Top Talent CRM extension backend (SSO-gated Notion proxy)

dewierwan:
## What

Server-side endpoints so the **Add to Top Talent CRM** browser extension no longer holds a shared Notion token on every teammate's machine. The Notion token now lives only as a server secret (`NOTION_API_TOKEN`); the extension authenticates each call with the user's BlueDot Keycloak identity, and the new row's **Owner** is resolved automatically from the signed-in user's email.

This is the backend half of moving the extension from "clone the repo + paste a shared token" to "install from the private Chrome Web Store + sign in with BlueDot."

## Changes

- **`lib/api/topTalentCrm.ts`** — Notion helpers (`findByLinkedIn`, `createPerson`, `updatePerson`, `resolveOwnerByEmail`), ported faithfully from the extension's old client-side `notion.js` (raw `fetch`, no new dep). Includes the dated note-bullet body and the update-prepends-a-note flow.
- **`api/top-talent-crm/{check-dup,create,update}.ts`** — authenticated routes via `makeApiRoute({ requireAuth: true })`. Each is gated to `@bluedot.org` emails because the `customers` Keycloak realm also contains course participants — realm membership alone is not staff.
- **env + infra** — `NOTION_API_TOKEN` added to `env.ts` (optional), `secrets.ts` (Pulumi→K8s), and the staging + production website services in `serviceDefinitions.ts`.
- **Tests** — `topTalentCrm.test.ts`: staff gate (incl. lookalike-domain), match mapping, owner resolution, create payload.

## Verification

- `npx tsc --noEmit` — clean (website + infra).
- `npm run lint` on changed files — clean.
- `npm test` — full website suite green (911 passed); new unit tests 7/7.
- The four Notion ops were validated end-to-end against the real CRM with a scratch token (read DB, resolve owner by email, query by LinkedIn, create+archive a throwaway row).

No UI changes in this repo, so no screenshots.

## Blocking prerequisites before this works in a deployed env

1. **Pulumi secret value.** `secrets.ts` uses `config.requireSecret('notionApiToken')`, so `pulumi up` will fail until someone with infra access runs `pulumi config set --secret notionApiToken <token>`. Set this before applying infra.
2. **Token capability.** The Notion integration must have *"Read user information including email addresses"* (needed for automatic Owner resolution).

## Open question for reviewers (auth)

The extension authenticates via a dedicated Keycloak public client (`top-talent-crm-extension`) in the `customers` realm, with an audience mapper adding `bluedot-web-apps`. `loginPresets.keycloak.verifyAndDecodeToken` does a **strict** `payload.aud !== 'bluedot-web-apps'` check (`Login.tsx`). If Keycloak emits `aud` as an array (`["bluedot-web-apps"]`) rather than a plain string, every call will 401. Confirming the exact `aud` shape from this client is the one remaining integration risk; if it's an array, we should make that check array-tolerant (spec-correct per RFC 7519) in a follow-up.


dewierwan:
@claude review

dewierwan:
Thanks for the review — addressed in `ff7965c2`:

- **Greptile P1 (security, `update.ts`): unverified `pageId`.** Fixed. `updatePerson` now calls `assertPageInCrm` first — `GET /pages/{id}` and asserts `parent.database_id === CRM_DATABASE_ID` (hyphen-insensitive) before any PATCH/DELETE, so an authenticated staffer can't mutate arbitrary pages the integration token can see. Added unit tests for both accept and reject.
- **Greptile P1: swallowed block-delete errors.** Fixed. The old-block `DELETE`s now go through `ensureOk`, so a failed deletion surfaces instead of silently leaving a doubled note list.
- **Greptile P1: array `aud` 401.** Fixed at the source in `Login.tsx` — the check now accepts `aud` as a string **or** an array containing `bluedot-web-apps` (RFC 7519 §4.1.3). This is the shared verifier, so flagging for explicit human review: it's strictly more permissive in a spec-correct way and the single-string path is unchanged, so existing apps behave identically. All `@bluedot/ui` tests still pass.
- **CodeRabbit (minor): `resolveLocalDate` accepted impossible dates.** Fixed via a round-trip parse; `2026-99-99` now falls back to the server date.

Verification: `tsc` clean (website + the `Login.tsx` change), eslint clean on changed files, website suite 920 passing, CRM unit tests 11/11.

Not changed: CodeRabbit's docstring-coverage warning (76.9% vs 80%) — the functions already carry explanatory comments; happy to add formal JSDoc if you'd prefer the threshold met.

@claude review

Issue #2709, opened 2026-06-29
Deprecate email-as-foreign-key in code (use Applications base user id instead)


Will-Howard:
### 🧩 Summary

There are three objects in code that still use `email` as the primary lookup key:
- Exercise responses
- Resource completions
- `meetPerson`s, i.e. the Course runner sync of course registration

For exercise responses and resource completions, we should start using the `userId`/`createdByUserId` that are already on the table. For `meetPerson`, we should add an `applicationsBaseUserId` to the table lookup via auth.email -> user -> meetPerson; this will make it straightforward to fully migrate away from `email` when we start using the Keycloak `sub` property to look up the user.

I suggest also migrating `resourceCompletionPgTable` to have a `userId` which points to the Applications base user, to be in line with exercise responses. The `createdByUserId` which points to a Course builder user is a leftover from when these tables were in Airtable, and can be deleted (once it's been used in for the migration).

This will require several distinct PRs to be deployed sequentially, and some local scripts to backfill existing values.

### ✅ Definition of done

We no longer use email as foreign key anywhere in code.

Will-Howard:
I realised the documentation of the backfilling steps has been a bit sparse so far (following [this review](https://github.com/bluedotimpact/bluedot/pull/2721#pullrequestreview-4668354906)). I'm summarising what has been done so far in this comment, and from now on I'll put any backfill scripts up as a PR so there is an audit trail.

Backfills done so far in service of this issue:

## 1. resource_completion.createdByUserId

- When: 2026-07-01 13:17
- Table / column: `resource_completion.createdByUserId` (-> Course-builder User table)
- Rows updated: 70,993 (out of ~253,859)
- Method: Local script

Prerequisite hop for step 2. `createdByUserId` was a pre-existing field where most values were filled, but some were null despite having a matching user (all old rows). For each empty row, the script matched its `email` to a
`course_builder_user` and wrote that id. The 70,993 figure is the linkable-and-updated count; These rows were skipped:
550 with no matching course-builder user, 25 with no email.

## 2. resource_completion.userId: bulk pass

- When: 2026-07-01 13:08 (pass 1) and 13:18 (pass 2, after step 1 populated more createdByUserId)
- Table / column: `resource_completion.userId` (-> Applications base user)
- Rows updated: 253,859 (almost all rows)
- Method: Local script

[...]

PR #2714, opened 2026-06-30
Add userId (pointing to Applications base) to resource completions

Will-Howard:
# Description

Previously this used the Course builder user id. Now this table doesn't live in Airtable, there is no need for this indirection, so we can point it at the regular `userTable`.

## Issue

#2709

## Developer checklist


Will-Howard:
Breadcrumb: I just ran a script locally to backfill this for existing rows, so it's now safe to rely on the userId field in reads

PR #2727, opened 2026-07-02
Get-or-create the user row in protectedProcedure and attach it to ctx

### LINKED ISSUE #2722: Remove the authenticated-but-no-user edge case shortly after first login
### 🧩 Summary

The current process for user creation is:
1. The user logs in via Keycloak, and their Keycloak auth info is set in local storage
2. They are redirected to the `/login/oauth-callback` page, which fires a `users.ensureExists` to create the user if one doesn't exist
(3). The user can also be created in other ways, e.g. when the someone applies to a course via [this automation](https://airtable.com/appnJbsG1eWbAdEvf/wflYcVcYhaPynHIOo/wacd7r58PBie7wbk8)

This creates a small time lag in which a user can be authenticated but not have a row in `userTable` yet. This has caused one issue in the past (#2670), and has the potential to cause more when we switch a bunch of email-based lookups to be userId-based in #2721.

Suggested solution to fix this: get-or-create the user in `createContext`, or in `protectedProcedure`, and add it to the `ctx` object. Reasoning:
- This makes it true by definition that the user exists in any `protectedProcedure`, which saves a lot of guard-ing/edge cases
- Following #2721, we will need to look up the user in almost all `protectedProcedure`s anyway, so attaching it to the context will avoid some boilerplate
- Downsides: It will add 1 SQL query to every request, which may be wasted in some cases (especially if we add it to `createContext`, which will make it run on logged out routes too

If we implement this, the `ensureExists` we run in `oauth-callback` will become almost redundant. The only other thing it achieves it to attach the `initialUtmSource` etc to the user (and, this is actually semi-broken currently because of point (3) above, if a user has already been created by an automation their `initialUtmSource` won't be recorded). I suggest we change `ensureExists` to something like `trackUtmOnLogin` and make it exclusively for that purpose (and fix the issue due to point (3)).

### ✅ Definition of done

As far as backend routes are concerned there is no longer an allowed condition where a user is authenticated but no `userTable` row exists

### 📌 Notes / Designs

_No response_

Will-Howard:
# Description

TODO

## Issue

Fixes #2722

## Developer checklist

### DIFF
--- a/apps/website/src/__tests__/posthog.test.tsx
+++ b/apps/website/src/__tests__/posthog.test.tsx
@@ -128,7 +128,7 @@ describe('PostHog UTM tracking: End-to-end tests of key points where UTM params 
     // Step 2: User completes OAuth and lands on oauth-callback
     // The OIDC response includes the redirectTo with UTM params
     mockProcessSigninResponse.mockResolvedValue(createMockOidcResponse({ userState: { redirectTo: redirectToWithUtm } }));
-    server.use(trpcMsw.users.ensureExists.mutation(() => ({ isNewUser: true })));
+    server.use(trpcMsw.users.trackUtmOnLogin.mutation(() => ({ isNewUser: true })));
 
     render(<OauthCallbackPage />, { wrapper: TrpcProvider });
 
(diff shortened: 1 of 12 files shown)


Will-Howard:
I've decided to abandon this approach in favour of [running the existing ensureExists earlier](https://github.com/bluedotimpact/bluedot/pull/2734). The main reason is because I realised this increases the chance of duplicates by a lot (because it runs this get-or-create on every request), and solving that with locking would be complex + potentially slow down every query.

PR #2728, opened 2026-07-02
[feat] website: backfill facilitator quick-apply name by email

anglilian:
# Description

When a facilitator uses the quick-apply flow (`/facilitator-applications/quick-apply?round=…`), the form collects their answers but never their name. As a result the new `course_registration` records were created with blank `firstName`/`lastName`. The name still *displayed* because the `fullName` formula field falls back to a lookup on the linked `user` record — but the discrete name columns themselves were empty.

This change backfills those columns at insert time. `resolveApplicantName(email)` prefers the human-entered split from the applicant's most recent prior application that recorded a name (any course/role — typically their original participant application, so the first/last split is accurate). When no prior application has a name, it falls back to the applicant's `user` account name (looked up by email, split on the first space) so the fields are populated regardless. Names are trimmed, and blank / whitespace-only values are treated as "no name".

## What changed

1. **`apps/website/src/server/routers/facilitator-applications.ts`** — new `resolveApplicantName` helper (with `cleanNamePart` and `splitName` sub-helpers), called in the `quickApply` mutation to set `firstName`/`lastName` on the inserted `course_registration`.
2. **`apps/website/src/server/routers/facilitator-applications.test.ts`** — unit tests for `resolveApplicantName`: prefers a prior application over the account, most-recent prior application wins, trims surrounding whitespace, ignores whitespace-only names, falls back to the account name split on first space, single-word name → null last name, and email isolation across both sources.

## Notes on the name source

No available source has a reliably clean first/last split: prior quick-apply records are blank, `meet_person` splits the same person's name inconsistently across records, and the `user` account stores only a single combined `name`. The hybrid here uses the most accurate split available (a human-entered prior application) and only falls back to splitting a combined name when nothing better exists. Splitting on the first space keeps `firstName` correct (the part used in greetings) at the cost of an approximate `lastName` for multi-word names.

## Issue

N/A — follow-up to the facilitator quick-apply feature; no linked issue.

## Developer checklist

- [x] Front-end code follows Component Accessibility Checklist — N/A, backend-only change (tRPC router).
- [x] Considered having snapshot tests / functional tests — added 8 unit tests for the new helper.
- [x] Considered adding Storybook stories — N/A, no component change.

No DB schema changes.

## Verification

Backend-only change (no UI surface). The `quickApply` mutation's insert can't run under the PGlite test harness because `courseId` is Airtable-computed and `notNull` in Postgres (same documented limitation as the existing `quickApply` test), so the name-resolution logic is verified directly via `resolveApplicantName` unit tests.

Type check + lint + full test suite all pass:
```
npm run typecheck   # clean
npm run lint        # clean (--max-warnings=0)
npm test            # 928 passed, 1 skipped (119 files) — run twice, stable
```

## Screenshots

N/A — invisible diff (backend logic only, no visual change).

### DIFF
--- a/apps/website/src/server/routers/facilitator-applications.test.ts
+++ b/apps/website/src/server/routers/facilitator-applications.test.ts
@@ -538,6 +538,11 @@ describe('resolveApplicantName', () => {
     expect(await resolveApplicantName(CALLER_EMAIL)).toEqual({ firstName: 'Cher', lastName: null });
   });
 
+  test('treats a whitespace-only user account name as missing', async () => {
+    await seedUser(CALLER_EMAIL, '   ');
+    expect(await resolveApplicantName(CALLER_EMAIL)).toEqual({ firstName: null, lastName: null });
+  });
+
   test('does not read another user\'s name from either source', async () => {
     await seedUser('other@example.com', 'Someone Else');
     await testDb.insert(courseRegistrationTable, {
--- a/apps/website/src/server/routers/facilitator-applications.test.ts
+++ b/apps/website/src/server/routers/facilitator-applications.test.ts
@@ -4,6 +4,7 @@ import {
   courseTable,
   groupDiscussionTable,
   meetPersonTable,
+  userTable,
 } from '@bluedot/db';
 import { describe, expect, test } from 'vitest';
 import {
@@ -13,6 +14,7 @@ import {
   testAuthContextLoggedOut,
   testDb,
 } from '../../__tests__/dbTestUtils';
+import { resolveApplicantName } from './facilitator-applications';
 
 setupTestDb();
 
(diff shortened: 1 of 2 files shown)


anglilian:
@claude review

anglilian:
Thanks @coderabbitai — good catch. Fixed in 1d9c66f0: the user-account fallback now cleans `user.name` with `cleanNamePart` before the truthiness check, so a whitespace-only account name falls through to `{ firstName: null, lastName: null }` instead of producing an empty-string first name. Added a test (`treats a whitespace-only user account name as missing`).

anglilian:
Re: Greptile P2 on `resolveApplicantName` (DB-level ORDER BY) — deferring intentionally. The sibling helper `getPriorFacilitatorRegs` in this same file also fetches by email and sorts in application code, so moving the sort into SQL here would introduce single-file inconsistency for a negligible gain (the query is bounded by one person's registrations). `desc()` in Postgres also orders NULLs first, which differs from the current `?? 0` tie-break, so it isn't a drop-in swap. Keeping the JS sort for consistency; happy to revisit as a separate sweep if we want to standardise the pattern.

Issue #2647, opened 2026-06-11
Exercises: support "optional" and no-text-required exercises


Will-Howard:
### 🧩 Summary

We got a [bug report](https://airtable.com/app63L1YChHfS6RJF/pagZHAWawFuIDo9cf?EAVEJ=recA8x4EhvK3ivXzr) about being unable to tick off an optional exercise. IMO there are two underlying problems here:
1. Optional exercises still count towards progress, and appear in the same section in the UI as regular exercises. This is in contrast to resources, which have a schema-level optional flag, and where optional ones appear under a collapsed section
2. The exercise doesn't really require any text, it requires you to go away and do something. There are a few other exercises like this throughout the site

### 🛝 Steps to reproduce

See the "(Optional)" exercise on [this page](https://bluedot.org/courses/technical-ai-safety/2/3) as an example

### ✅ Definition of done

1. Optional exercises are supported in the UI (don't count towards completion, distinct beyond just having "(Optional)" in the title)
2. The current no-text exercises are supported better, either by explicitly allowing no text, or rephrasing them to require something concrete (e.g. a link to what you did elsewhere)

Will-Howard:
As a stop-gap, I'll make it so optional exercises don't count towards course completion. @anglilian can I put it in your court to decide what to do here as the ultimate solution, as it's more of a product decision. A straightforward option would be:
- Do explicitly support no-text exercises that you can just tick off
- Move optional exercises to a collapsed section as we do with optional resources

anglilian:
Let's move them to optional exercises to be similar with optional resources! We can change[ this field ](https://airtable.com/appbiNKDcn1sGPGOG/tbla7lc2MtSSbWVvS/viwFbvo6ixuMWr8pq/flda5e542i9w1nBzv) to have the options: Core, Further, Maybe so it follows a similar pattern to the Chunk resources.

@Will-Howard

Will-Howard:
> Let's move them to optional exercises to be similar with optional resources!

Ok I'll do that! On a similar topic, what are your thoughts on how to handle the case in the AGI Strategy course where we want users to complete 1 out of 4 exercises, each of which live in a different chunk ([Issue](https://github.com/bluedotimpact/bluedot/issues/2718))?

Will-Howard:
[...]

PR #2732, opened 2026-07-03
Cleanup following exercise status field migration

Will-Howard:
# Description

Cleanup following #2731, now that all of the exercises have been migrated to the new values in Airtable (all 575 exercises: Active+Optional -> Further, Active -> Core, Inactive unchanged):

- Removes support for the legacy `Active` + `isOptional` states. The possible states are now just: Status == "Core" (required), Status == "Further" (optional, shown under "Optional Exercises", not counted), Status == "Inactive" or "Maybe" (hidden, not counted)
- Inlines the now-trivial status checks and deletes the transitional helpers #2731 added
- Moves `isOptional` to `deprecatedColumns` in the db schema (keeps the Postgres column during the deploy window, stops syncing/using it). A follow-up PR deletes it fully once this has deployed to production, per the two-PR column-removal process

## Issue

#2647 (follow-up to #2731)

## Developer checklist

## Screenshot

No visual changes (the new rendering shipped in #2731; this only removes dead code paths).


Will-Howard:
Regarding all the comments regarding losing support for `status = "Active"`: I have migrate all of these to "Core" now, so this shouldn't be an issue.

PR #2738, opened 2026-07-06
[feat] Route pg-sync validation warnings to #update_pg-sync, escalate spikes

joshestein:
## Problem

`#update_tech-prod` is flooded with pg-sync Airtable validation warnings. These are mostly **tolerated schema/Airtable drift**, so they drown out genuinely actionable alerts and we miss them.

## Change

- **Route validation warnings to a dedicated channel** (`PG_SYNC_SLACK_CHANNEL_ID` → `#update_pg-sync`) instead of the main alerts channel.
- **Spike escalation** to avoid the side channel becoming a graveyard: `slackAlert` gains opt-in `spikeThreshold` + `escalationChannelId`. When a batched signature affects ≥ `spikeThreshold` distinct records in one flush window, the summary is cross-posted to the prod alerts channel. 
- **Removed the per-warning `RateLimiter` from `db.ts`.** Batching already bounds output, and the limiter dropped warnings *before* they reached the batcher, which blinded the spike record-count. `RateLimiter` itself is unchanged and still used for Airtable API throttling.

## Notes

- Threshold is **absolute** (distinct records/window), not relative to a baseline. Catches "a drift blew up base-wide"; won't catch a slow ramp. Baseline detection would need a persistent per-signature store, intentionally out of scope.
- Trade-off: a *new, low-volume* validation warning still only lands in `#update_pg-sync`. A novelty allowlist could promote first-seen signatures later; deferred.

## Ops before/at merge

- Invite bot to `#update_pg-sync` — `chat.postMessage` fails **silently** with `not_in_channel` (caught + logged, no message sent) if the bot isn't a member.


joshestein:
@claude review

joshestein:
@claude review

joshestein:
Thanks @greptile-apps. Addressed in cb250dc6:

- **Spike options frozen at batcher creation (P2):** fixed. `addToBatch` now refreshes `spikeThreshold`/`escalationChannelId` on the existing batcher (last-write-wins) instead of silently dropping a later call's options.
- **Occurrence-count fallback coverage:** added a test for warnings with no record IDs (spike driven by raw occurrences), plus a test for the later-call option refresh. 18 → 20 tests.
- **Escalation fires even if the primary side-channel send failed:** intentional, leaving as-is. The escalation message embeds the full batch summary, so it's self-contained — if the `#update_pg-sync` post fails (e.g. `not_in_channel`), failing *open* to the main alerts channel is safer than suppressing the spike entirely.

PR #2743, opened 2026-07-07
[chore] Provision PG_SYNC_SLACK_CHANNEL_ID for pg-sync-service

joshestein:
Infra-only PR: adds the (non-secret) `PG_SYNC_SLACK_CHANNEL_ID` env var (channel `#update_pg-sync`, `C0BFEG755PG`) to the `bluedot-pg-sync-service` deployment.

Split out from the code change per `apps/infra/ADDING_ENV_GUIDE.md`: the variable must exist on the deployment **before** the code marks it required, otherwise the pod crashloops on "Missing required environment variable".

**Merge + deploy this first**, then the code PR (#2738) that consumes it.

After merge, `pulumi up` runs but does not auto-restart pods for infra-only changes; a rollout may be needed so the new pod picks up the var (the code PR's deploy will also restart it).


joshestein:
@claude review

Issue #2760, opened 2026-07-09
posthog-computed-events: Remove `identify(email)` events emitted by `application_submitted`


Will-Howard:
### 🧩 Summary

Currently, the `application_submitted` event definition in computed-posthog-events fires `identify(email)` events as a side effect. This let's us join up the "click apply button" -> "submit application" -> "logged in activity" funnel.

Sending these `identify(email)` calls is a spanner in the works for deprecating email as a user identifier (#2713), because it breaks the assumption (that we're trying to add) that email is just a mutable user property, and Keycloak's `sub` is the immutable user identifier.

Proposed solution:
1. Submit `application_submitted` events as anonymous, using a deterministic `$anon_distinct_id` that let's us later join it back up to the user. Mental model: `email` is now a mutable user property like `location`, so the events are anonymous at the point where we only know `email`, but it's legitimate to opportunistically join them later
  a. For course registrations (about 60%) where the "PostHog Distinct ID" is populated, use that (lets us connect pre-apply -> apply -> log in activity)
  b. For rows (about 40%) where "PostHog Distinct ID" is missing, use the course registration record id (still lets us connect apply -> log in activity)
2. Create a new event definition in `libraries/computed-posthog-events/src/definitions.ts`, called e.g. 'identify_applicants'. This should look at recently created *users* and recently created *course registrations*, and emit an `identify(user.email, {$anon_distinct_id: the_id_from_the_prev_step}`) when there is a newly logged in user against an existing application, or vice versa. This is then trivial to migrate to the new `identify(user.keycloakIdentifier)` system (#2713), because it is only fired once the user has logged in to the website.

To keep the data continuous, step 2 should be deployed before step 1 (step 2 will be a no-op until step 1 is done)

### ✅ Definition of done

1. The above migration is completed
2. It's possible to verify (with HogQL queries) that we get continuous data for users from logged out browsing -> applying -> logged in browsing

PR #2764, opened 2026-07-09
Add firstLoggedInAt to userTable

Will-Howard:
# Description

`createdAt` on the user table can be different to the first time the user logged in, because we auto-create users via an Airtable automation (when an application is submitted). We can now tell from the existence or not of `keycloakIdentifier` whether the user has actually logged in before. This is useful for #2760, for finding recently created users.

## Issue

#2760

## Developer checklist


libraries/db/src/schema.ts
```diff
@@ -1489,6 +1489,10 @@ export const userTable = pgAirtable('user', {
       pgColumn: text(),
       airtableId: 'fld2AGYp0VLOz3Pg6',
     },
+    firstLoggedInAt: {
+      pgColumn: text(),
+      airtableId: 'fldy6HecnK8seKW9s',
+    },
```
Will-Howard, line 1495: It's safe to do this and the code change in one step as long as we let pg-sync-service finish deploying first (so the column always exists for the website code)

PR #2765, opened 2026-07-09
Don't write empty name string to DB

### LINKED ISSUE #2763: Regression: writing `name: ''` on user creation makes fullName lookups return `[null]`, causing validation warning spam
### 🧩 Summary

**Correction: this was originally diagnosed as a long-standing issue with the `fullName` lookup field. Further investigation showed it's a regression introduced by #2734, deployed in `website/v2.32.3` (2026-07-09 08:31Z).**

Airtable distinguishes between a text field that was *never written* and one that was *explicitly written `""`*. Both read as omitted when fetching the record directly, but a **lookup** exposes the difference:

- never-written name → lookup field omitted from the API response → maps cleanly to `null`
- explicitly-written `""` → lookup returns `[null]` → `airtable-ts` cannot coerce it to `string | null` → validation warning

#2734 added `name: ''` to the user insert in `users.ensureExists` (`apps/website/src/server/routers/users.ts`). Since `v2.32.3`, every new signup creates a user whose `name` cell is in the poisoned explicit-`""` state. When that user registers for FoAI, the `fullName` lookup (`fldsS2lCVlk1WDSDw`) on `self_serve_course_registration` returns `[null]` and fires:

```
Failed to map field fullName (fldsS2lCVlk1WDSDw) from Airtable: Cannot convert value from
airtable type 'multipleLookupValues' to 'string | null', as the Airtable API provided a 'object'.
```

This fires from both `website` (read-back after creating the registration, alerting `#update_tech-prod`) and `pg-sync-service` (every sync of the record, alerting `#update_pg-sync`).

**Evidence:** all registrations linked to nameless users created *before* the `v2.32.3` deploy (11,078 of 15,645 self-serve registrations, dating back to 2024) return an omitted lookup and have never warned. Only users created after 08:31Z today return `[null]` — a clean temporal split at the deploy time.

The `name: ''` was never needed: `db.insert` accepts `Partial`, and the pg `notNull` column is satisfied by the `''` fallback of the string mapper either way.

### ✅ Definition of done

1. Remove `name: ''` from the user insert in `users.ensureExists` (one line).
2. Repair users created since the `v2.32.3` deploy: PATCH `name: null` on their Airtable user records to truly clear the cell, so their registrations stop warning.

### 📌 Notes

- The previously proposed fixes (converting the lookup to a rollup, or wrapping it in a formula) are **not needed** — the field was working correctly for years.
- #2723 (collect name at signup) remains relevant for the blank-certificate problem, but is unrelated to this warning spam.
- Long-term hardening option: `airtable-ts` could coerce all-null arrays to `null` for nullable types, making lookups immune to the explicit-`""` state entirely.

joshestein:
# Description

`website/v2.32.3` (#2734) started writing `name: ''` when creating users. Airtable stores an explicitly-written `""` differently from a never-set field: reading the record looks identical, but lookups of the field return `[null]` instead of being omitted. `airtable-ts` can't coerce `[null]` to `string | null`, so every FoAI registration by a new user fired an Airtable validation warning from both `website` and `pg-sync-service` (see #2763 for the full diagnosis).

The `name: ''` was never needed — `db.insert` accepts partial data, and the real Airtable client maps the created record back with `''` for missing non-nullable string fields, satisfying the pg `notNull` constraint. This PR removes it, restoring the pre-#2734 behaviour.

It also fixes the divergence that likely prompted `name: ''` in the first place: `MockAirtableTs.insert` echoed input as-is instead of applying the type-appropriate defaults the real client produces, so omitting `name` failed in tests while working fine in production. The mock now fills omitted fields the same way airtable-ts does (`''` / `false` / `[]` / `null`).

Synced data is unaffected: nameless users produce `fullName = NULL` in Postgres on both the clean and the warning path. Users created between the `v2.32.3` deploy and this fix still carry the poisoned `""` in Airtable and need a one-off repair (tracked in #2763).

## Issue

Fixes #2763

### DIFF
--- a/libraries/db/src/lib/mock-airtable-ts.ts
+++ b/libraries/db/src/lib/mock-airtable-ts.ts
@@ -10,6 +10,16 @@ function generateRecordId(): string {
   return `rec${Math.random().toString(36).slice(2, 16).padEnd(14, 'a')}`;
 }
 
+function defaultForTsType(tsType: string): unknown {
+  if (tsType.includes('| null')) return null;
+  if (tsType.endsWith('[]')) return [];
+  if (tsType === 'string') return '';
+  if (tsType === 'boolean') return false;
+  // Matches airtable-ts, which throws when a missing value can't be coerced to
+  // the declared non-nullable type (e.g. 'number').
+  throw new Error(`MockAirtableTs: no value provided for non-nullable field of type '${tsType}'.`);
+}
+
 /**
  * Test mock of AirtableTs.
  *
@@ -66,10 +76,20 @@ export class MockAirtableTs extends AirtableTs {
     throw new Error('MockAirtableTs does not support scan(). PgAirtableDb.scan() reads from Postgres directly, so this method should not be called.');
   }
 
-  override async insert<T extends Item>(_table: Table<T>, data: Partial<Omit<T, 'id'>>): Promise<T> {
+  override async insert<T extends Item>(table: Table<T>, data: Partial<Omit<T, 'id'>>): Promise<T> {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const id = (data as any).id ?? generateRecordId();
-    return { ...data, id } as T;
+    // Real Airtable omits unset fields in the created record, and airtable-ts maps
+    // them to type-appropriate defaults ('' / false / [] / null). Mirror that here so
+    // inserts that omit non-nullable fields behave the same as in production.
+    const defaults: Record<string, unknown> = {};
+    for (const [field, tsType] of Object.entries<string>(table.schema)) {
+      if (!(field in data)) {
+        defaults[field] = defaultForTsType(tsType);
+      }
+    }
+
+    return { ...defaults, ...data, id } as T;
   }
 
   override async update<T extends Item>(table: Table<T>, data: Partial<T> & { id: string }): Promise<T> {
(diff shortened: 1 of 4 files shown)


joshestein:
@claude review

joshestein:
@Will-Howard I'm going to merge this to stop our prod channel being spammed. Happy to revise / revert if you have any callouts.

Will-Howard:
> I'm going to merge this to stop our prod channel being spammed

Ok sounds good 👍. I'll review it now anyway

libraries/db/src/lib/mock-airtable-ts.ts
```diff
@@ -10,6 +10,16 @@ function generateRecordId(): string {
...
 
+function defaultForTsType(tsType: string): unknown {
+  if (tsType.includes('| null')) return null;
+  if (tsType.endsWith('[]')) return [];
+  if (tsType === 'string') return '';
+  if (tsType === 'boolean') return false;
+  // Matches airtable-ts, which throws when a missing value can't be coerced to
+  // the declared non-nullable type (e.g. 'number').
+  throw new Error(`MockAirtableTs: no value provided for non-nullable field of type '${tsType}'.`);
+}
```
Will-Howard, line 21: This is the only bit I'm slightly concerned about, this could drift from what's in `airtable-ts`. But I think it's better than the status quo so happy to keep it 👍

Will-Howard:
[Retroactively approved]

PR #2780, opened 2026-07-14
Update Airtable PAT

### LINKED ISSUE #2759: pg-sync-service: Redact Authorization header before logging
The `winston` logger in pg-sync-service dumps axios error objects, which include the `Authorization` header if it is present. We should sanitise it before logging, and rotate any credentials that have already appeared.

joshestein:
# Description

In https://github.com/bluedotimpact/bluedot/pull/2761 we redacted the Airtable PAT from our logs. This is a follow-up to rotate the PAT. The new PAT was generated by Li-lian, available in 1Pass.

## Issue

Fixes #2759


joshestein:
While validating it seems the token is missing `webhook:manage` scopes. I've pinged about this, will update / merge once I have more certainty about model scope.

joshestein:
webhook:manage scope has now been added

PR #2782, opened 2026-07-14
Use Google login user's name

### LINKED ISSUE #2723: Require user's full name during sign up flow
### 🧩 Summary

Multiple users have complained via bug reports that their names are not being shown on their FoAI certificates. During sign up, we only require an email. This leads to an empty space on the certificate where their name would be shown.

Users need to manually update their name via their profile settings. After they've done this, there is a delay where our webhooks sync data.

### ✅ Definition of done

We need a way to preemptively get users to submit their names. We could:

1. Change the sign up form to take a name input
2. Warn the users on the certificate page if there is no name
3. Prompt the users for their name at some point (after signing in)

I lean towards (1); this would require modifying bluedot-keycloak-theme, getBluedotKeycloakTheme.sh, Login.tsx and users.ts. This will only fix the issue for future sign ups. For currently signed up users we could consider some sort of in-app name capture - I don't feel strongly about this.

I also suggest that if users sign in with Google we automatically extract their name - there is no cost (aside from additional permissions).

### 📌 Notes / Designs

_No response_

joshestein:
# Description

Populate a user's name from their Google login.

Previously, signing in with Google only stored email, Keycloak ID, and timestamps. The name had to be set manually via the profile settings editor. Now the name from the Google/OIDC token is written automatically on login.

## Issue

Fixes #2723

## Developer checklist

### DIFF
--- a/apps/website/src/server/routers/users.test.ts
+++ b/apps/website/src/server/routers/users.test.ts
@@ -165,6 +165,63 @@ describe('users.ensureExists', () => {
     insertSpy.mockRestore();
   });
 
+  test('writes the name from the token when creating a new user', async () => {
+    vi.mocked(loginPresets.keycloak.verifyAndDecodeToken).mockResolvedValue({
+      sub: 'test-sub',
+      email: 'test@example.com',
+      name: 'John Doe',
+      iss: 'test-issuer',
+      aud: 'test-audience',
+      exp: Math.floor(Date.now() / 1000) + 3600,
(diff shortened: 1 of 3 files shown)


joshestein:
@Will-Howard can you add your eyes here 👀 ? This is my first time touching keycloak infrastructure, would appreciate your guidance 🙌

Will-Howard (approved):
Looks good! This was a lot simpler than I expected

Issue #2459, opened 2026-05-11
Deprecate email as a user identifier


Will-Howard:
### 🧩 Summary

Per the [comment below](https://github.com/bluedotimpact/bluedot/issues/2459#issuecomment-4833948583), we rely on the assumption of a fixed email as the primary user identifier in these places:
1. In code, as a foreign key and the primary key to get-or-create a user by when they log in
2. In customer.io, which is what we use to send almost all emails to users
3. In Keycloak (where we want to keep login-by-email, but enable the `update-email` setting)
4. In PostHog, we use email as the `distinct_id` for users
5. In the [CRM base](https://airtable.com/apppOzz9fPg59PxLa/tblMYYK8bL2fRJmv7/viwQybZvpyq8NsPOf?blocks=hide)
6. In automations throughout other bases (mostly sending things to customer.io, but some other cases)

We want to migrate all of these to handle potentially changing emails (brief summary, more detail will be in sub-issues etc):
1. In code: We can use the user id (i.e. the `id` of the User record in the Applications base) as the foreign key. For the `auth` object itself and the initial get-or-create of the User, we can use the `sub` property that comes from Keycloak.
2. In customer.io: Roughly speaking, we can run an [Identify](https://docs.customer.io/integrations/data-in/source-spec/identify-spec/) API call when someone updates their email, and then leave the existing automations as-is (which will pick up the new email, and a continuous history for the user will be preserved in customer.io). There are some edge cases here: emails relating to a specific application form (where we should keep the form email), and newsletter subscribers
3. In Keycloak: Enable the `update-email` flow (verify-new-then-swap) so email becomes mutable, and treat the immutable `sub` as the identity.
4. In PostHog: Use the stable user id as the `distinct_id` and set the email as a person property (so people stay searchable by email).
5. In the CRM base: Keep Primary email as the human lookup key — the CRM also holds people with no account, so email has to stay the key here — but make matching dedupe on ANY known email (not just the current primary), and carry the stable user id on the Person where possible.
6. In automations throughout other bases: Update on a case my case basis depending on what they do

I'll break this down into sub-issues which plan out the individual changes more concretely and what-blocks-what.

### ✅ Definition of done

The above consumers of email are migrated to support potentially changing emails

Will-Howard:
[...]

PR #2794, opened 2026-07-16
Remove stray non-user email reads from website/ code

Will-Howard:
# Description

Removes the last two `apps/website` reads of `meetPersonTable.email` and `selfServeCourseRegistrationTable.email`, ahead of moving those columns to `deprecatedColumns` as part of the email-as-identifier deprecation. These were missed in #2721 because they aren't used to _look up_ the user.

- `facilitators.getFeedbackFormData` returned `meetPerson.email` to prefill the facilitator invoice form. It now resolves the email from the canonical `user` table via `meetPerson.userId`
- The my-bluedot self-serve (Future of AI) rows forwarded `reg.email` to the frontend. This can only be consumed by the availability-form CTA, which doesn't apply for self-serve registrations. I changed this to return `email: ''` (workaround, to avoid complicated type plumbing). I also left a [comment](https://github.com/bluedotimpact/bluedot/issues/2459#issuecomment-4993161152) on the overarching issue to review how the availability form interacts with the email -> sub change

## Issue

#2709

## Developer checklist

### DIFF
--- a/apps/website/src/components/my-courses/useCourseListRow.tsx
+++ b/apps/website/src/components/my-courses/useCourseListRow.tsx
@@ -325,7 +325,7 @@ const getParticipantActions = (
     },
     {
       id: 'availability',
-      isVisible: state === 'upcoming' && courseRegistration.decision !== 'Reject',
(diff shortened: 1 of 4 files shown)


Will-Howard:
@greptileai review @coderabbitai review

PR #2802, opened 2026-07-18
Remove writes to `email` fields on `resourceCompletion` and `exerciseResponse` tables

Will-Howard:
# Description

More cleanup following the migration from email to `keycloakIdentifier` as the primary identifier for users.

## Issue

#2709

## Developer checklist

### DIFF
--- a/libraries/db/src/schema.ts
+++ b/libraries/db/src/schema.ts
@@ -147,7 +147,7 @@ export const courseTable = pgAirtable('course', {
 export const exerciseResponsePgTable = deprecationSafePgTable('exercise_response', {
   columns: {
     id: text('id').primaryKey().default(sql`gen_random_uuid()::text`),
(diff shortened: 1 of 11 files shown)


libraries/db/src/schema.ts
```diff
@@ -147,7 +147,7 @@ export const courseTable = pgAirtable('course', {
 export const exerciseResponsePgTable = deprecationSafePgTable('exercise_response', {
   columns: {
     id: text('id').primaryKey().default(sql`gen_random_uuid()::text`),
-    email: text().notNull(),
+    email: text(),
```
Will-Howard, line 150: I'm removing the `notNull()` in this PR, to then deprecate it in a followup PR. I'm separating these to avoid any risk of drizzle/pg-sync-service getting wires crossed or entering interactive mode

PR #2821, opened 2026-07-23
Reuse `seedLoggedInUser` utility

### LINKED ISSUE #2756: Standardize router-test user seeding with seedLoggedInUser()
### 🧩 Summary

In the stack of PRs to migrate user lookups from `email` to `keycloakIdentifier` (#2721, #2734), we've ended up with a mix of `seedLoggedInUser()` and manual setup of users in tests. We should standardise the remaining manual cases to use `seedLoggedInUser()`. I'm splitting this out into a separate issue just to reduce the scope of those PRs.

### ✅ Definition of done

Tests that do standard user setup use `seedLoggedInUser()` rather than manual setup.

### 📌 Notes / Designs

_No response_

joshestein:
# Description

- `seedLoggedInUser()` now takes an optional `Partial<User>` overrides arg
- Converted the remaining hand-rolled caller rows:
  - `CourseListRow.test.tsx` — `seedFacilitator`/`seedParticipant` → `seedLoggedInUser({ id: 'user-1' })`
  - `ImpersonationBadge.test.tsx` — admin callers → `seedLoggedInUser({ isAdmin: true })`
  - `ProfileNameEditor.test.tsx` — manual insert → `seedLoggedInUser()`

Deliberately left manual:
- `users.test.ts` — the user row *is* the subject under test (`users.ensureExists` creates /
  backfills it), so seeding a fully-formed row would defeat the assertions.
- `admin.test.ts` / `context.test.ts` — callers use non-standard identities (`admin-sub`,
  `scoped-sub`), not the `test-sub` caller `seedLoggedInUser()` represents; routing them
  through the helper would mean overriding every field.

## Issue

Fixes #2756

### DIFF
--- a/apps/website/src/__tests__/dbTestUtils.tsx
+++ b/apps/website/src/__tests__/dbTestUtils.tsx
@@ -4,7 +4,7 @@ import { unstable_localLink } from '@trpc/client';
 import { createTRPCReact } from '@trpc/react-query';
 import { useState } from 'react';
 import {
-  pushTestSchema, resetTestDb, type TestPgAirtableDb, userTable,
+  pushTestSchema, resetTestDb, type TestPgAirtableDb, type User, userTable,
 } from '@bluedot/db';
 import type { AppRouter } from '../server/routers/_app';
 import type { Context } from '../server/context';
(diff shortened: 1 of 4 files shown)


joshestein:
@Will-Howard the issue asked for updating 'router-test user seeding', but I haven't actually touched any router tests here (some were deliberately excluded, see PR description). Not sure if there are other places I've left out you think would benefit from reuse of this utility?

Will-Howard (approved):
@joshestein Nope I don't think there's anything else worth converting! I agree on the ones you've left as manual inserts

Issue #2810, opened 2026-07-19
Support changing emails in customer.io


Will-Howard:
### 🧩 Summary

There are many automations in Airtable which submit to customer.io using an email field that exists on the recording question. We are working towards deprecating many of these email fields (#2459), and ultimately supporting changing the primary email on the user row, so we may need to change these automations.

Proposed general strategy: Keep most of the automations largely as-is, but make sure we're using the authoritative user.email field. Carefully migrate the few (such as [this one](https://airtable.com/apppOzz9fPg59PxLa/wflp8Fk3qpzCCukmX/wac1ZS6MXbjKapik1) in the CRM base) that set user-level properties.

### ✅ Definition of done

We support changing emails in customer.io, or at least have a plan to do so once we implement the overall email-change feature (#2570).

Will-Howard:
## Migration plan

At a high level, there are a few automations in Airtable that do identity-level things in customer.io (near-exhaustive list: [1](https://airtable.com/appnJbsG1eWbAdEvf/wflVpvIGT3ZfnkuF9/wacI0mg7fEnwNGqHD), [2](https://airtable.com/apppOzz9fPg59PxLa/wflp8Fk3qpzCCukmX/wac1ZS6MXbjKapik1), [3](https://airtable.com/apppOzz9fPg59PxLa/wflU14dAsziJDFGu4/wacwShZjZl2RbAEs8)), but most just trigger sending emails via a standardised `sendDataToUniversalWebhook` webhook (e.g. [this automation](https://airtable.com/appnJbsG1eWbAdEvf/wflOy3DWC5q7SLQdM/waci48HHWE8bjupge)).

With this in mind, these are the migration steps required to support changing emails in customer.io (keeping all history attached to the same consistent person). The general idea is to set the Applications User id as the `id` in customer.io, overriding the current email-based profiles. In cases where this depends on subtle behaviour of customer.io, I have tested with fake users how it behaves:
- [x] In the website email-change code (#2570, tbd) add an `updateCustomerIoEmail(userId, oldEmail, newEmail)` side effect. This should (async, not blocking the user flow). _Edit: In-progress as #2825._:
  a. Check if any profile exists for the old email. If not, do nothing (the new email will be naturally picked up)
  b. If any profile exists with the old email, ensure the userId-based-profile exists: look up by id; if absent, identify({id}, {email: old})
  c. Run `identify({id}, {email: new})` to update the email on the existing profile
[...]

Issue #2570, opened 2026-05-28
Allow users to change emails


Will-Howard:
### 🧩 Summary

Some users have requested to change their account email. Currently we can do this manually, but we want to add a self-serve version.

This is blocked by #2459 (decoupling emails from everything apart from User). It also needs some scoping (best done after #2459 is finished), covering for instance:
1. How to securely verify the old and new email
2. The UI for doing this
3. Whether to roll "support multiple emails" into this issue (can make (1) easier by making the workflow add-new -> delete-old, which is lower stakes than switching)

### ✅ Definition of done

Users are able to change their account emails with no admin intervention

Will-Howard:
Case [flagged by CodeRabbit](https://github.com/bluedotimpact/bluedot/pull/2742#discussion_r3530768371) that I'm tracking here: What should we do if we get auth back from Keycloak (containing `email` and `sub`), and `sub` matches a user, but the user's email doesn't match `auth.email`.

This would effectively mean the email has been changed outside of the endorsed flow in some way. We need to decide how to handle this based on what the flow ultimately looks like.

Edit 2026-07-29: I've decided the correct behaviour here is to proceed normally in this case, i.e. require one email to sign in, and send account emails to another emails. It's an odd case to get into, and shouldn't be possible, but there's nothing inherently wrong with it from a security or logical perspective.

Will-Howard:
Another case I've realised is worth handling explicitly:
1. Email A applies to a course, this creates a User row as a side effect, but without any login info (no Keycloak identifier)
2. Email B signs up to the course hub
3. The account with email B later tries to change their email to email A

With the email-change flow as written (#2828, #2830), the user would get an "Email already taken" error when trying to trigger switching to the new email. This is unhelpful in the case (which is quite common) where someone applies under one email and signs up under another, or vice versa.

The main problem to solve here is what to do with any applications that use email A. Suggested solution:
- Allow the user to trigger an email change request from email B -> email A
- On the confirm screen, check if there are any applications registered against email A. If so, prompt the user to either adopt them or drop them

Reasoning:
- We already auto-link applications to the person who owns email A (on signup). By switching the email, the user is verifying that they own email A, so it's correct to allow them to adopt the applications
- It would be surprising for random applications (which may be old) to appear in your account without a prompt to accept them, so we should prompt the user somewhere (and allow them to not adopt old cruft)
- It straightforward to do this in the `/confirm-email-change` page, and there's lots of UI space to display it to the user

[...]

PR #2825, opened 2026-07-24
Add customer.io email-change primitives (unused, tested)

Will-Howard:
# Description

Part of the work to allow users to change their email (#2570), specifically making customer.io handle this correctly (#2810).

This PR adds `updateCustomerIoEmail({ userId, oldEmail, newEmail })`, which tries to keep a stable customer.io profile pinned to the userId, and change the email from `oldEmail` to `newEmail`. There are edge cases where e.g. the `newEmail` already exists in customer.io (which can happen if a user applied to ~any BlueDot thing under the new email already), it's best to look at the tests to see how that is handled in each case.

I also added `sendEmailChangeVerification` in this PR, in order to meaningfully test more of `updateCustomerIoEmail`, but `sendEmailChangeVerification` isn't yet used by anything in this PR. It will be adopted by the actual email-change flow when that exists.

I ran some tests against the prod environment in preparing this branch, per the comment in the test file you can review these tests here:
> These semantics were validated against the real production workspace with a live-fire
> model-based test, which is preserved on the `wh-2810-cio-livetest-2026-07-archive` git branch,
> file `apps/website/src/lib/api/customerio.prod.test.ts`.
I'm not committing these tests themselves because they're not practical to maintain (they create and delete real users in customer.io every time).

## Issue

#2810

## Developer checklist

### DIFF
--- a/apps/website/src/lib/api/customerio.test.ts
+++ b/apps/website/src/lib/api/customerio.test.ts
@@ -313,7 +313,7 @@ describe('sendEmailChangeVerification', () => {
     const fetchMock = vi.fn(async () => jsonResponse(200, { delivery_id: 'd1' }));
     vi.stubGlobal('fetch', fetchMock);
 
-    await sendEmailChangeVerification({ oldEmail: '  Old@Example.COM ', newEmail: ' New@Example.com ', confirmUrl: 'http://localhost:8000/account/confirm-email-change?token=abc' });
+    await sendEmailChangeVerification({ oldEmail: '  Old@Example.COM ', newEmail: ' New@Example.com ', confirmUrl: 'http://localhost:8000/account/confirm-email-change?token=abc&uid=u1' });
 
     expect(fetchMock).toHaveBeenCalledTimes(1);
     const [url, init] = fetchMock.mock.calls[0]! as unknown as [string, { method: string; body: string }];
@@ -323,7 +323,8 @@ describe('sendEmailChangeVerification', () => {
     expect(body.to).toBe('new@example.com');
     expect(body.identifiers).toEqual({ email: 'old@example.com' });
     expect(body.send_to_unsubscribed).toBe(true);
-    expect(body.body).toContain('http://localhost:8000/account/confirm-email-change?token=abc');
+    expect(body.body).toContain('http://localhost:8000/account/confirm-email-change?token=abc&#38;uid=u1');
+    expect(body.body).not.toContain('token=abc&uid=u1');
     expect(body.subject).toBeTruthy();
   });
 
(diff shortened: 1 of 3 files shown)


Will-Howard:
@coderabbitai review @greptileai review

Will-Howard:
> Up to ~60s of synchronous blocking per call, once wired into a request path.

On this high level comment: This is intended for use as a non-blocking side effect after email change succeeds, so it's ok for it to take some time

apps/website/src/lib/api/customerio.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (!res.ok) throw new Error(`customer.io identify failed: HTTP ${res.status}`);
+}
+
+const profileEmail = (profile: CioProfile | null): string | null => {
+  const email = profile?.identifiers?.email ?? profile?.attributes?.email;
+  return typeof email === 'string' ? normaliseEmail(email) : null;
+};
+
+async function emailIsVisible({ userId, email }: { userId: string; email: string }): Promise<boolean> {
+  for (let attempt = 0; attempt < VERIFY_POLL_ATTEMPTS; attempt++) {
```
marn-in-prod, line 69: Nit: the loop retries when the write isn't possible, but not when getProfileById throws. Is it intentional that a single throw will stop the loop?

apps/website/src/lib/api/customerio.test.ts
```diff
@@ -0,0 +1,353 @@
...
+};
+
+const makeFakeCio = (profiles: FakeProfile[], { applyWrites = true }: { applyWrites?: boolean } = {}): FakeCio => ({
+  profiles,
+  applyWrites,
+  identifyCalls: [],
+});
+
+const applyIdentify = (cio: FakeCio, id: string, email: string) => {
+  const owner = cio.profiles.find((p) => p.email === email);
```
marn-in-prod: Nit: If cio is email case-insensitive IRL then we should make the test also behave that way, right? Plus we should also add a test where the stored email isn't all lowercase

marn-in-prod (approved):
Generally looks good! Comments are about edge cases and me wanting clarity on what you intended

apps/website/src/lib/api/customerio.ts
Will-Howard: > Is it intentional that a single throw will stop the loop?

Not intentional, I'll add a `catch` in here. It should be safe to retry `getProfileById` since it's just a read

apps/website/src/lib/api/customerio.test.ts
Will-Howard: I've updated `applyIdentify` to do everything case-insensitively, and added a test before for the case where the email stored on the user is in mixed case.

I also checked how CIO itself works, and it converts the email to lower case before writing it (so any email you read back will always be lowercase), the tests are now consistent with that

PR #2828, opened 2026-07-24
[feat] Admin-initiated email change flow

Will-Howard:
# Description

_Blocked by #2810._

Adds support for changing emails, initiated via an admin-only UI. I'm building this ahead of the user-facing UI to be able to test it end-to-end in production before exposing it to users. Once the user-facing UI exists, I'm planning to leave this in place as I expect we'll still get some email-change requests over support (e.g. when the user has lost access to the old email and can't log in).

Places for reviewers to focus:
- The screenshots below show the help text that will display in different cases (e.g. existing Google-and-email/password, existing Google-only). You might want to go through the flow end to end with these different cases
- The two endpoints `requestEmailChange` and `confirmEmailChange`, especially on security, and on correctness in the case of errors (e.g. the Keycloak -> Airtable -> customer.io sequence of changes)

## Issue

#2570

## Developer checklist

## Screenshot

| 📸 | After |
|---------|---|
| 🖥️ | Screenshot 2026-07-27 at 11 43 08 |
| | Screenshot 2026-07-27 at 11 43 37 |
| | Screenshot 2026-07-27 at 11 43 41 |
| | Screenshot 2026-07-27 at 11 10 28 |
| | Screenshot 2026-07-27 at 11 10 41 |
| | Screenshot 2026-07-27 at 11 13 29 |
| | Screenshot 2026-07-27 at 11 13 52 |
| | Screenshot 2026-07-27 at 11 14 19 |
| | Screenshot 2026-07-27 at 11 31 03 |
| | Case where changing the email succeeds but there is an error in attempting to unlink an old Google account: Screenshot 2026-07-27 at 11 35 30 |
| | Screenshot 2026-07-27 at 11 35 54 |


apps/website/package.json
```diff
@@ -54,6 +54,7 @@
     "clsx": "^2.1.1",
     "framer-motion": "^12.23.12",
     "http-errors": "^2.0.0",
+    "jose": "^6.2.4",
```
Will-Howard, line 57: Using [jose](https://github.com/panva/jose) over `jsonwebtoken` (which is used in the meet app) because it is more modern and doesn't introduce transitive dependencies

apps/website/src/server/routers/users.ts
```diff
@@ -1,15 +1,46 @@
...
+
+  return existing.rows.length > 0;
+}
+
+async function unlinkStaleGoogleIdentitiesOrAlert(userId: string, keycloakIdentifier: string, newEmail: string): Promise<{ loginMethods: LoginMethods | null }> {
+  try {
+    return { loginMethods: await unlinkStaleGoogleIdentities(keycloakIdentifier, newEmail) };
+  } catch (error) {
+    await slackAlert(env, [`[EmailChange] Failed to unlink stale google identities for user ${userId}: ${error instanceof Error ? error.message : String(error)}`]);
+    return { loginMethods: null };
```
marn-in-prod: I understand why this returns normally. Rolling back changes across Keycloak and the user table is hard, and the rollback could also fail. With that said, could we add a code comment that documents the partial-success behavior? Could the Slack alert also say that the email changed but the old Google account might remain linked and require manual cleanup? 

So the slack alert says something like:
```text
Email changed successfully, but stale Google identity cleanup failed.
The old Google account may still be able to sign in.
Check the Keycloak user and remove the stale Google identity manually.
```

Separately, what do you think of retries before sending the slack alert?

marn-in-prod: Maybe worth adding a test for this scenario

marn-in-prod (approved):
LGTM. One real comment that is mostly a nit after I wrapped my head around the new code

apps/website/src/server/routers/users.ts
Will-Howard: All seems reasonable 👍. I've updated the alert text, added retries, and a test that asserts the alert is emitted

apps/website/src/components/analytics/PostHogProvider.tsx
```diff
@@ -27,6 +27,10 @@ const ActivePostHogProvider: React.FC<React.PropsWithChildren> = ({ children })
         }
       },
       persistence: useConsentStore.getState().isConsented ? 'localStorage+cookie' : 'memory',
+      sanitize_properties: (properties) => Object.fromEntries(Object.entries(properties).map(([key, value]) => [
+        key,
+        typeof value === 'string' ? value.replace(/([?&])token=[^&#]+/g, '$1token=redacted') : value,
```
joshestein: ❤️

Will-Howard:
Counting this as ready now, just waiting for #2810 then will merge

apps/website/src/components/analytics/PostHogProvider.tsx
```diff
@@ -27,6 +27,10 @@ const ActivePostHogProvider: React.FC<React.PropsWithChildren> = ({ children })
         }
       },
       persistence: useConsentStore.getState().isConsented ? 'localStorage+cookie' : 'memory',
+      sanitize_properties: (properties) => Object.fromEntries(Object.entries(properties).map(([key, value]) => [
```
joshestein: Not sure about this but the docs for redaction seem to use `before_send`, not `sanitize_properties`: https://posthog.com/docs/libraries/js/usage#redacting-information-in-events

joshestein:
> Counting this as ready now, just waiting for #2810 then will merge

one sec

apps/website/src/pages/account/confirm-email-change.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  if (loginMethods.hasPassword) {
+    return <P>Log in with your new email address and your existing password.</P>;
+  }
+
+  return (
+    <P>
+      You used to sign in with Google, so there is no password set for this account. To log in, sign
+      with Google from your new email address and click "Add to existing account". Alternatively, choose
```
joshestein: nit: 'To log in, sign [in] with Google' (missing 'in')

apps/website/src/server/routers/users.ts
```diff
@@ -1,15 +1,57 @@
...
+const getEmailChangeConfirmUrl = (token: string) => {
+  const siteUrl = process.env.NEXT_PUBLIC_SITE_URL ?? 'https://bluedot.org';
+  return `${siteUrl}${ROUTES.confirmEmailChange.url}?token=${encodeURIComponent(token)}`;
+};
+
+async function isEmailTakenByAnotherUser(email: string, excludeUserId: string): Promise<boolean> {
+  const existing = await db.pg.execute(sql`SELECT id FROM ${userTable.pg} WHERE LOWER(TRIM(email)) = ${email} AND id != ${excludeUserId} LIMIT 1`);
+
+  return existing.rows.length > 0;
+}
```
joshestein: We only check DB - do we not also need to check Keycloak itself?

apps/website/src/lib/api/keycloak.ts
```diff
@@ -110,6 +111,82 @@ export async function updateKeycloakPassword(
...
+
+// The customers realm uses registrationEmailAsUsername, so the username tracks the email.
+export async function updateKeycloakEmail(userSub: string, newEmail: string): Promise<void> {
+  const user = await adminRequest<Record<string, unknown>>({ method: 'get', path: `/users/${userSub}` });
+
+  await adminRequest({
+    method: 'put',
+    path: `/users/${userSub}`,
+    data: {
+      ...user,
```
joshestein, line 167: nit: do we need this? or can we directly call the 'put' request without first getting the user?

apps/website/src/lib/api/keycloak.ts
```diff
@@ -110,6 +111,82 @@ export async function updateKeycloakPassword(
   }
 }
 
+type KeycloakFederatedIdentity = {
+  identityProvider: string;
+  userId: string;
+  userName: string;
+};
+
+async function adminRequest<T>(config: { method: 'get' | 'put' | 'delete'; path: string; data?: unknown }): Promise<T> {
```
joshestein: Suggest removing duplication between this and `updateKeycloakPassword` in a follow-up PR

joshestein (commented):
My only real question is about not checking Keycloak for existing email when calling `requestEmailChange`❓

apps/website/src/server/routers/users.ts
```diff
@@ -129,4 +171,81 @@ export const usersRouter = router({
...
+        throw new TRPCError({ code: 'BAD_REQUEST', message: `Email change failed: ${error instanceof Error ? error.message : String(error)}`, cause: error });
+      }
+
+      try {
+        await db.update(userTable, { id: user.id, email: payload.newEmail });
+      } catch (error) {
+        await slackAlert(env, [`[EmailChange] Keycloak now has ${payload.newEmail} for user ${user.id} but the user table update failed and still has ${payload.oldEmail}: ${error instanceof Error ? error.message : String(error)}`]);
+        throw error;
+      }
+
```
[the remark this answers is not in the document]
Will-Howard: Technically correct, but there's not much we can do about this

apps/website/src/server/routers/users.ts
```diff
@@ -1,15 +1,57 @@
...
+    } catch (error) {
+      lastError = error;
+    }
+  }
+
+  await slackAlert(env, [
+    `[EmailChange] Email for user ${userId} changed successfully to ${newEmail}, but stale Google identity cleanup failed after 3 attempts: ${lastError instanceof Error ? lastError.message : String(lastError)}. The old Google account may still be able to sign in. Check the Keycloak user and remove the stale Google identity manually.`,
+  ]);
+  return { loginMethods: null };
+}
```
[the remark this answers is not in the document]
Will-Howard: The decision to unlink the Google account was made for product reasons (to not confuse the user), rather than primarily for security reasons.

There is a security angle in the case where a. they want to change actual ownership of the account (e.g. when one employee leaves and another takes over), or b. the Google account is compromised in some way. These are both quite niche cases, that are unlikely to apply in practice to the course platform.

apps/website/src/components/analytics/PostHogProvider.tsx
Will-Howard: Oh yep you're right, good catch, `sanitize_properties` is being deprecated. I'll change it to `before_send`

apps/website/src/lib/api/keycloak.ts
Will-Howard: Generally `PUT` means "overwrite the entire object", as opposed to `PATCH` which means "update specific fields". I think it's safer/more readable to pass the whole correct object if we're doing a `PUT` request (regardless of whether under the hood keycloak actually merges vs overwriting)

apps/website/src/server/routers/users.ts
Will-Howard: Yep checking Keycloak would be more correct (it should be in sync, but could get out of sync due to admin edits/errors), I've updated this to check both.

Also per [this comment](https://github.com/bluedotimpact/bluedot/issues/2570#issuecomment-4895694467), after reviewing the behaviour related to this I decided to generally try to proceed gracefully when Keycloak and the user email diverge, because it can happen and needn't be a halt-and-catch-fire error

apps/website/src/server/routers/users.ts
```diff
@@ -129,4 +171,81 @@ export const usersRouter = router({
...
+
+      if (await isEmailTakenByAnotherUser(payload.newEmail, user.id)) {
+        throw new TRPCError({ code: 'CONFLICT', message: 'Another account already has this email' });
+      }
+
+      try {
+        await updateKeycloakEmail(user.keycloakIdentifier, payload.newEmail);
+      } catch (error) {
+        throw new TRPCError({ code: 'BAD_REQUEST', message: `Email change failed: ${error instanceof Error ? error.message : String(error)}`, cause: error });
+      }
```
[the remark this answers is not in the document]
Will-Howard: Fixed

apps/website/src/server/routers/users.ts
```diff
@@ -129,4 +171,81 @@ export const usersRouter = router({
...
+
+      if (await isEmailTakenByAnotherUser(input.newEmail, user.id)) {
+        throw new TRPCError({ code: 'CONFLICT', message: 'Another user already has this email' });
+      }
+
+      const token = await createEmailChangeToken({ userId: user.id, oldEmail: user.email, newEmail: input.newEmail });
+      await sendEmailChangeVerification({ oldEmail: user.email, newEmail: input.newEmail, confirmUrl: getEmailChangeConfirmUrl(token) });
+
+      logger.info(`[EmailChange] admin ${impersonationRealIdentity(ctx).sub} requested email change for user ${user.id}: ${user.email} -> ${input.newEmail}`);
```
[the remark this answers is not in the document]
Will-Howard: We already do this for user-impersonation logs, and having this as a breadcrumb could be quite useful to trace bugs in future. Given that logs are only admin-accessible anyway, I think this is worth the tradeoff

apps/website/src/lib/api/emailChangeToken.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  return new TextEncoder().encode(env.EMAIL_CHANGE_TOKEN_SECRET);
+};
+
+export async function createEmailChangeToken({ userId, oldEmail, newEmail }: EmailChangePayload): Promise<string> {
+  return new SignJWT({ userId, oldEmail: normaliseEmail(oldEmail), newEmail: normaliseEmail(newEmail) })
+    .setProtectedHeader({ alg: 'HS256' })
+    .setExpirationTime('48h')
+    .sign(getSecret());
```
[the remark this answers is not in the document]
Will-Howard: I changed it to encrypt rather than just signing with the secret so that the emails can't be read

PR #2830, opened 2026-07-28
[feat] User-initiated email change flow

### LINKED ISSUE #2570: Allow users to change emails

Will-Howard:
# Description

Builds on the admin-initiated version of the change-email flow (#2828), the code here should be relatively self explanatory given that. The UI is the more important thing to focus on in review.

## Issue

Fixes #2570 

## Developer checklist

### Account page & change email modal

| State | 🖥️ Desktop | 📱 Mobile |
|---|---|---|
| Account page |  |  |
| Change email modal |  |  |
| Invalid email (client-side) |  |  |
| In flight |  |  |
| Confirmation link sent |  |  |
| Email already in use |  |  |
| Unexpected error |  |  |

### Confirm page

_These screenshots are actually taken from #2828._

| State | 🖥️ Desktop |
|---|---|
| Confirm page |  |
| Success (password account) |  |
| Success (Google account) |  |
| Success (Google + password) |  |
| Success (Google-only, no password set) |  |
| Success (stale Google identity cleanup failed) |  |
| Error: link expired |  |
| Error: link no longer valid |  |


apps/website/src/components/settings/AccountSettingsSection.tsx
```diff
@@ -26,16 +28,25 @@ const PasswordSection = () => {
...
-      {/* Password Section */}
+      {/* Account Settings Section */}
       <div className="mb-6">
-        <P className="font-semibold mb-2">Password</P>
-        <CTALinkOrButton
-          variant="secondary"
-          onClick={() => setShowChangePasswordModal(true)}
-          aria-label="Change password"
-        >
-          Change Password
```
Will-Howard, line 37: "Change Password" -> "Change password". From looking at recent changes I think sentence case is the direction we're going in, though it's not completely consistent

Will-Howard:
@greptileai review @coderabbitai review

apps/website/src/server/routers/users.ts
```diff
@@ -32,6 +32,28 @@ async function isEmailTakenByAnotherUser(email: string, excludeUserId: string):
...
+    throw new TRPCError({ code: 'BAD_REQUEST', message: 'New email is the same as the current email' });
+  }
+
+  if (await isEmailTakenByAnotherUser(newEmail, user.id)) {
+    throw new TRPCError({ code: 'CONFLICT', message: 'Another user already has this email' });
+  }
+
+  const token = await createEmailChangeToken({ userId: user.id, oldEmail: user.email, newEmail });
+  await sendEmailChangeVerification({ oldEmail: user.email, newEmail, confirmUrl: getEmailChangeConfirmUrl(token) });
+}
```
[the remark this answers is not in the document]
Will-Howard: The email is already normalised within `sendEmailChangeVerification`. I've just added a test `'the flow works when the stored email is not normalised'` to assert this rather than normalising it again.

apps/website/src/server/routers/users.ts
```diff
@@ -175,30 +197,33 @@ export const usersRouter = router({
...
+    .mutation(async ({ input, ctx }) => {
+      if (ctx.impersonation) {
+        throw new TRPCError({ code: 'BAD_REQUEST', message: 'Cannot change email when impersonating another user' });
       }
 
-      const token = await createEmailChangeToken({ userId: user.id, oldEmail: user.email, newEmail: input.newEmail });
-      await sendEmailChangeVerification({ oldEmail: user.email, newEmail: input.newEmail, confirmUrl: getEmailChangeConfirmUrl(token) });
+      const user = await getUserFromAuthOrThrow(ctx.auth);
+
+      await sendEmailChangeConfirmation(user, input.newEmail);
```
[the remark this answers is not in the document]
Will-Howard: I think this is quite useful to have, so I added a rate limit (3 attempts in 30 mins) rather than removing it entirely. In practice, people can enumerate emails via the Keycloak signup form already, so this doesn't make the situation much worse

apps/website/src/components/settings/AccountSettingsSection.tsx
```diff
@@ -228,7 +246,7 @@ const ChangePasswordModal = ({
           />
           {!errors.new && (
             <p
-              className="text-gray-600 text-size-sm mt-1"
+              className="text-charcoal-mid text-size-sm mt-1"
```
joshestein, line 249: ❤️ thanks for making this change! I started pivoting towards `text-bluedot-normal/80` as default muted text but don't think we need to make that change here, just something to keep in mind for future (looks like we still have many uses of `text-charcoal-mid` for secondary/muted so think this is still net win)

apps/website/src/components/settings/ChangeEmailModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const { reset: resetMutation } = requestEmailChange;
+
+  useEffect(() => {
+    if (isOpen) {
+      setNewEmail('');
+      setValidationError('');
+      resetMutation();
+      setTimeout(() => {
+        newEmailRef.current?.focus();
+      }, 100);
```
joshestein: Why do we need this `setTimeout`? Can we not just call `.focus()` directly? Or put `autoFocus` on the input element?

apps/website/src/server/routers/users.ts
```diff
@@ -175,30 +223,35 @@ export const usersRouter = router({
...
-      const token = await createEmailChangeToken({ userId: user.id, oldEmail: user.email, newEmail: input.newEmail });
-      await sendEmailChangeVerification({ oldEmail: user.email, newEmail: input.newEmail, confirmUrl: getEmailChangeConfirmUrl(token) });
+      const user = await getUserFromAuthOrThrow(ctx.auth);
 
-      logger.info(`[EmailChange] admin ${impersonationRealIdentity(ctx).sub} requested email change for user ${user.id}: ${user.email} -> ${input.newEmail}`);
+      recordEmailChangeAttemptOrThrow(user.id);
+
+      await sendEmailChangeConfirmation(user, input.newEmail);
+
+      logger.info(`[EmailChange] user ${user.id} requested their own email change: ${user.email} -> ${input.newEmail}`);
```
joshestein, line 271: Not sure we should have this p2 info in our raw logs

apps/website/src/components/settings/ChangeEmailModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    if (!result.success) {
+      setValidationError(result.error.issues[0]?.message ?? 'Please enter a valid email address');
+      return;
+    }
+
+    setValidationError('');
+    requestEmailChange.mutate({ newEmail: result.data });
+  };
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
```
joshestein: If we use a semantic `<form>` do we still need this?

apps/website/src/components/settings/ChangeEmailModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+  Modal,
+  P,
+  ProgressDots,
+} from '@bluedot/ui';
+import { TRPCClientError } from '@trpc/client';
+import { useEffect, useRef, useState } from 'react';
+import { z } from 'zod';
+import { trpc } from '../../utils/trpc';
+
+const emailSchema = z.string().trim().toLowerCase().email('Please enter a valid email address');
```
joshestein: For changing passwords we have a dedicated `changePassword.schema.ts` file to handle the dual-use in components and router. Maybe we should follow that pattern here?

joshestein (commented):
For changing password, the user currently needs to validate their old password. I understand we can't validate a past user email in the same way, but maybe for user-initiated email changes it's worth sending a courtesy email to their old email as a form of 'validation'?

I don't think there's real risk to accounts being taken over but might still be worth letting the user know.

Will-Howard:
> but maybe for user-initiated email changes it's worth sending a courtesy email to their old email as a form of 'validation'?

This is a good idea 👍, I'll add this

apps/website/src/server/routers/users.ts
Will-Howard: CodeRabbit also [raised this](https://github.com/bluedotimpact/bluedot/pull/2828#discussion_r3674354850) in the other PR: Personally I think it's not ideal, but doesn't pose any real risk because logs are only admin-accessible anyway, and it's much lower sensitivity than e.g. secrets appearing in logs

apps/website/src/components/settings/AccountSettingsSection.tsx
marn-in-prod: +1 this cleanup is great Will!

Regarding pivoting towards `text-bluedot-normal/80`. I think it's worth aligning in the meeting on where we go next. From the discussions with Cyrus it sounded like we'd want semantic aliases where possible, so it's obvious which colour is the de facto one for a given purpose (i.e. `text-muted`) rather than just naming the hue.

So roughly two rules:
1. Use a semantic alias when one exists (`text-muted`)
2. When it doesn't, use a named primitive from our palette (`text-charcoal-mid`), rather than a Tailwind default like `text-gray-600` or an inline opacity like `text-bluedot-normal/80`. Opacity values tend to drift over time (/80 in one place, /70 or /60 in another), so naming the colour once should keep us more consistent going forward.

apps/website/src/server/routers/users.ts
```diff
@@ -178,30 +226,35 @@ export const usersRouter = router({
...
       }
 
-      const token = await createEmailChangeToken({ userId: user.id, oldEmail: user.email, newEmail: input.newEmail });
-      await sendEmailChangeVerification({ oldEmail: user.email, newEmail: input.newEmail, confirmUrl: getEmailChangeConfirmUrl(token) });
+      const user = await getUserFromAuthOrThrow(ctx.auth);
 
-      logger.info(`[EmailChange] admin ${impersonationRealIdentity(ctx).sub} requested email change for user ${user.id}: ${user.email} -> ${input.newEmail}`);
+      recordEmailChangeAttemptOrThrow(user.id);
+
+      await sendEmailChangeConfirmation(user, input.newEmail);
```
marn-in-prod: Can we get an email confirmation sent to the original email that their login info has changed (Somewhere, doesn't have to be at this line of code).

apps/website/src/components/settings/ChangeEmailModal.tsx
marn-in-prod: I think either following it or at least making them both aligned would be good to do in this PR.

apps/website/src/server/routers/users.ts
```diff
@@ -178,30 +226,35 @@ export const usersRouter = router({
...
+      if (ctx.impersonation) {
+        throw new TRPCError({ code: 'BAD_REQUEST', message: 'Cannot change email when impersonating another user' });
       }
 
-      const token = await createEmailChangeToken({ userId: user.id, oldEmail: user.email, newEmail: input.newEmail });
-      await sendEmailChangeVerification({ oldEmail: user.email, newEmail: input.newEmail, confirmUrl: getEmailChangeConfirmUrl(token) });
+      const user = await getUserFromAuthOrThrow(ctx.auth);
 
-      logger.info(`[EmailChange] admin ${impersonationRealIdentity(ctx).sub} requested email change for user ${user.id}: ${user.email} -> ${input.newEmail}`);
+      recordEmailChangeAttemptOrThrow(user.id);
```
marn-in-prod: We still track that the user attempted an email change when it failed because of our own infra failing. i.e. customer.io returning an error because they're down will still count towards their usage. 

Maybe do a try catch and check if it's not a TRPCError and refund if that's the case?

apps/website/src/components/settings/AccountSettingsSection.tsx
marn-in-prod, line 1: I think just using `<form>` here is preferable? We really don't make that many forms atm, but figuring out some sort of consistency for "forms" on the site would be good.

We already use plain `<form>` in `NewsletterBanner`, `DropoutModal`, `FacilitatorSwitchModal` and `GroupSwitchModal`, but this section (and `ChangeEmailModal`) hand-roll Enter with `onKeyDown` on each input. Wrapping these in `<form onSubmit={...}>` would overall be cleaner right?

Not blocking, mostly flagging that a convention here would stop this diverging further.

marn-in-prod (approved):
Getting an email change notification sent to the original email would be good + not counting the user towards usage if our own infra failed. The others are mainly consistency things

apps/website/src/components/settings/ChangeEmailModal.tsx
Will-Howard: > Or put autoFocus on the input element?

I changed it to this. The timeout was just to make sure the element had finished rendering, but `autoFocus` is the right way to do this.

apps/website/src/components/settings/ChangeEmailModal.tsx
Will-Howard: Nope, `<form>` is much better. I've changed it!

Will-Howard:
Here's the new courtesy email for reference:
Screenshot 2026-07-30 at 09 38 15

apps/website/src/server/routers/users.ts
Will-Howard: > Maybe do a try catch and check if it's not a TRPCError and refund if that's the case?

Good idea, I've added this

marn-in-prod:
> Here's the new courtesy email for reference: Screenshot 2026-07-30 at 09 38 15

Looks good, but seperately, the email addressed attached to email changes isn't a noreply email address right? It's one where Bluedot folks can easily see messages from?

I'm asking because the last time I reset my password, the email came from "noreply@bluedot.org"

@Will-Howard

Will-Howard:
> Looks good, but seperately, the email addressed attached to email changes isn't a noreply email address right?

It uses team@bluedot.org, which is the general contact email we use. I also just sent an email to check that it's monitored. I won't wait for that before shipping though, it can be fixed after the fact if not.

PR #2834, opened 2026-07-29
 Remove unauthorised query spam

### LINKED ISSUE #2754: Recently logged out sessions spam UNAUTHORIZED warnings from protected Nav queries
## 🧩 Summary

The nav profile menu fires protected tRPC queries: `admin.canImpersonate` and `myBluedot.hasFacilitatorRegistrations` (×2). Shortly after logging out, the client still fires these, and the server rejects each with `UNAUTHORIZED`, which spams error logs in the console.

This goes away after ~1 minute, and doesn't happen if you start on a logged out session, so it is likely to do with the client failing to clear the auth object. Since the server-side rejection still works, I believe there is no actual user-facing risk here, but it would be good to clear up the log spam.

## ✅ Definition of done

We no longer get these `UNAUTHORIZED` warnings just after logging out.

joshestein:
Fixes #2754 

Logging out spammed the server logs with `Client error handling request: Authentication required` -  four per logout. No user-facing impact; the server rejected the requests correctly.

## Why it happens

Most marketing pages set `pageRendersOwnNav`. `/login/clear` does not. So navigating to logout swaps the Nav and remounts `ProfileLinks`, which schedules four protected queries while the user is still authenticated. `httpBatchLink`
resolves its `headers` callback at request time - by then `setAuth(null)` has run, so the batch goes out with an empty `authorization` header and every protected procedure in it returns UNAUTHORIZED.

## Why this is fixed server-side

A tokenless request to a protected procedure is routine client/server disagreement, not an anomaly worth a warning. A request that *did* carry a token but failed verification still warns, and `createContext` logs verification failures separately at `error`, so no real signal is lost.

Three client-side fixes were tried first and all failed:

- `enabled: !!auth` — no-op. Auth is still set at mount, and `_NavCta` already
  gates render on `isLoggedIn`.
- `cancelQueries()` on logout — the observers are still mounted at that instant,
  so React Query immediately refetches and *that* request carries the empty
  header.
- `abortOnUnmount: true` — necessary but insufficient. While it's `false` (the
  default) tRPC passes `signal: null`, so cancellation can't abort a request at
  all; and even when true, `dataLoader` dispatches on a `setTimeout` that React
  may not beat with the unmount re-render.

The batch request still fires and still returns 207.

Tested locally and ensured no spam in server logs.


joshestein:
@greptile-apps review

PR #2836, opened 2026-07-31
[fix] Load InterDisplay on production (headings were silently falling back to Inter)

### LINKED ISSUE #2832: Load InterDisplay on production (headings silently fall back to Inter)
### 🧩 Summary

`interDisplay` (`next/font`, `src/lib/fonts.ts`) is imported only in `_document.tsx`. The pages router emits `next/font` CSS only for fonts reachable from `_app`, and `_app.tsx` imports only `inter` — so `--font-inter-display` resolves empty in prod builds. `globals.css` points `.bluedot-h1` / `.bluedot-h2` at that empty var, so every website heading renders Inter on prod/staging while local dev renders true InterDisplay.

Other monorepo apps are unaffected: the library's `.bluedot-h1` uses `font-family: InterDisplay` from the `bluedot.org/fonts` `@font-face`.

Found during #2790.

### ✅ Definition of done

1. `interDisplay` reachable from `_app` so its CSS ships
2. Verified on staging (not local dev) that `.bluedot-h1` / `.bluedot-h2` compute to InterDisplay
3. Design sign-off — this flips every heading on prod, so it's a sitewide visual change
4. Homepage hero stays pinned to Inter (deliberate inline `style` override)

### 📌 Notes / References

Blocks closing #2735 (parent DoD: "typeface across all apps"). Don't compare dev vs prod typography until this lands.

joshestein:
# Description

`interDisplay` (`next/font`, `src/lib/fonts.ts`) was imported only from `_document.tsx`. The pages router emits `next/font` CSS only for fonts reachable from `_app`, so `--font-inter-display` was never *defined* in production builds, only referenced. That made the `.bluedot-h1`/`.bluedot-h2` rule in `globals.css` invalid at computed-value time, so every heading inherited `--font-sans` and silently rendered Inter. Local dev was unaffected, which is why it went unnoticed. Confirmed pre-fix: prod and staging both served CSS `89631b6d…` with the variable referenced once and defined zero times.

**Changes**

- `_app.tsx` imports `interDisplay` and applies `interDisplay.variable`, making the module reachable from the `_app` graph so its CSS ships. `_document.tsx` still puts both variable classes on `<Html>` — that's what lands the custom properties on `:root`, which `globals.css` needs for `--font-sans`. Both are load-bearing.
- Dropped weight 700 from the `next/font` call: nothing renders `.bluedot-h1`/`.bluedot-h2` at bold, and next/font preloads every declared weight. Ships 400/500/600.
- `adjustFontFallback: false`, so the swap falls through to already-preloaded Inter instead of next/font's generated Arial-metric face. Inter and InterDisplay are the same typeface at two optical sizes, so that swap is far less visible.
- Four homepage/lander headings hand-rolled `.bluedot-h2`'s styling on a raw `<h2>` *without* the font-family, so they'd have stayed Inter while `<H2>` flipped everywhere else. Added `bluedot-h2` and removed the classes that merely restated its defaults, keeping the real `font-medium`/`tracking-tighter` overrides.
- **Subsetted InterDisplay to latin + latin-ext**, 106 kB → ~42 kB per weight. Without this the PR added 327 kB of preloaded font to every page; it now adds 130 kB.
- Matched the homepage carousel heading ("Our community") to the other section headings. Its `variant === 'homepage'` branch stepped up to `text-size-2xl` above `bd-md`, rendering it at 56 px against 48 px for every other H2. `.bluedot-h2` already supplies `text-size-xl`, so the size classes went entirely. Pre-existing, surfaced while auditing heading consistency for this change.

## Font subsetting

Subsetted **in place** in `public/fonts/`, so the other apps that load these over HTTPS from `bluedot.org/fonts` (8 `@font-face` rules in `libraries/ui/src/default-config/tailwind.css`) get the smaller files too. That's deliberate: the library's `.bluedot-h1` resolves to `InterDisplay, var(--font-sans)` and its `--font-sans` leads with full-coverage `Inter`, so any glyph outside the subset degrades to Inter — the same typeface at a different optical size, not a system font.

All four weights subsetted, including 700, which the website no longer references but the library still declares; leaving it full-coverage would make glyph support vary by weight.

The subset range, the exact regeneration command and the `ss04` caveat are documented in `apps/website/README.md` under Font Inventory. Pre-subset originals are recoverable from `28b8c9fd9` if the range ever needs widening — re-run against those, not against already-subsetted files.

Two things worth knowing if you touch this:

- `--layout-features+=` is required. I first wrote `--layout-features='+ss04,…'` and it silently emptied GSUB — `ss04` gone, no error, no warning. Four of these headings set `fontFeatureSettings: "'ss04' on"`, so that would have shipped as an invisible typography regression. The README documents a one-liner to verify `ss04` survived.
- Out-of-range scripts: Cyrillic/Greek fall back to Inter (full-coverage, already loaded). CJK falls to a system font, as it did before — Inter Display never covered it.

## Testing

Verified against a local production build, with `origin/master` built and served alongside it so before/after differ by this change alone.

- `--font-inter-display:"interDisplay"` now defined, with no `"interDisplay Fallback"` companion (confirms `adjustFontFallback: false`).
- 3 `@font-face` at 400/500/600; 7 font preloads on `/about` (4 Inter + 3 InterDisplay), totalling 573 kB vs 443 kB on master.
- Rendered fonts via Chrome DevTools `CSS.getPlatformFontsForNode` — what actually painted, not what was requested:

| Page | Heading | Weight | Before | After |
|---|---|---|---|---|
| `/about`, `/our-community`, `/programs/advising` | H1 | 400 | Inter | **Inter Display** |
| `/programs/advising` | H2 | 600 | Inter | **Inter Display SemiBold** |
| `/` | 7 section H2s | 500 | Inter Medium | **Inter Display Medium** |
| `/` | hero H1 | 400 | Inter | Inter (unchanged, deliberate pin) |

- Subsetting verified two ways. Rendering the original and subsetted files side by side across **all four weights** with `ss04` enabled, over the real heading copy plus latin-ext and `ss04`-sensitive glyphs: **0 differing pixels** out of 6.72M. And re-running the rendered-font check against the served build with requests to `bluedot.org` blocked, so nothing could be served by production's still-full copy: 0 blocked requests, weights 400/500/600 all resolving to Inter Display faces. All four subsetted files confirmed to carry `ss04` and `ss02`.
- The screenshots below were captured pre-subset; the 0-pixel diff is why they remain accurate.
- Full suite green (127 files / 1077 tests), lint + typecheck clean. No snapshots reference the changed class strings.
- Homepage heading audit across 390/680/768/1024/1280/1440/1920: all 7 H2s uniform at every breakpoint (32 px below `bd-md`, 48 px above). The 66 px horizontal overflow at 1024 px is pre-existing header behaviour — staging on `master` shows the identical overflow from the same `nav-cta`/dropdown elements.

**No regression test added** — nothing stops a future `_app` cleanup or Next upgrade from silently reverting this, and the failure mode is invisible in dev.

## Deployment

⚠️ This flips every heading on the site. Merging goes to staging only; **the next `website/vX.Y.Z` release will carry a sitewide visual change**, so please don't cut one before design sign-off.

Note the subsetted fonts reach the other apps on the *website's* production release, not on their own deploys, since they fetch from `bluedot.org/fonts` at runtime.

Two things this makes visible for the first time, both worth a designer's eye:

- `prose` sets no heading font-family and `<H1>`/`<H2>` carry `not-prose`, so blog and markdown course-content headings stay Inter while chrome/marketing headings become InterDisplay.
- The homepage hero stays pinned to Inter (deliberate, `HomeHeroContent.tsx`), making it the only Inter hero on the site.

Latent before, live now: the `globals.css` heading rule falls back to Tailwind's stock *serif* stack, so headings would render Times if both webfonts failed. Left alone to keep this diff minimal.

Inter itself is still unsubsetted (443 kB across 4 weights). Subsetting it is the bigger win but carries real glyph risk — it renders user-generated names and its fallback is a system font, not a sibling cut — so it belongs in its own PR.

## Issue

Fixes #2832

## Developer checklist

## Screenshot

Zoomed crops first, since at full-page scale the difference reads as slightly tighter letterforms rather than anything obvious.

| Zoomed crop | Before | After |
|---|---|---|
| `/about` H1 (weight 400) |  |  |
| Homepage section H2 (weight 500) |  |  |

| 📸 | Before | After |
|---------|---|---|
| 📱 |  |  |
| 🖥️ |  |  |
| 🖥️ home |  |  |

The logo carousel differs between the homepage shots only because it's an infinite-scroll animation caught at a different frame.


joshestein:
@claude review

joshestein:
Added font subsetting: InterDisplay latin + latin-ext, 106kB → ~42kB per weight, cutting this PR's added preload from 327kB to 130kB. Subsetted in place so the other apps loading these from `bluedot.org/fonts` benefit too; out-of-range glyphs degrade to Inter, which is full-coverage and already in the chain.

Verified by pixel-diffing rendered headings before/after subsetting: 0 differing pixels, so `ss04` survived. Worth flagging for anyone regenerating these — `--layout-features='+ss04'` silently empties GSUB with no error; the correct syntax is `--layout-features+=ss04`. `apps/website/scripts/subset-fonts.sh` asserts `ss04` is present and fails if not.

@claude review

Issue #2808, opened 2026-07-19
Rename CERTIFICATE_CREATION_TOKEN -> AIRTABLE_AUTOMATION_TOKEN


Will-Howard:
### 🧩 Summary

CERTIFICATE_CREATION_TOKEN was first added to allow an Airtable automation to generate a course certificate via the `createFacilitatedCourseCertificate` endpoint. In #2777 I added another admin-action case, where an Airtable automation pings the `linkToUser` endpoint.

We should rename it to `AIRTABLE_AUTOMATION_TOKEN` to cover the generic admin-action use case. This is slightly fiddly because secrets are scoped per user, so both @joshestein and I will need to update a secret in Airtable.

### ✅ Definition of done

`AIRTABLE_AUTOMATION_TOKEN` is used in place of `CERTIFICATE_CREATION_TOKEN`, both in code and in the Airtable automations that call into the code.

joshestein:
@Will-Howard I did the rename in the Airtable secret + automation I wrote.

Would be nice if there was an Airtable service account that manages all secrets, probably something for another day.

joshestein:
@Will-Howard I realised I kinda jumped the gun here, would've been better to do a proper roll-over to avoid any intermediary errors rather than renaming like the reckless goose I am. I tried to find your automation but haven't been able to do so.

edit: I think it's this one: https://airtable.com/appnJbsG1eWbAdEvf/wfl2kgUBszKwhhyrw

Will-Howard:
No worries! There doesn't appear to have been any user impact in the end

Issue #2567, opened 2026-05-28
Allow users to delete their account and details from Course Hub


anglilian:
### 🧩 Summary

A handful of users have asked to have their data deleted in accordance with GDPR. It might be useful to include this under their account (https://bluedot.org/account).

### ✅ Definition of done

Users should see a confirmation screen that says it will permanently delete _all_ data BlueDot has associated with this email.

Deleting their account should delete all records linked to the user from Airtable and postgres. This includes:
- Course Builder: exercise responses, resource completions
- Course Runner: project submission, course feedback, peer feedback
- Applications: course registration, drop out/ defer

Will-Howard:
I'm working on building this now. I think we should separate the regular account deletion and GDPR cases. GDPR-compliant deletion entails several things that we/the user might consider to be overkill in the normal case where someone wants to delete their account e.g.:
- Unsubscribing from all newsletters (e.g. "Job opportunities")
- Deleting data not directly tied to their account, e.g. grant applications (which is just an email-based form submission)
- Deleting their profiles in customer.io, posthog, potentially Slack, etc

I checked the bug reports people have submitted requesting to delete their account, and only one out of 3 referenced GDPR specifically, the others just requested account deletion generically.

Here's my suggestion of the user-facing design for this:

| State | 🖥️ Desktop | 📱 Mobile |
|---|---|---|
| Account page | <img width="1280" height="800" alt="image" src="https://github.com/user-attachments/assets/93a18173-5c58-48d2-ba0b-e93fa25c6c7d" /> | <img width="390" height="844" alt="image" src="https://github.com/user-attachments/assets/4deefa19-c823-4e06-b5e6-6d4d9f8fbacd" /> |
| Delete account modal | <img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/06c54487-e498-410f-aa99-17346d206522" /> | <img width="390" height="844" alt="Image" src="https://github.com/user-attachments/assets/b1b1b792-be5a-424c-b710-beb4720bbc68" /> |
| After GDPR clickthrough | <img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/b4b0dedc-4548-406c-a8e2-f305e017a49a" /> | <img width="390" height="844" alt="Image" src="https://github.com/user-attachments/assets/0782545f-9694-4f4b-840a-4f60d08be571" /> |
[...]

Issue #2485, opened 2026-05-14
Scope proposal to "fix uniqueness" in the database


Will-Howard:
### 🧩 Summary

We have run into several issues of duplicate rows being created due to:
1. Airtable's inability to enforce uniqueness
2. Specifically, check-then-insert race conditions occurring due to `db.insert` having quite a long round trip from Airtable insert (~0.5s) -> Postgres insert (which is when we're able to "check" if the row already exists)

I believe (discussed in the engineering meeting a couple of weeks ago) we can fix this using advisory locks, doing something like:
- Defining pseudo-unique constraints at the level of `pgAirtable` (e.g. on `email` for User)
- Acquiring an advisory lock on these when doing `db.insert` or `db.update`

The devil is in the detail here. Naively this wouldn't work because the check in check-then-act is already outside `db.insert`, so we will need to come up with a pattern that makes it work.

### ✅ Definition of done

A proposal (in the form of a new Issue) outlining a refactor which will allow us to enforce uniqueness up to the application level. I.e. if someone goes into Airtable and creates duplicate rows we can't stop that, but we can guarantee that the website backend will never insert duplicate rows (according to defined constraints) itself.

PR #2856, opened 2026-08-06
[feat] Process ordinary account deletion requests

Will-Howard:
# Description

Plan from the [Issue](https://github.com/bluedotimpact/bluedot/issues/2567):

> Summary of the architecture I'm planning to go with here:
> 1. The on-site flow will insert a row into a [Deletion requests](https://airtable.com/appnJbsG1eWbAdEvf/tbl0b9zoIUjBPsL84/viwBZSACf21Ua4o2L?blocks=hide) table (similar to drop out/group switching requests)
> 2. This will then be picked up by an Airtable automation which kicks off the actual deletion
> 3. The automation itself pings a `executeAccountDeletion` endpoint on the website (using the [`AIRTABLE_AUTOMATION_TOKEN`](https://github.com/bluedotimpact/bluedot/issues/2808) for authentication)

> And the UI I'm planning to add on the website:
> - A new "Delete accounts" admin page (select user -> confirm modal)
> - The user-facing "Delete account" button as in the screenshots above

This PR sets up the backend processing of ordinary (non-GDPR) deletion requests. The user-facing behaviour I am aiming for when an account is deleted is:
- The email is freed. You are able to change another account to use that email, or create a fresh account with the same email
- If you do create an account with the same email, it starts empty and behaves the same as if you had signed up for the first time
- You stop receiving any communication resulting from you having an account

Things which are notably not included in this:
- Stopping email marketing that _is not_ the result of having an account. Specifically, newsletter signups through the anonymous form (e.g. at the bottom of the [programs page](https://bluedot.org/programs)), which also come through customer.io (so we need to keep the profile alive)
- Deleting data that was not visible to the user in the first place. E.g. groups switching/drop out requests
- Deleting data that is not closely tied to the course hub website, e.g. grant applications

Apart from that, the main decision I made was on the question of **soft deletion** (adding `deletedAt` columns to tables) vs **hard deletion**. I opted for hard deletion, because it simplifies things massively, and we don't have a clear use case for soft deletion (i.e. we don't need low-friction restore/undo). In the case where an account is accidentally deleted, we could bring it back from database backups if absolutely needed.

To do after deploying this:
- [x] Create an Airtable automation "process deletion request", triggered "on record matches conditions ("Pending" status)" in the Deletion requests table, which pings this `executeAccountDeletion` endpoint (edit: created [here](https://airtable.com/appnJbsG1eWbAdEvf/wflyrP2PiE5A51RZL/wachmD0zIcMc7RYuI), just need to turn on once this merges)

## Issue

#2567

## Developer checklist

### DIFF
--- a/apps/website/src/lib/api/accountDeletion.ts
+++ b/apps/website/src/lib/api/accountDeletion.ts
@@ -123,8 +123,7 @@ const buildDeletionSteps = ({ userId, keycloakIdentifier, registrationIds, meetP
     throw new Error('userId is null or empty');
   }
 
-  // Rows linked to these ids and nobody else. The overlap check is what keeps `<@` from also matching
-  // every row linked to nobody, which is how an empty Airtable cell syncs.
+  // Rows linked to these ids and nobody else. `arrayContained` on its own would otherwise match `[]` and delete rows with no owner
   const linkedSolelyTo = (column: SQLWrapper, ids: string[]): SQL | 'skip' => (
     ids.length === 0 ? 'skip' : and(arrayOverlaps(column, ids), arrayContained(column, ids))!
   );
--- a/apps/website/src/lib/api/accountDeletion.ts
+++ b/apps/website/src/lib/api/accountDeletion.ts
@@ -1,0 +1,233 @@
+import {
+  arrayContained,
+  arrayContains,
+  courseRegistrationTable,
+  deletionRequestTable,
+  eq,
+  exerciseResponsePgTable,
+  inArray,
+  meetPersonTable,
+  projectSubmissionTable,
+  resourceCompletionPgTable,
+  selfServeCourseRegistrationTable,
+  userTable,
+  type PgAirtableColumnInput,
+  type PgAirtableTable,
+  type SQL,
+} from '@bluedot/db';
+import { TRPCError } from '@trpc/server';
+import { logger } from '@bluedot/ui/src/api';
+import { slackAlert } from '@bluedot/utils/src/slackNotifications';
+import { deleteKeycloakUser, keycloakUserExists } from './keycloak';
+import { CIO_API_BASE, getProfileById, setSubscriptionTopics } from './customerio';
+import db from './db';
+import env from './env';
+import { DELETION_REQUEST_STATUS } from '../constants';
+
+// A newsletter signup is independent of the account, so deletion leaves it subscribed.
+const NEWSLETTER_TOPIC = 'topic_15';
+
+// Error if we try to delete more rows than a single account plausibly owns, to limit
+// the blast radius of a "match all rows" bug
+const MAX_DELETIONS_PER_STEP = 1000;
+
+type DeletionStep = {
+  name: string;
+  /** Deletes everything the step owns and returns how many records were deleted. */
+  run: () => Promise<number>;
+};
+
+export const runAccountDeletion = async (deletionRequestId: string) => {
+  const [request] = await db.scan(deletionRequestTable, { id: deletionRequestId });
+  if (!request) {
+    throw new TRPCError({ code: 'NOT_FOUND', message: 'Deletion request not found' });
+  }
+
+  if (request.status === DELETION_REQUEST_STATUS.completed) {
+    return { status: DELETION_REQUEST_STATUS.completed, steps: [] as { step: string; deleted: number }[] };
+  }
+
+  if (request.status !== DELETION_REQUEST_STATUS.pending && request.status !== DELETION_REQUEST_STATUS.failed) {
+    throw new TRPCError({ code: 'CONFLICT', message: `Deletion request is "${request.status}"` });
+  }
+
+  if (!request.userId) {
+    throw new TRPCError({ code: 'BAD_REQUEST', message: 'Deletion request is missing the user id' });
+  }
+
+  const { userId } = request;
+  // Airtable gives back '' rather than null for an empty cell
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  const keycloakIdentifier = request.keycloakIdentifier || null;
+
+  // Both anchors are read before anything is deleted, so a retry can still find the dependents.
+  const registrationIds = (await db.pg
+    .select({ id: courseRegistrationTable.pg.id })
+    .from(courseRegistrationTable.pg)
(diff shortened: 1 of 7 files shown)


Will-Howard:
@coderabbitai review @greptileai review

apps/website/src/lib/api/accountDeletion.ts
```diff
@@ -0,0 +1,233 @@
...
+    .select({ id: courseRegistrationTable.pg.id })
+    .from(courseRegistrationTable.pg)
+    .where(eq(courseRegistrationTable.pg.userId, userId))).map((row) => row.id);
+
+  const meetPersonIds = registrationIds.length === 0 ? [] : (await db.pg
+    .select({ id: meetPersonTable.pg.id })
+    .from(meetPersonTable.pg)
+    .where(inArray(meetPersonTable.pg.applicationsBaseRecordId, registrationIds))).map((row) => row.id);
+
+  await db.update(deletionRequestTable, { id: request.id, status: DELETION_REQUEST_STATUS.inProgress });
```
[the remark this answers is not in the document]
Will-Howard: This can only happen in the case of the ~unavoidable check-then-act race we can get due to our Airtable setup. I've added one guard to stop a Completed row being set back to Failed, but apart from that we just have to accept this until #2485 is resolved

apps/website/src/lib/api/accountDeletion.ts
```diff
@@ -0,0 +1,246 @@
...
+    },
+    deleteMatchingPgRecords({ name: 'exercise-responses', table: exerciseResponsePgTable, sqlCondition: arrayContains(exerciseResponsePgTable.pg.userId, [userId]) }),
+    deleteMatchingPgRecords({ name: 'resource-completions', table: resourceCompletionPgTable, sqlCondition: arrayContains(resourceCompletionPgTable.pg.userId, [userId]) }),
+    deleteMatchingAirtableRecords({ name: 'self-serve-course-registrations', table: selfServeCourseRegistrationTable, sqlCondition: eq(selfServeCourseRegistrationTable.pg.userId, userId) }),
+    // Leave project submissions that have co-authors
+    deleteMatchingAirtableRecords({ name: 'project-submissions', table: projectSubmissionTable, sqlCondition: meetPersonIds.length === 0 ? 'skip' : arrayContained(projectSubmissionTable.pg.participant, meetPersonIds) }),
+    // The Course runner registration and the Course builder registration are syncs of this one, and follow it.
+    // TODO flag facilitator compensation risk in PR
+    deleteMatchingAirtableRecords({ name: 'course-registrations', table: courseRegistrationTable, sqlCondition: registrationIds.length === 0 ? 'skip' : inArray(courseRegistrationTable.pg.id, registrationIds) }),
```
Will-Howard, line 168: Flagging for someone on the core team (e.g. @anglilian): Course registrations are the only place that the facilitator pay calculation lives until they submit an invoice, so there is the risk of losing this information if a facilitator account is deleted.

Things that mitigate the downside of this:
- Once they have _submitted_ an invoice it lives in the contractor payments base, so this only applies to the (odd) case where someone wants to delete a facilitator account before they have submitted their invoice
- In the user-facing version, I'm planning to not allow facilitators to delete their own accounts, so this only applies to the admin-initiated version
- In the worst case, we can recover the data from backup

So I think this is acceptable, I just wanted to flag it

apps/website/src/lib/api/accountDeletion.ts
```diff
@@ -0,0 +1,244 @@
...
+
+        await setSubscriptionTopics({ cioId, topics: Object.fromEntries(topics.map((topic) => [topic, false])) });
+        return 1;
+      },
+    },
+    deleteMatchingPgRecords({ name: 'exercise-responses', table: exerciseResponsePgTable, sqlCondition: arrayContains(exerciseResponsePgTable.pg.userId, [userId]) }),
+    deleteMatchingPgRecords({ name: 'resource-completions', table: resourceCompletionPgTable, sqlCondition: arrayContains(resourceCompletionPgTable.pg.userId, [userId]) }),
+    deleteMatchingAirtableRecords({ name: 'self-serve-course-registrations', table: selfServeCourseRegistrationTable, sqlCondition: eq(selfServeCourseRegistrationTable.pg.userId, userId) }),
+    // Leave project submissions that have co-authors
+    deleteMatchingAirtableRecords({ name: 'project-submissions', table: projectSubmissionTable, sqlCondition: meetPersonIds.length === 0 ? 'skip' : arrayContained(projectSubmissionTable.pg.participant, meetPersonIds) }),
```
marn-in-prod: This line means that every project with zero participants in the participant array (which we currently have 18 of) gets deleted whenever a user with at least one Meet Person is deleted. This feels like an unintended side effect that we should fix.

apps/website/src/lib/api/accountDeletion.ts
```diff
@@ -0,0 +1,244 @@
...
+    markFailedError = error instanceof Error ? error : new Error(String(error));
+  }
+
+  let message: string;
+  if (completedByAnotherRun) {
+    message = `[AccountDeletion] deletion request ${id} errored, but another run already completed it, so it stays "Completed". ${detail}`;
+  } else if (markFailedError === null) {
+    message = `[AccountDeletion] deletion request ${id} did not complete. ${detail} Re-firing the endpoint resumes it.`;
+  } else {
+    message = `[AccountDeletion] deletion request ${id} did not complete, and could not be marked failed (${markFailedError.message}). ${detail} It is stuck "In progress": set its status to "Failed" in Airtable before re-firing the endpoint.`;
```
marn-in-prod, line 246: Really happy to see messages that are more actionable during errors!

apps/website/src/lib/api/accountDeletion.ts
marn-in-prod: No idea why 18 projects have no participants though. I didn't look into it

marn-in-prod (approved):
Main flow looks good,  I found one project-selection issue that can delete unrelated rows

apps/website/src/lib/api/accountDeletion.ts
Will-Howard: Very good catch! I've added this helper to fix it:
```typescript
  const linkedSolelyTo = (column: SQLWrapper, ids: string[]): SQL | 'skip' => (
    ids.length === 0 ? 'skip' : and(arrayOverlaps(column, ids), arrayContained(column, ids))!
  );
```

PR #2866, opened 2026-08-10
[fix] Revert "Add Sentry error capturing to website" (#2864)

joshestein:
## What

Reverts #2864 to restore staging (and unblock production releases).

## Why

#2864 broke every Next-handled route on staging and production (plain-text 500s) starting ~12:20 UTC today. Production was restored by re-running the `website/v2.34.7` deploy; staging is still serving the broken build until this merges.

Root cause: `@sentry/nextjs` v10 depends on `@sentry/server-utils`, which lazily imports `meriyah/dist/meriyah.mjs` at request time. With `output: 'standalone'`, Next's file tracing only picked up meriyah's CJS entry, so the Docker image shipped without the `.mjs` file. First request → `MODULE_NOT_FOUND` → every page/API route (including `/api/health` and the 404 page) returned `Internal Server Error`. Static files from `public/` still worked, which is why the pods looked healthy.

Verified locally: prod standalone build reproduces the 500s with the exact `meriyah` error; adding the missing file restores normal responses.

A fresh PR will reintroduce Sentry with the tracing issue fixed and a runtime verification of the standalone bundle.


joshestein:
@claude review

PR #2867, opened 2026-08-10
[fix] Reintroduce website Sentry with @sentry/nextjs 10.70.0 (fixes standalone meriyah crash)

joshestein:
## What

Reintroduces Sentry error capturing for the website (#2864, reverted in #2866 after it took down staging and production), with the root cause fixed: `@sentry/nextjs` bumped 10.69.0 → 10.70.0.

## Why the original PR caused the outage

`@sentry/nextjs` 10.69.0 depends on `@sentry/server-utils`, which lazily imports `meriyah/dist/meriyah.mjs` at request time. With `output: 'standalone'`, Next's file tracing only picked up meriyah's CJS entry, so the Docker image shipped without the `.mjs` file. The first request hit `MODULE_NOT_FOUND` and every Next-handled route (pages, API, `/api/health`, even the 404 page) returned a plain-text 500. Static files from `public/` still worked, so pods looked healthy and the rollout "succeeded".

Fixed upstream in 10.70.0: [`fix(v10/nextjs): meriyah issue for standalone build`](https://github.com/getsentry/sentry-javascript/pull/23055).

## Changes vs #2864

- `@sentry/nextjs` `^10.69.0` → `^10.70.0` (lockfile pins 10.70.0). Everything else is an exact reapply of #2864.

## Verification

Reproduced the outage and the fix locally against the production standalone bundle:

- On 10.69.0: `npm run build`, ran `dist/standalone/apps/website/server.js` → every route 500 with `Cannot find module '.../node_modules/meriyah/dist/meriyah.mjs'` — identical to prod/staging behaviour during the incident.
- On 10.70.0: clean rebuild → `meriyah.mjs` present in `dist/standalone/node_modules/meriyah/dist/`; standalone server returns 200 on `/` and `/about`, a proper 404 page, and tRPC responds with well-formed JSON.

## Deployment notes

- Merging deploys to staging only. Please watch staging after merge before cutting a production release (`website/vX.Y.Z`).
- Follow-up worth considering: a CI smoke test that boots the standalone bundle and curls `/` — build success + rollout success both stayed green through this outage.


joshestein:
@claude review

joshestein:
@claude review

joshestein:
Re the two CodeRabbit findings — both are covered by the existing env-template mechanism, no changes needed:

1. **"Set client-side Sentry values before `next build`"** — the deploy scripts already do this. `deploy-production.sh` runs `cp .env.production.template .env.production.local` before `npm run build`, and `.env.production.template` (added in this PR) contains `NEXT_PUBLIC_SENTRY_DSN` and `NEXT_PUBLIC_SENTRY_ENVIRONMENT`. Next inlines them into the client bundle at build time. `deploy-staging.sh` does the same with `.env.staging.template`. This is the identical mechanism that delivers `NEXT_PUBLIC_POSTHOG_KEY` and `NEXT_PUBLIC_GTM_ID` today.

2. **"Map `SENTRY_DSN` to `NEXT_PUBLIC_SENTRY_DSN` in k8s"** — this wouldn't work: `NEXT_PUBLIC_*` vars are inlined at build time, so a pod-level env var can't reach the browser bundle. The k8s `SENTRY_DSN` var correctly serves the server/edge runtime (read via `process.env` at request time), while the templates above serve the client at build time.

PR #2868, opened 2026-08-10
[fix] Add website liveness, readiness probes against a Next-handled route

joshestein:
## What

Adds `/api/health` to the website and points a k8s readiness probe at it, for both staging and production deployments.

## Why

Today's Sentry outage (#2864, reverted in #2866) shipped an image that booted successfully but threw `MODULE_NOT_FOUND` on every request. Because the website has no readiness probe, k8s saw the container as healthy, `kubectl rollout status` reported success, and the deploy "passed" while every page and API route returned a 500. Static files from `public/` kept serving, which made the pod look even healthier.

A probe against a Next-handled route makes that failure mode fail the rollout instead. A TCP-level check wouldn't help here — the server was listening the whole time.

## Notes

- The endpoint deliberately has no database or Airtable dependency. Readiness should mean "this pod can serve requests", not "all upstream systems are up" — otherwise an Airtable blip pulls every pod out of the load balancer.
- Added to the site-gate exemption in `middleware.ts` alongside `/site-login`, so the probe still works if `SITE_ACCESS_PASSWORD` is set.
- No `nginx.template.conf` change needed: the proxy's default `location /` already forwards everything, and the probe hits the pod directly on port 8080.
- `periodSeconds: 10`, `failureThreshold: 3` — a pod has to fail for ~30s before being pulled, which won't flap on a slow first request.

## Verification

- Ran the dev server with `SITE_ACCESS_PASSWORD` set: `/api/health` → `200 {"status":"ok"}`, `/` → `307` to the login page. Confirms the middleware exemption works and the gate is otherwise intact.
- Unit test covers the handler's status and body.
- `tsc --noEmit` and eslint clean for both `apps/website` and `apps/infra`.

Suggested merge order: this first, then #2867 (Sentry reintroduction), so that PR's rollout is actually gated on the app serving traffic.

### DIFF
--- a/apps/infra/src/k8s/serviceDefinitions.ts
+++ b/apps/infra/src/k8s/serviceDefinitions.ts
@@ -19,11 +19,10 @@ const PG_SYNC_SLACK_CHANNEL_ID = 'C0BFEG755PG'; // #update_pg-sync
 // (apps/website/.env.production.template). Not a secret: it's exposed in the browser bundle.
 const POSTHOG_PROJECT_API_KEY = 'phc_NVqRwH6xZ0MiP9I5a9tx3l9zsTWBvMI1YEbvbCOXDcI';
 
-// Shared with the Service's targetPort below so the probe can't silently drift from it.
 const WEBSITE_TARGET_PORT = 8080;
 
-// Hits a Next-handled route, so a build that boots but throws on every request fails the
-// rollout instead of "succeeding". A plain TCP check can't tell those apart.
+// A Next-handled route, so a build that boots but throws on every request fails the
+// rollout instead of "succeeding".
 // Readiness pulls a stalled pod out of rotation; liveness restarts it so single-replica
 // website doesn't stay stuck NotReady until the stall clears on its own.
 const WEBSITE_HEALTH_CHECK = {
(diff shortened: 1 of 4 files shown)


joshestein:
@claude review

joshestein:
Re Greptile's P1 on `health.ts` — the limitation is real and I called it out in the PR description, but I don't think adding app dependencies to the probe is the right fix. Reasoning:

**It does catch today's outage.** The `meriyah` failure was in the request pipeline, before routing. Evidence from my local repro on 10.69.0: requesting a path that doesn't exist (`/api/public/pages`) returned 500 rather than 404, so nothing reached the router. `/api/health` would have 500'd identically, the probe would have failed, and the rollout would have blocked.

**Probing a real page wouldn't close the gap Greptile describes.** `/` and most pages are statically prerendered (`○` in the build output), so their HTML is served from disk and a runtime failure in the `_app` shell wouldn't surface there either. Catching that specific case would mean probing a dynamic, db-backed page.

**Which is the thing to avoid.** If readiness depends on Postgres or Airtable, an Airtable blip fails the probe on every pod simultaneously and k8s pulls the entire deployment out of the load balancer — converting a degraded dependency into a total outage. That's a strictly worse failure mode than the one being guarded against.

So the split I'd argue for: readiness answers "can this pod serve requests" (cheap, dependency-free, no correlated failure), and "does the app actually render" belongs in a pre-deploy CI smoke test that boots the standalone bundle and curls a real page. That's the remaining follow-up from today's incident and it closes the gap properly, before the image ever reaches the cluster.

joshestein:
@greptile-apps review

PR #2869, opened 2026-08-10
[fix] Add startupProbe to website deployments to fix prod crash loop

joshestein:
## What

Hotfix for the prod CD failure caused by #2868 (merged as `d2d97852b`).

## What happened

Merging #2868 crash-looped `bluedot-website-production-deployment` in prod: [CD run](https://github.com/bluedotimpact/bluedot/actions/runs/31396060959/job/93480742080). The container terminated every 30s exactly (`periodSeconds: 10 × failureThreshold: 3`) — the new `livenessProbe` had no grace period, so a pod that took longer than 30s to come up got killed and restarted before it ever passed, resetting the clock each cycle. Pulumi gave up after its 10-minute rollout timeout.

**No user-facing outage**: `bluedot.org` served `200` throughout (old pod kept serving — k8s `maxUnavailable` rounds to 0 at `replicas: 1`, so it never got torn down). `bluedot.org/api/health` returned `404` the whole time, confirming the new image never went live. The rollout itself just never completed, and any future push to master will hit the same failure until this merges.

## Fix

Add a `startupProbe` — same pattern `bluedot-login` already uses in this file. It gates readiness/liveness until the app responds once, with its own generous budget (180s), so a slow cold start (whatever the actual cause — cold boot, node contention) doesn't get misread as a liveness failure and killed prematurely.

## Testing

- `npx tsc --noEmit` (apps/infra) — clean
- `npx eslint src/k8s/serviceDefinitions.ts` — clean
- Can't fully reproduce prod cold-start timing locally; this needs to go through staging→prod CD to confirm the rollout completes cleanly this time.

### DIFF
--- a/apps/infra/src/k8s/serviceDefinitions.ts
+++ b/apps/infra/src/k8s/serviceDefinitions.ts
@@ -31,6 +31,15 @@ const WEBSITE_HEALTH_CHECK = {
   failureThreshold: 3,
 };
 
+// startupProbe gates readiness/liveness until the app responds once, so a slow cold start
+// (seen in prod: pod took >30s to come up) doesn't get misread as a liveness failure and
+// killed before it ever gets a chance to pass. Same budget as bluedot-login's startupProbe.
+const WEBSITE_STARTUP_PROBE = {
+  httpGet: { path: '/api/health', port: WEBSITE_TARGET_PORT },
+  periodSeconds: 10,
+  failureThreshold: 18,
+};
+
 const MCP_AGGREGATOR_HOST = 'mcp.k8s.bluedot.org';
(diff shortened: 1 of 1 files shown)


joshestein:
This is not the right fix. The problem is a version mismatch between infra config and prod image.

- Master commit d2d97852b added k8s probes hitting /api/health and the route itself.
- Infra deploys from master → probes applied to prod deployment immediately.
- Prod image pinned to release website/v2.34.8 (commit cc15aef56) — built before route existed. /api/health → 404 there.
- Probe 404 = fail. Readiness never passes, liveness kills container every ~30s. New pod crash loops forever, Pulumi times out, CI job fails.

PR #2872, opened 2026-08-10
[fix] Give missions hero titles more top breathing room

oliviahelens:
# Description

Mission detail pages (e.g. `/missions/us-public-support-for-biodefense`) have titles that routinely wrap to two lines. When a hero title wraps, it grows the hero past its min-height, and because `MarketingHero` bottom-aligns its content (`flex ... justify-end`), the title's top edge pins to the `pt-20` top-padding line — which sits right where the transparent nav ends, leaving almost no gap. The result looks cramped against the nav. Single-line titles (like "Mirror Life Policy") don't hit this because they bottom-align with room to spare above them.

This adds an opt-in `extraTopSpacing` prop to `MarketingHero` that bumps the top padding to `pt-24 bd-md:pt-32`, and applies it on the two missions pages (`/missions` and `/missions/<slug>`).

**Why opt-in rather than a global default change:** because the hero uses `justify-end`, top padding only affects titles tall enough to grow the hero past its min-height — i.e. wrapping titles. Every other page that uses `MarketingHero` (programs, courses, about, etc.) has a short, single-line title today, so a global bump would change their DOM without changing their layout, and would silently apply to any future long title. Scoping it to missions keeps the change to exactly the pages that have the problem; the shared default (`pt-20`) is unchanged, so no other hero moves.

## Issue

No linked issue — small UI polish fix.

Fixes #

## Developer checklist

- [x] Front-end code follows [Component Accessibility Checklist (for Testing)](https://github.com/bluedotimpact/bluedot/blob/master/DEVELOPMENT_HANDBOOK.md#component-accessibility-checklist) — spacing-only change, no semantic/interaction changes
- [x] Considered having snapshot tests and/or happy path functional tests — existing hero snapshots (about, join-us) confirmed unchanged vs. default; full suite green (1106/1106)
- [x] Considered adding Storybook stories — added a `TwoLineTitle` story demonstrating the missions case with `extraTopSpacing`

## Screenshot

Visual change is limited to mission detail pages. I was unable to capture live screenshots (this environment has no Airtable credentials, so the DB-backed mission pages don't render locally). **A human should attach before/after screenshots of a two-line mission title** (e.g. "US Public Support for Biodefense") at mobile and desktop widths.

| 📸 | Before | After |
|---------|---|---|
| 📱  | _needs capture_ | _needs capture_ |
| 🖥️ | _needs capture_ | _needs capture_ |

---
_Generated by [Claude Code](https://claude.ai/code/session_01LQDZchC2bnwvBwrF8Tfgy9)_

### DIFF
--- a/apps/website/src/__tests__/pages/__snapshots__/about.test.tsx.snap
+++ b/apps/website/src/__tests__/pages/__snapshots__/about.test.tsx.snap
@@ -543,7 +543,7 @@ exports[`AboutPage > should render correctly 1`] = `
         src="/images/homepage/hero.webp"
       />
       <div
-        class="relative z-10 flex flex-col justify-end h-full min-h-[317px] bd-md:min-h-[366px] pb-12 pt-20 bd-md:pb-16"
+        class="relative z-10 flex flex-col justify-end h-full min-h-[317px] bd-md:min-h-[366px] pb-12 pt-24 bd-md:pt-32 bd-md:pb-16"
       >
         <div
           class="section-base"
--- a/apps/website/src/__tests__/pages/__snapshots__/about.test.tsx.snap
+++ b/apps/website/src/__tests__/pages/__snapshots__/about.test.tsx.snap
@@ -543,7 +543,7 @@ exports[`AboutPage > should render correctly 1`] = `
         src="/images/homepage/hero.webp"
       />
       <div
-        class="relative z-10 flex flex-col justify-end h-full min-h-[317px] bd-md:min-h-[366px] pb-12 pt-24 bd-md:pt-32 bd-md:pb-16"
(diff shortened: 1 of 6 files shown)


oliviahelens:
@claude review

---
_Generated by [Claude Code](https://claude.ai/code)_

dewierwan:
Closing in favour of #2873, which tackles the same issue Li-Lian raised (thanks for the diagnosis here — the justify-end/pt-20 analysis matches what I found). #2873 centres the hero text between the nav and the hero's bottom edge in the shared component, so the gaps are equal for any title length on every page rather than only reduced on missions pages, and it has before/after screenshots + a viewport sweep attached. Discussed with Olivia, she's happy either way.

PR #2886, opened 2026-08-11
Return 404 instead of 500 for unknown course slugs

### LINKED ISSUE #2882: Unknown course slug throws 500 instead of 404
### 🧩 Summary

In `course.ts` we do:

```ts
  const course = await db.get(courseTable, { slug: courseSlug });

  if (!course) {
    throw new TRPCError({ code: 'NOT_FOUND', message: `Course with slug "${courseSlug}" not found.` });
  }
```

However, `db.get()` throws if there are no records found: https://bluedotimpact.sentry.io/issues/139953996/?environment=production&project=4511885468500049&query=is%3Aunresolved&referrer=issue-stream

This means the guard `if (!course)` never fires and we get 500s instead of 404s.

### 🛝 Steps to reproduce

Go to any `/courses/does-not-exist`, observe error

### ✅ Definition of done

1. Use `.getFirst()` instead of `.get()`
2. Potentially update `[[...chunkNumber]].ts` L173?

### 📌 Notes / Designs

_No response_

joshestein:
# Description

Visiting a course URL with a slug that doesn't exist (e.g. /courses/does-not-exist) returned a 500 error page. It now returns a proper 404. Same for unit pages, the congratulations page, and the group discussions endpoint under an unknown course.

`db.get()` was throwing an `AirtableTsError` when no records match, so guards would fail and the error would propagate up as `INTERNAL_SERVER_ERROR` (tRPC) or unhandled SSR exception (pages), producing the 500s in Sentry.

## Changes

New helper - getCourseBySlugOrThrow(courseSlug) in courses.ts:

const courses = await db.scan(courseTable, { slug: courseSlug });
// 0 rows  → TRPCError NOT_FOUND        (the 404 this PR is about)
// >1 rows → TRPCError INTERNAL_SERVER_ERROR (data-integrity failure, loud)
// 1 row   → returned

Why scan + trichotomy rather than getFirst: slugs are hand-authored in Airtable with no uniqueness constraint anywhere in the sync path. getFirst silently returns an arbitrary row on a slug collision - a quiet wrong-data failure. The old db.get threw loudly on >1; the helper keeps that protection but as a typed TRPCError, so the 0-row case is distinguishable as a 404 instead of being swallowed into a generic 500.

Call sites moved to the helper (all previously had the dead-guard pattern or an equivalent silent-collision getFirst):

- courses.ts — getCourseData (feeds tRPC getBySlug + three pages' SSR/SSG)
- courses.ts — getCurriculumMetadata (existence check only; bare await)
- group-discussions.ts — getByCourseSlug
- exercises.ts — getGroupExerciseResponses

Intentionally untouched:
- index.tsx course page: bare catch-all already returns notFound: true.
- certificates.ts:88: looks up the course by id from an already-validated registration, so a throw there is genuinely exceptional, not a user-input 404.

## Issue

Fixes #2882

### DIFF
--- a/apps/website/src/pages/courses/[courseSlug]/[unitNumber]/[[...chunkNumber]].tsx
+++ b/apps/website/src/pages/courses/[courseSlug]/[unitNumber]/[[...chunkNumber]].tsx
@@ -209,7 +209,7 @@ async function getUnitWithChunks(courseSlug: string, unitNumber: string) {
 
   const unit = units.find((u) => Number(u.unitNumber) === Number(unitNumber));
   if (!unit) {
-    throw new Error('NOT_FOUND');
+    throw new TRPCError({ code: 'NOT_FOUND', message: `Unit ${unitNumber} not found for course "${courseSlug}".` });
   }
 
   const allUnitChunks = await getActiveChunksByUnit(units);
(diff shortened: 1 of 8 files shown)


joshestein:
@greptile-apps rereview

PR #2917, opened 2026-08-20
Provision 2nd k8s node, cap loki cache

joshestein:
# Description

Part of the 2026-08-19/20 production incident. The single-node VKE cluster suffered repeated hard resets; the instance main-node-pool-f1af172060bd was eventually proven to be power-cycled externally every ~2 min (Vultr ticket open). The website and meet were temporarily moved to Render; Keycloak/auth was down ~19h.

1. Second node (main-node-pool-f20f7012c7cc) restored ingress, login, monitoring and currently stable
2. Capped cache memory: helm chart defaults committed ~11GiB of a 16GiB node to log caches alone (memcached requests/limits = 1.2× these values)

### DIFF
--- a/apps/infra/src/k8s/observability.ts
+++ b/apps/infra/src/k8s/observability.ts
@@ -326,6 +326,16 @@ new k8s.helm.v3.Release('loki', {
         },
       },
     },
+    // The chart defaults allocate 8GiB to the chunks cache and 1GiB to the results cache
+    // (memcached requests/limits are set to 1.2x these values), which committed ~11GiB of a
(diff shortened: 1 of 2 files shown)


joshestein:
@greptile-apps, @coderabbitai You are correct that the pulumi provider fails this update - it erroneously sends autoscaler min/max fields when auto_scaler is disabled (provider bug in @ediri/vultr; Vultr's API accepts plain quantity PATCHes fine). During the incident we applied the change out-of-band via the Vultr API and ran a targeted pulumi refresh, so state already matches this code - verified with pulumi preview (0 changes). Merging this PR aligns code with deployed reality; CI's pulumi up will no-op on the pool.

PR #2922, opened 2026-08-24
[feat] infra: raise inotify limits on every node via a DaemonSet

Will-Howard:
# Description

See comment:
```
// VKE nodes ship with kernel-default inotify limits (max_user_instances=128), which the
// pods on a node exhaust ("failed to create fsnotify watcher: too many open files").
// Vultr has no node-level config hook, so raise them from a DaemonSet instead: the
// drop-in file survives reboots, `sysctl -w` applies them immediately.
```

This low limit was discovered during the recent outage (see e.g. #2917), and potentially contributed to the cluster failing to recover fully from crashes.

## Issue

N/A

## Developer checklist

### DIFF
--- a/apps/infra/src/k8s/index.ts
+++ b/apps/infra/src/k8s/index.ts
@@ -4,5 +4,6 @@ import './postgres';
 import './certManager';
 import './observability';
 import './pvc';
+import './inotifyPatch';
 import './services';
 import './backups';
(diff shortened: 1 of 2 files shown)


marn-in-prod (approved):
`node-sysctl` already exists in prod from the manual fix, but it doesn't seem to be in Pulumi state. Won't Pulumi try to create it again when this is deployed? Wouldn't this cause the rollout to fail?

We also added `/etc/sysctl.d/99-bluedot-incident-inotify.conf` to both workers during the incident, and it seems to do the same thing as the new file. Could we delete the old file in this PR once the new config is applied?

Will-Howard:
> `node-sysctl` already exists in prod from the manual fix, but it doesn't seem to be in Pulumi state

> We also added `/etc/sysctl.d/99-bluedot-incident-inotify.conf` to both workers during the incident, and it seems to do the same thing as the new file

I'll delete both of these just before I merge so it starts from a clean slate

PR #2924, opened 2026-08-26
Forward utm_campaign and utm_content to the application form

Chris-Strahle:
## Problem

The Campaign and Content columns on Course registration in Airtable are empty on all applications. The site captures all five UTM params via `useLatestUtmParams`, but `buildApplicationUrl` only forwards `utm_source` to the MiniExtensions application form (as `prefill_Source`), so `utm_campaign` and `utm_content` are dropped.

These columns were historically populated by a Google Tag Manager tag (per the Airtable field description, last reviewed 2024-04-18), which the current prefill-URL flow replaced, wiring up only Source. We verified today (2026-08-26) with a test submission that the MiniExtensions form still accepts `prefill_Campaign` and `prefill_Content` and that both Airtable columns are populated from them.

## Change

- `buildApplicationUrl` now takes the full UTM params record (`Record<string, string> | null | undefined`) instead of just the source string, and forwards `utm_source` → `prefill_Source`, `utm_campaign` → `prefill_Campaign`, `utm_content` → `prefill_Content` (each only when non-empty). PostHog session/distinct id prefills, the %20 space-encoding workaround, and the relative-URL early return are unchanged.
- Updated all six call sites to pass `latestUtmParams` instead of `latestUtmParams.utm_source`.
- Updated `utils.test.ts` for the new signature and added cases: campaign+content forwarded when present, omitted when absent or empty, source-only unchanged, and utm_medium/utm_term not forwarded.

## Testing

- `vitest --run src/lib/utils.test.ts` in apps/website: 41 tests passed.
- `tsc --noEmit` and `eslint --max-warnings=0` on the changed files: clean.
- Not tested against the live form from this branch; the prefill params themselves were verified manually as described above.

### DIFF
--- a/apps/website/src/components/courses/CourseCompletionSection.tsx
+++ b/apps/website/src/components/courses/CourseCompletionSection.tsx
@@ -118,7 +118,7 @@ export default function CourseCompletionSection({
   }
 
   if (showEnrollmentCTA) {
-    const applicationUrlWithUtm = buildApplicationUrl(application.applicationUrl, latestUtmParams.utm_source);
+    const applicationUrlWithUtm = buildApplicationUrl(application.applicationUrl, latestUtmParams);
 
     const info = COURSE_INFORMATION_DETAILS[courseSlug];
     const accentColor = COURSE_CONFIG[courseSlug]?.accentColor;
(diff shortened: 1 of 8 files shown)


Will-Howard (approved):
Looks good to me, thanks for improving this!

One potential followup: The Future of AI course still only collects `utm_source` (registrations live in a different table to facilitated courses). Would you like me to add `utm_campaign` and `utm_content` to that too?

PR #2928, opened 2026-08-26
[chore] add optional pre-mortem section to change-proposal issue templates

anglilian:
## Summary

- Adds an optional **🔮 Pre-mortem** section to the Enhancement, Feature, and DevEx issue templates, between "Definition of done" and "Notes"
- Prompts the author to name the biggest things that could go wrong if the change shipped, so risks get surfaced and reviewed before implementation
- Agreed in the email-handling automations retro (26 Aug): a small-looking change had non-obvious downstream effects that a pre-mortem would have caught
- Bug Report and Design templates are unchanged — they describe problems and design work rather than proposed changes

## Test plan

- [x] All three YAML files validated with js-yaml
- [x] Diff is additive only (one textarea block per template, no other changes)

### DIFF
--- a/.github/ISSUE_TEMPLATE/2-enhancement.yml
+++ b/.github/ISSUE_TEMPLATE/2-enhancement.yml
@@ -23,6 +23,15 @@ body:
       required: true
 
   - type: textarea
+    id: premortem
+    attributes:
+      label: 🔮 Pre-mortem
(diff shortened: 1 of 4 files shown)


anglilian:
Also removed the default assignee (herrhaase) from the Design task template — new design issues no longer auto-assign.

Issue #2934, opened 2026-08-28
Set memory limits on kubernetes services


Will-Howard:
### 🧩 Summary

We had a recent outage (see e.g. #2917) due to an Out Of Memory issue (though we're not sure about the root cause exactly). This wasn't helped by the fact that our services don't have memory limits set, so in principle any one could OOM the entire node. We should set (initially generous) limits to reduce the chance of this in future

### ✅ Definition of done

Limits are set on most services, summing to significantly less than the memory (16gb) available in the node. For critical services, we could potentially leave them with no limit (check the specific cases)

Will-Howard:
I went through all the services we have and came up with a proposed memory request / limit. When I started coming up with these, we had linux's `cgroup` max memory as a measure of how much memory pods currently use. I then re-checked them with the more granular Prometheus metrics added by #2936.

These are the general principles I tried to follow:
- If it's a critical service, don't set a limit
- If there is nothing else to go on, size the limit at roughly 2-5x the observed peak, based on anon (heap) memory used (non-anon memory is page cache, which should be reclaimable)
- Set the memory _request_ around the current average usage

_Partly Claude written:_

✅ = applied

| Priority | Service | Current max usage / comments | Proposed request / limit |
|---|---|---|---|
| P1 | bluedot-website-production | 422Mi, but idle, traffic has been on Render since 20 Aug. Render shows it getting up to around 1 Gi | ✅ 1Gi / no limit |
| P1 | bluedot-website-proxy | 8Mi. Per nginx settings usage is ~50KB * num concurrent connections | ✅ 16Mi / no limit |
| P1 | bluedot-login | 691Mi | ✅ no request / 1000Mi (keep current values) |
| P1 | keycloak-pg | 960Mi working set, but most of that is reclaimable page cache. The floor is 30Mi anon + 141Mi shmem | ✅ no request / no limit (keep current values) |
| P1 | bluedot-pg-sync-service | 386Mi, covers a 202k-record full sync on 27 Aug. Grows with Airtable row count. Seems like it is at risk of runaway memory usage | ✅ 400Mi / 2000Mi |
| P1 | bluedot-meet | 55Mi anon (3d max); traffic is bursty on the hour | ✅ 128Mi / no limit |
| P2 | bluedot-availability | 71Mi anon (3d max) | ✅ 128Mi / 1Gi |
| P3 | bluedot-website (staging) | 403Mi (191Mi anon) | ✅ 256Mi / 1.5Gi |
[...]

PR #2936, opened 2026-08-28
[fix] infra: export k8s resource attributes as Prometheus labels from otel-collector

Will-Howard:
# Description

> // Without this, resource attributes (k8s.pod.name etc.) are dropped and every pod's
> // kubeletstats series collapses into one

This will let us view metrics like `container_memory_working_set_bytes` at a per-pod level, which will make it easier to diagnose memory issues in future

## Issue

Related to #2934

## Developer checklist


Will-Howard:
We currently still have the node that was experiencing the restart-loop in our cluster, and it has an otel DaemonSet that is (harmlessly) crash-looping. I think this would cause this to fail to deploy if we merged it now, so I'll remove that node first

Issue #2954, opened 2026-09-01
meet: add error states


Will-Howard:
### 🧩 Summary

The meet app has no error states, which caused confusion recently for a user trying to join a call while the app was having an outage. We should add error states for every API call, using the same pattern as the website app

### ✅ Definition of done

The meet app displays error states on API failures

PR #2958, opened 2026-09-02
[chore] meet: delete unused meeting-jwt endpoint

Will-Howard:
# Description

This was previously used to render the Zoom meeting under our url, but we switched in [~Feb 2025](https://github.com/bluedotimpact/bluedot/pull/355) to just directly linking out to Zoom, so this has been dead code since. I noticed this while adding error states (#2954), this removes a case to add error states for

## Issue

#2954

## Developer checklist


apps/infra/src/k8s/serviceDefinitions.ts
```diff
@@ -313,8 +313,6 @@ export const services: ServiceDefinition[] = [
         env: [
           { name: 'AIRTABLE_PERSONAL_ACCESS_TOKEN', valueFrom: envVarSources.airtablePat },
           { name: 'PG_URL', valueFrom: appPgConnectionDetails.uri },
-          { name: 'NEXT_PUBLIC_ZOOM_CLIENT_ID', value: 'lX1NBglbQWO2ERYSS1xdfA' },
-          { name: 'ZOOM_CLIENT_SECRET', valueFrom: envVarSources.meetZoomClientSecret },
```
Will-Howard, line 317: These are no longer used, so I'm deleting them too. I've saved them in 1Password in case we need them in future

marn-in-prod (approved):
LGTM

PR #2971, opened 2026-09-04
[fix] Make groupDiscussion.endDateTime nullable to stop sync warning spam

Will-Howard:
# Description

This field is often temporarily null while people are setting up new cohorts, and this causes spam alerts from pg-sync-service trying to coerce the value.

We plan to ban `notNull` eventually for any field that syncs from Airtable (#2081), so this is in line with that direction.

## Issue

N/A (cleaning up spam alerts)

## Developer checklist

### DIFF
--- a/apps/website/src/pages/api/calendar/discussions/[discussionId].ts
+++ b/apps/website/src/pages/api/calendar/discussions/[discussionId].ts
@@ -104,7 +104,7 @@ export default makeApiRoute({
     unitTitle: unit?.title,
     unitFallback: discussion.unitFallback,
     startDateTime: discussion.startDateTime,
-    endDateTime: discussion.endDateTime,
+    endDateTime: discussion.endDateTime ?? discussion.startDateTime + 60 * 60,
     zoomLink: discussion.zoomLink,
     activityDoc: discussion.activityDoc,
   });
--- a/apps/website/src/pages/api/calendar/discussions/[discussionId].ts
+++ b/apps/website/src/pages/api/calendar/discussions/[discussionId].ts
@@ -87,6 +87,10 @@ export default makeApiRoute({
     throw new createHttpError.Forbidden('You do not have access to this discussion');
   }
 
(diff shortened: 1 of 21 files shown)


apps/website/src/pages/api/calendar/discussions/[discussionId].ts
```diff
@@ -87,6 +87,10 @@ export default makeApiRoute({
     throw new createHttpError.Forbidden('You do not have access to this discussion');
   }
 
+  if (discussion.endDateTime === null) {
+    throw new createHttpError.Conflict('Discussion does not have an end time yet');
+  }
```
Comment (Will-Howard) [inline apps/website/src/pages/api/calendar/discussions/[discussionId].ts:92]: I decided to not allow downloading a calendar event if there is no end date. In Airtable, the creation of calendar events is gated on a manual checkbox, so I think almost always a null `endDate` will mean that the person setting up the cohort hasn't finished and won't want calendar events created yet.

libraries/computed-posthog-events/src/definitions.ts
```diff
@@ -84,7 +84,7 @@ const calculateDiscussionAttendanceEvents = async (
   const events: Event[] = [];
   for (const d of discussions) {
     // absent only materialises once a discussion has comfortably ended; attended emits as soon as recorded.
-    const hasEnded = d.endDateTime * 1000 <= nowMs - FIFTEEN_MINUTES_MS;
+    const hasEnded = d.endDateTime !== null && d.endDateTime * 1000 <= nowMs - FIFTEEN_MINUTES_MS;
```
Will-Howard, line 87: Always count a discussion with no `endDateTime` as _not_ having ended (relevant to 'attended' and 'absent' events)

Will-Howard:
Regarding the AI review comments: I decided to make it a stricter filter-out on the backend, rather than supporting `null` in an ad hoc way on the frontend, because it makes it easier to follow what the rule is (`null` == draft or otherwise not fully set up event)

PR #2976, opened 2026-09-07
Retry transient Airtable errors on website requests

marn-in-prod:
# Description

Sporadic Airtable downtime pages us with "temporarily unavailable" errors that pass on their own. They're usually 429 rate limits or 503s. This retries twice with backoff and jitter

429s always retry, inserts included, since a rejected call never went through. 5xx errors only retry on idempotent reads and writes by id. Inserts never retry a 5xx, so a lost response can't double-insert. [Airtable's docs](https://airtable.com/developers/web/api/rate-limits) recommend doing something similar (but with 30s wait timeouts which IMO is too long)

It quiets errors like these (all from `#updates_tech-prod`):

- `website: Error: Failed request on route users.getUser, type query: The service is temporarily unavailable. Please retry shortly.` (dozens across getUser, ensureExists, dropoutOrDeferral, and saveExerciseResponse)
- `website: Error: Failed request on route users.getUser, type query: Failed to get base schema: Status: 429.` (once)

## Issue

Fixes users seeing errors when Airtable has a blip.

## Developer checklist

- [x] Front-end code follows Component Accessibility Checklist, N/A, no UI changes
- [ ] Considered adding Storybook stories, N/A, no UI changes

## Screenshot

N/A, no visual changes.

### DIFF
--- a/libraries/db/src/lib/client.ts
+++ b/libraries/db/src/lib/client.ts
@@ -6,6 +6,7 @@ import {
 } from 'airtable-ts';
 import { ErrorType } from 'airtable-ts/dist/AirtableTsError';
 import { type PgAirtableTable } from './db-core';
+import { withAirtableRetry } from './retry';
 import { type AirtableItemFromColumnsMap, type BasePgTableType, type PgAirtableColumnInput } from './typeUtils';
 import {
   buildWhereClause, getFirstFromPg, type Filter, type PgDatabase,
@@ -211,7 +212,7 @@ export class PgAirtableDb {
     table: PgAirtableTable<TTableName, TColumnsMap>,
     data: Partial<Omit<AirtableItemFromColumnsMap<TColumnsMap>, 'id'>>,
   ): Promise<BasePgTableType<TTableName, TColumnsMap>['$inferSelect']> {
-    const fullData = await this.airtableClient.insert(table.airtable, data);
+    const fullData = await withAirtableRetry(() => this.airtableClient.insert(table.airtable, data));
 
     // `ensureReplicated` only returns undefined for idempotent deletes; upserts always return a result
     const pgResult = await this.ensureReplicated({ table, id: fullData.id, fullData });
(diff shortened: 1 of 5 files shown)


marn-in-prod:
@claude review

libraries/db/src/lib/retry.ts
```diff
@@ -0,0 +1,69 @@
+import { AirtableTsError } from 'airtable-ts/dist/AirtableTsError';
+import { ErrorType } from 'airtable-ts/dist/AirtableTsError';
+
+export const MAX_ATTEMPTS = 3;
+export const BASE_DELAY_MS = 1000;
```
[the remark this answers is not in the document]
marn-in-prod: You missed the point. Nobody is waiting 30s for a response

libraries/db/src/lib/retry.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const statusCode = (error as { statusCode?: unknown } | null | undefined)?.statusCode;
+  return typeof statusCode === 'number' ? statusCode : undefined;
+}
+
+export function isRetryableAirtableError(error: unknown, idempotent = false): boolean {
+  if (error instanceof AirtableTsError && error.type !== ErrorType.API_ERROR) return false;
+  const message = error instanceof AirtableTsError ? error.message : '';
+  const status = airtableStatus(error);
+  if (status === 429 || /rate limit/i.test(message)) return true;
```
Will-Howard, line 38: Minor: I discovered that the underlying `airtable` library actually already retries `429`s, see [here](https://github.com/Airtable/airtable.js/blob/v0.12.2/src/run_action.ts#L70-L74). This means 429s will never surface here. This doesn't make any difference to correctness, but could be slightly confusing.

Will-Howard: Oh I saw that this can happen for the schema fetch specifically, retracted

libraries/db/src/lib/client.ts
```diff
@@ -211,7 +212,7 @@ export class PgAirtableDb {
     table: PgAirtableTable<TTableName, TColumnsMap>,
     data: Partial<Omit<AirtableItemFromColumnsMap<TColumnsMap>, 'id'>>,
   ): Promise<BasePgTableType<TTableName, TColumnsMap>['$inferSelect']> {
-    const fullData = await this.airtableClient.insert(table.airtable, data);
+    const fullData = await withAirtableRetry(() => this.airtableClient.insert(table.airtable, data));
```
Will-Howard, line 215: There are some custom uses of `airtableClient` in:
- `apps/website/src/server/routers/course-registrations.ts` (`linkToUser`)
- `apps/website/src/lib/api/accountDeletion.ts`

This could be added to those too (though they are much lower frequency). `linkToUser` does sometimes have transient errors so I think it's worthwhile to add

Will-Howard (approved):
LGTM, it seems worth the complexity since Airtable continually has these errors

PR #2985, opened 2026-09-08
[fix] website: keep Past Courses visible for participants who deferred

eleniKougiou:
## What this fixes

A participant who finished a cohort course could not find it anywhere in My Courses — not In Progress, not Upcoming, not Past Courses. The course was visible while the round was Active and disappeared the day the round flipped to Past. With no row, there was no way to submit an action plan or reach a certificate.

This affects **anyone who has ever deferred**, whose round has ended, and who has no certificate yet — roughly **120 people today**, growing with every round that ends. No backfill is needed; they reappear on deploy.

## Root cause

`assignTab` in `apps/website/src/components/my-courses/useCourseListRow.tsx` returned `null` — no bucket, so `bucketCoursesByTab` dropped the row entirely — for any deferred registration whose round had ended without a certificate.

That was correct under the old model: deferring created a **duplicate** registration on the new round and left the original behind, so hiding the stale original was right.

Since #2363 a deferral **updates the round in place on the same registration**, and the Deferral record stays attached to it. `isDeferred` is therefore now true for the participant's current, completed course — so it rendered under In Progress while the round was Active, then fell into the hidden branch the moment the round turned Past.

## The change

`fetchDropoutStatusByRegId` now also selects `dropoutTable.newRoundId` (already declared and synced — no schema change needed) and derives a new row prop, `isDeferredToAnotherRound`: true only when a deferral moved the person onto a round *other* than the one this registration sits on.

`assignTab` hides only those rows. Modern in-place deferrals now bucket like any other row, landing in Past Courses with a working **Submit action plan** button. Legacy deferrals, which left a duplicate registration behind, behave exactly as before.

`isDeferred` is untouched in `classifyCourseRegistration` — it is what stops a deferred participant being labelled "Dropped".

The two near-identical branches in `assignTab` also collapsed into one priority list, since they differed only in what happens after the round ends.

## Rejected alternative

Clearing or deleting Deferral records in Airtable when someone starts their new round. It destroys the deferral audit trail (deferral-rate reporting, the confirmation automation and the `Is dropped out?` formula all read that table), needs a "started the round" event that doesn't exist, doesn't help anyone already affected without a backfill, and re-breaks on every future deferral.

## Tests

The table-driven bucketing suite in `my-courses.test.tsx` **encoded the buggy behaviour as intended** ("deferred + no cert stays hidden"), so those cases were deliberately updated. `defer: boolean` became `deferral: 'none' | 'toThisRound' | 'toAnotherRound'`, with new cases across Active / Future / Past / null-roundStatus × certificate / no certificate. The case that flipped:

```
rs: 'Past', deferral: 'toThisRound', cert: false  →  'pastCourses'   (was: null)
```

The fixture previously pre-resolved drop-vs-deferral precedence (`isDroppedOut: drop && !defer`), so that rule was never actually exercised; it now passes both flags through and tests it.

Four new router tests in `my-bluedot.test.ts` cover each branch of the derivation: in-place deferral (including repeat deferrals), deferral to another round, deferral with no `newRoundId` recorded, and a registration with no round at all.

**Verification:** 1290 tests across 142 files pass; `npm run typecheck` and `npm run lint` clean.

## Screenshots

None — this is bucketing logic with no visual change; the row and its buttons already render correctly once the row exists.

**A human should verify on staging**, since reproducing it locally needs a participant with a real deferral record:

| Person | Registration | Round | Expected |
|---|---|---|---|
| Eleni Kougioumtzi | `recIV0HdjvlGqPC1A` | AGI Strategy 2026 Jul W28 PT (Past) | visible in Past Courses, with Submit action plan |
| Zoe Isabel Senon | `recX7Z4vR8z7JjPi8` | Biosecurity 2026 Jul W30 PT (Past) | visible in Past Courses, with Submit action plan |
| Ryan Kawles | — | AGI Strategy 2026 Jul W29 Int (Past), no deferral, no cert | unchanged (already visible) |
| Todd Horinouchi | — | AGI Strategy 2026 Jul W29 Int (Past), no deferral, has cert | unchanged (already visible) |

`/?impersonate=<email>` reaches these without their credentials.

Merging deploys to **staging only**. Production needs a release tagged `website/vX.Y.Z`.

## Out of scope

**A second bug with the same symptom.** Mike Ushakov (`mike.ushakov@gmail.com`) reports an identical symptom but has no deferral record. His registration is linked to a User record for `mike.ushakov+bluedot@gmail.com`, which has never been logged into and carries no Keycloak identifier, while he logs in with the plain address. `myCoursesPage` filters on `eq(courseRegistrationTable.pg.userId, user.id)`, so his registration is never returned. That is the known duplicate-User pattern, fixed by him logging in with the registration email.

**An adjacent pre-existing flaw.** The router sets `isDroppedOut = false` whenever any Deferral record exists, so someone who defers and then genuinely drops out of the new round will never show as "Dropped". That was already true, but mattered less when deferrals spawned a fresh registration; now that the record stays attached permanently, it does not self-correct. Left alone here — it needs its own decision.

**A dead certificate link.** Reaching the course via the public course page and Browse curriculum yields a certificate button pointing at `bluedot.org/certification?id=` with an empty id, which renders an error page. That is `certificateUrl` falling back when `certificateId` is null, and it still affects anyone who arrives that way.


eleniKougiou:
@claude review

eleniKougiou:
manually checked mine and Zoe's cases, it looks good now

PR #2986, opened 2026-09-08
[chore] Remove the temporary legacy name-field fallback

### LINKED ISSUE #2913: Save first + last name on the User table and sync edits from course hub
### 🧩 Summary

The [User table](https://airtable.com/appnJbsG1eWbAdEvf/tblCgeKADNDSCXPpR/viwY9B6xQnrJGnB0f?blocks=hide) (Applications base) only stores a combined `Name`. The course hub signup form already collects first and last name separately (Keycloak stores both per user), but the website flattens them before saving: `users.ensureExists` writes the token's combined `name` claim, and `courseRegistrations.linkToUser` joins the registration's first + last name into one string.

Why this matters: applications sometimes arrive without first/last names, so the Applications base has gaps. Course hub captures exactly this data — it should flow back to the base, and stay in sync when users edit their name in course hub settings.

### ✅ Definition of done

- The User table has a `Last name` field (a legacy `First name` field already exists: `fldP9OEa1kOB0KARl`, currently unpopulated). Both are in `userTable` in `libraries/db/src/schema.ts`.
- New users get First name / Last name saved from:
  - the login token's `given_name` / `family_name` claims in `users.ensureExists` (all three branches, following the existing `name` pattern — backfill empty fields only, never clobber), and
  - the course registration's `First name` (`fldIhZ4wc5t1Yabgz`) / `Last name` (`fldHa6GR5aBsOBtkz`) in `linkToUser`'s created-user branch.
- The settings name editor (`ProfileNameEditor.tsx` / `users.updateName`) is split into First name + Last name inputs. Saving writes `firstName`, `lastName`, **and** `name` (the two joined) so every existing consumer of `name` keeps working and the fields can't drift.
- Existing users are backfilled from their linked course registration's first/last name (fill empty fields only; ~51k records, most recent registration wins if they disagree).

### 📌 Notes / Designs

- **Verify the token claims first.** The keycloak login preset requests scope `openid email profile` (`libraries/ui/src/Login.tsx`), so `given_name`/`family_name` should be in the token, and Keycloak definitely stores firstName/lastName per user (checked via the admin API). But the mapper emission hasn't been confirmed against a real token — decode one before building on it. If the claims are missing, fix the `profile` client scope mappers in the customers realm rather than working around it in code.
- **Two-PR sequencing.** Schema PR first (add `firstName`/`lastName` columns to `userTable`, edit in place), merge + deploy, wait for schema-sync to materialise the pg columns, then the consumer PR. Consumer code before the columns exist fails with `errorMissingColumn` locally too.
- Claim extraction belongs in `Login.tsx`'s `verifyAndDecodeToken` (mirror the existing `name` extraction) so the typing stays clean.
- Settings editor prefill: use existing `firstName`/`lastName` if present, otherwise split `Name` on the last space as a prefill the user confirms. Never bulk-apply that split — it's wrong for many names.
- Recommended: when a user saves their name in settings, also push firstName/lastName to Keycloak via the admin API (pattern in `apps/website/src/lib/api/keycloak.ts`), so future login tokens carry the corrected names instead of re-spreading stale ones.
- Signup's name fields are optional, so coverage will be partial either way — blank is acceptable.

Will-Howard:
# Description

The fields `name`, `firstName`, and `lastName` are now 100% in sync in production, so we can remove the fallback that #2972 included for the case of a `name` being set with no `firstName` or `lastName`.

## Issue

Fixes #2913

## Developer checklist

### DIFF
--- a/apps/website/src/lib/utils.test.ts
+++ b/apps/website/src/lib/utils.test.ts
@@ -23,7 +23,6 @@ const {
   buildApplicationUrl,
   userNameFields,
   normalisedFirstAndLastName,
-  legacyUserNameFields,
 } = await import('./utils');
 
 // Test constants
@@ -404,21 +403,5 @@ describe('normalisedFirstAndLastName', () => {
   it('returns null when the source has no name at all', () => {
     expect(normalisedFirstAndLastName({ name: ' ', firstName: '', lastName: null })).toBeNull();
     expect(normalisedFirstAndLastName({})).toBeNull();
-  });
-});
-
-describe('legacyUserNameFields', () => {
-  it('fills first/last for a user that only has a combined name, keeping the name', () => {
-    expect(legacyUserNameFields({ name: 'Mary Jane Smith', firstName: null, lastName: '' })).toEqual({ firstName: 'Mary', lastName: 'Jane Smith', name: 'Mary Jane Smith' });
-  });
-
-  it('changes nothing for a user whose fields are already stored together', () => {
-    expect(legacyUserNameFields({ name: 'Jane Doe', firstName: 'Jane', lastName: 'Doe' })).toEqual({});
-    expect(legacyUserNameFields({ name: 'Jane', firstName: 'Jane', lastName: null })).toEqual({});
-  });
-
-  it('changes nothing for a user with no name', () => {
-    expect(legacyUserNameFields({ name: null, firstName: null, lastName: null })).toEqual({});
-    expect(legacyUserNameFields({ name: '', firstName: '', lastName: '' })).toEqual({});
   });
 });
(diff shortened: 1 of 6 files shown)


Will-Howard:
> This PR is not safe to merge until name-only users are either prevented by an enforced invariant or handled when separate name fields are consumed.

> The user.name check assumes that a combined name guarantees populated firstName and lastName, but those columns remain nullable and current registration or manual-import paths can create name-only users.

Regarding these: This can't actually happen via any code path, and the fields are currently 100% in sync, so they could only diverge via manual edits or a future code path that breaks this invariant.

We could prevent this fully by e.g. changing `name` in Airtable to be a formula field that joins the other two fields (or in principle via a constraint in Postgres, though this would be fragile), or dropping it entirely. However I don't think this is worth doing at the moment because we would need to handle dependencies of that field throughout the codebase and in Airtable

Issue #2983, opened 2026-09-08
pg-sync-service: harden against getting stuck in a "sync in progress" state


Will-Howard:
### 🧩 Summary

Currently if a pg-sync-service pod dies in the middle of a full sync, it can easily leave the system in a "sync in progress" state that prevents a new full sync from starting. We should harden against this:
1. Add SIGTERM handling to reset `syncInProgress` so a normal deployment rollout can't pin the system in this state
2. Add a timeout (of ~3 hours initially) for full syncs, after which starting a new full sync is allowed

### ✅ Definition of done

The above two changes are implemented

PR #2992, opened 2026-09-10
[chore] Remove --initial-sync-tables flag from pg-sync-service

Will-Howard:
# Description

I'm working on hardening pg-sync-service against some bugs relating to full syncs ([Issue](https://github.com/bluedotimpact/bluedot/issues/2983), [draft PR](https://github.com/bluedotimpact/bluedot/pull/2991)). In doing so, I'm finding that the need to pass around the set of table names complicates the code quite a bit. Without this we could treat all types of full sync (admin-initiated, schema-changes, `--initial-sync` flag) in the same way, so don't need to keep track of the trigger after it has been kicked off.

I only added this flag originally as a developer aid. I personally haven't used it for many months. I don't know if other people use it, but any data setup can be replicated with `pg_dump` and `pg_restore` so I don't think it's worth having. I'm open to other takes though!

## Issue

#2983

## Developer checklist

### DIFF
--- a/apps/pg-sync-service/README.md
+++ b/apps/pg-sync-service/README.md
@@ -10,13 +10,10 @@ npm run start
 
 # Start with initial full sync
 npm run start -- --initial-sync
-
-# Start with initial sync for specific tables only, using postgres table names
-npm run start -- --initial-sync-tables course person user
 ```
 
 ## How it works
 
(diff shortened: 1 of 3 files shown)


eleniKougiou (approved):
Makes sense to me!

marn-in-prod:
Commented to +1 that just dumping the data initially is so much faster and would highly recommend instead.

marn-in-prod (approved):
LGTM

eleniKougiou (approved):
Makes sense to me!

PR #3000, opened 2026-09-14
[fix] website: stop telling intensive participants they've missed discussions that haven't happened

eleniKougiou:
## The bug

Participants on **intensive** rounds saw, from day one of the course, a sidebar card reading **"Course complete – See what's next"**, which led to **"Certificate requirement not met. Discussions attended: 1 of 6 – Missed 5 (max 1 allowed)"**.

Reported by a participant on Technical AI Safety (2026 Sep W38) Intensive, round `rec2dJgFfjUHngqIP`, which started today. His data was correct — 6 units, 1 discussion attended so far. The page was counting discussions that hadn't happened yet.

Present since #2421 (6 May 2026), so not a recent regression. The two intensive rounds live this week have 193 participants with a group between them.

## Cause

`certificates.getStatus` judged attendance with `numUnits - uniqueDiscussionAttendance <= 1`, where `numUnits` is every discussion in the course. On day one that shortfall is maximal by definition.

The only thing hiding it was a window measured in **calendar days** — `round.lastDiscussionDate` within seven days — which suited weekly cohorts, where a week out means one discussion remains. An intensive runs six discussions in six days, so the window covered the entire course.

## Fix

Measure the end of the course in **discussions**, not days, and count misses from **discussions that have already been held**.

- `getStatus` loads the participant's `expectedDiscussionsParticipant` records and counts the ones whose `endDateTime` has passed.
- `attendance-ineligible` is reported only when at most one expected discussion is still to come *and* the shortfall among held discussions is 2 or more. A missed discussion can't be un-missed, so at that point the verdict is final.
- The same test gates the action-plan nudge, so an intensive participant isn't asked for their action plan on day one either. One flag, `hasAtMostOneDiscussionLeft`, serves both.
- Anything that can't be placed in the past counts as still to come: a discussion with no `endDateTime` (not scheduled yet) and a linked id whose row hasn't synced. The expected total comes from the ids on the registration, not from the rows we managed to read, so a partial sync can't shrink the denominator and make a shortfall look final.
- No format special-casing and no string-matching on "Intensive".

The client-side gate in `isCongratulationsAccessible` loses its `&& isLastDiscussionSoonOrPassed` clause: the server no longer reports the status prematurely, so there's one rule instead of two that had to agree.

**Also fixed:** because "missed" was `total − attended`, a part-time participant who had missed exactly one discussion and still had the last one to attend was told "Missed 2 – requirement not met" at the moment the screen unlocked. That was the most common case.

**Registrations with no discussions linked** (never assigned a group, or not yet synced) have nothing to count, so both the verdict and the nudge fall back to the round's end date, and wait for the whole of `lastDiscussionDate` to be behind us — it's a date-only field parsed as midnight UTC, so anything less fires early. Today the verdict fires a week before the round ends; the nudge also moves from a week out to the round's end, which is the one deliberate behaviour change here beyond the bug itself.

## Behaviour

| | Before | After |
|---|---|---|
| Intensive, day 1, attended 1 of 1 held | "Course complete" → "Missed 5" | locked panel, "Finish the course and meet requirements to unlock" |
| Intensive, day 3, attended 1 of 3 held | "Missed 5" | nothing — two discussions still to come |
| Intensive, final day, attended 3 of 5 held | "Missed 3" | "Missed 2" |
| Part-time, last discussion still to attend, missed 1 | "Missed 2 – requirement not met" | action-plan state |
| Part-time, last discussion still to attend, missed 2 | "Missed 3" | "Missed 2" |

`has-certificate`, facilitator, `not-enrolled` and the self-paced Future of AI path are untouched.

## Tests

Nine cases under a new `participant attendance` block in `certificates.test.ts`, seeding discussion end times relative to now: intensive day 1 and day 3, final day with a real shortfall, one-miss-with-final-pending, whole round over, unscheduled discussions treated as pending, a linked discussion whose row hasn't synced treated as pending, and both no-schedule fallback paths. Day one asserts the action-plan nudge stays off, not just the status.

Ran `certificates.test.ts` (39), `congratulations.test.tsx` (15) and `SidebarCertificatePanel.test.tsx` (8) — all pass. eslint and `tsc --noEmit` clean. The full `apps/website` suite (144 files, 1315 tests) passed on an earlier revision of this branch; since then I've run the affected files only, and the changes touch nothing outside them.

## Screenshots

Not captured — the states here need a seeded participant mid-round. The user-visible text change is the card body, from "Discussions attended: 3 of 6 discussions - Missed 3" to "Discussions attended: 3 of 5 discussions so far - Missed 2", plus which of the three existing sidebar panels renders. `SidebarCertificatePanel` and `Congratulations` stories cover both states in Storybook if someone wants a look.


eleniKougiou:
@claude review

eleniKougiou:
Thanks — both P1s were real, fixed in 0dbfce7a, and the P2 convention is now followed. The logic is also simpler than it was: one flag instead of two.

**Missing rows finalize attendance** — correct, and the important one. The remaining-discussion count came from the rows returned by `groupDiscussionTable`, so a future discussion whose row hadn't synced shrank the denominator and could make the shortfall look final. The total now comes from `expectedDiscussionsParticipant` — the ids on the registration itself — so a linked discussion with no row behaves exactly like one with no `endDateTime`: something we can't place in the past, therefore still to come. Covered by a new test that links six ids with only four rows present.

**Final day ends prematurely** — correct for the no-schedule fallback. `lastDiscussionDate` is date-only, parsed as midnight UTC, so the fallback fired from the start of the final day. It now requires `roundEndTime + ONE_DAY_MS <= Date.now()`, i.e. the whole day behind us. Worth noting the blast radius is small: this branch only runs for registrations with no discussions linked at all, who have no discussion that day to attend — but the off-by-a-day was real and the fix is a line.

**Array access lacks guard** — restructured so there's no non-null assertion to guard. The fixture now carries `{ id, offset }` together through one `map` instead of building an id array and indexing back into it, which is tidier regardless of the rule.

While in there, the two near-identical end-of-course tests collapsed into a single `hasAtMostOneDiscussionLeft`. They only differed in the no-schedule fallback, where one waited a week before the round's end and the other waited for the end itself; both now wait for the end. That shifts the action-plan nudge slightly later for registrations with no linked discussions, and removes a whole second definition of "the course is nearly over".

Tests: `certificates.test.ts` 39/39, `congratulations.test.tsx` 15/15, `SidebarCertificatePanel.test.tsx` 8/8. eslint and `tsc --noEmit` clean.

eleniKougiou:
@claude review

eleniKougiou:
Second pass checked — no code changes needed, one documentation fix applied.

**Claude bot**: confirms both P1s and the P2 are correctly fixed in 0dbfce7a, and its one note is fair — the PR description still claimed the nudge "keeps its original week-out window" for registrations with no discussions linked, which stopped being true when the two flags collapsed into one. Description updated: that fallback now waits for the round's end for both the verdict and the nudge, and it's called out as the one deliberate behaviour change beyond the bug.

**CodeRabbit**: its single actionable finding — count missing expected discussion ids as pending — is the same issue Greptile raised as a P1, and it reviewed 684c0256, the commit before the fix. Already resolved in 0dbfce7a: the expected total comes from `expectedDiscussionsParticipant` rather than the returned rows, so a deleted or not-yet-synced row counts as still to come. `treats a linked discussion with no row yet as still to come` covers it.

Its ast-grep "Avoid SQL injection" hit on the whole `getStatus` procedure is a false positive — the only query added here is a drizzle `inArray` on `groupDiscussionTable`, parameterised, with ids read from the registration row rather than from user input. No change made.

eleniKougiou:
Closing in favour of a fresh PR on the same branch — the review threads here all point at lines that no longer exist, which made this hard to read. The code is unchanged (squashed to one commit); replacement PR linked below.

For the record, the two genuine findings from this thread and their fixes: the expected-discussion total now comes from the ids linked on the registration rather than the rows returned from `groupDiscussionTable`, so a deleted or unsynced row counts as still to come; and the no-schedule fallback waits for the whole of `lastDiscussionDate` to pass, since it's a date-only field parsed as midnight UTC. Both carried into the new PR.

PR #3001, opened 2026-09-14
[fix] website: stop telling intensive participants they've missed discussions that haven't happened

eleniKougiou:
## The bug

Participants on **intensive** rounds saw, from day one of the course, a sidebar card reading **"Course complete – See what's next"**, which led to **"Certificate requirement not met..."**.

Present since #2421 (6 May 2026), so not a recent regression. It affects every participant of an intensive course.

## Cause

`certificates.getStatus` judged attendance with `numUnits - uniqueDiscussionAttendance <= 1`, where `numUnits` is every discussion in the course. On day one that shortfall is maximal by definition.

The only thing hiding it was a window measured in **calendar days** — `round.lastDiscussionDate` within seven days. This suited weekly cohorts, where a week out means one discussion remains. _An intensive runs six discussions in six days, so the window covered the entire course._

## Fix

Measure the end of the course in **discussions**, not days, and count misses from **discussions that have already been held**.

- `getStatus` loads the participant's `expectedDiscussionsParticipant` records and counts the ones whose `endDateTime` has passed.
- `attendance-ineligible` is reported only when at most one expected discussion is still to come *and* the shortfall among held discussions is 2 or more. A missed discussion can't be un-missed, so at that point the verdict is final.
- The same test gates the action-plan nudge, so an intensive participant isn't asked for their action plan on day one either. One flag, `hasAtMostOneDiscussionLeft`, serves both.
- Anything that can't be placed in the past counts as still to come: a discussion with no `endDateTime` (not scheduled yet) and a linked id whose row hasn't synced. The expected total comes from the ids on the registration, not from the rows we managed to read, so a partial sync can't shrink the denominator and make a shortfall look final.
- No format special-casing and no string-matching on "Intensive".

The client-side gate in `isCongratulationsAccessible` loses its `&& isLastDiscussionSoonOrPassed` clause: the server no longer reports the status prematurely, so there's one rule instead of two that had to agree.

**Also fixed:** because "missed" was `total − attended`, a part-time participant who had missed exactly one discussion and still had the last one to attend was told "Missed 2 – requirement not met" at the moment the screen unlocked. That was the most common case.

**Registrations with no discussions linked** (never assigned a group, or not yet synced) have nothing to count, so both the verdict and the nudge fall back to the round's end date, and wait for the whole of `lastDiscussionDate` to be behind us — it's a date-only field parsed as midnight UTC, so anything less fires early. Today the verdict fires a week before the round ends; the nudge also moves from a week out to the round's end, which is the one deliberate behaviour change here beyond the bug itself.

## Behaviour

| | Before | After |
|---|---|---|
| Intensive, day 1, attended 1 of 1 held | "Course complete" → "Missed 5" | locked panel, "Finish the course and meet requirements to unlock" |
| Intensive, day 3, attended 1 of 3 held | "Missed 5" | nothing — two discussions still to come |
| Intensive, final day, attended 3 of 5 held | "Missed 3" | "Missed 2" |
| Part-time, last discussion still to attend, missed 1 | "Missed 2 – requirement not met" | action-plan state |
| Part-time, last discussion still to attend, missed 2 | "Missed 3" | "Missed 2" |

`has-certificate`, facilitator, `not-enrolled` and the self-paced Future of AI path are untouched.

## Tests

Nine cases under a new `participant attendance` block in `certificates.test.ts`, seeding discussion end times relative to now: intensive day 1 and day 3, final day with a real shortfall, one-miss-with-final-pending, whole round over, unscheduled discussions treated as pending, a linked discussion whose row hasn't synced treated as pending, and both no-schedule fallback paths. Day one asserts the action-plan nudge stays off, not just the status.

Ran `certificates.test.ts` (39), `congratulations.test.tsx` (15) and `SidebarCertificatePanel.test.tsx` (8) — all pass. eslint and `tsc --noEmit` clean. The full `apps/website` suite (144 files, 1315 tests) passed on an earlier revision of this branch; since then I've run the affected files only, and the changes touch nothing outside them.

## Screenshots

Before: Day 1 of intensive course incorrectly shows the course complete panel item
Screenshot 2026-09-14 at 17 55 18
Screenshot 2026-09-14 at 17 55 09

After: Day 1 of intensive course correctly shows the locked panel item
Screenshot 2026-09-14 at 18 02 05

🤖 Generated with [Claude Code](https://claude.com/claude-code), small adjustments by Eleni


eleniKougiou:
@claude review

eleniKougiou:
Good catch from both of you — that one's real, fixed in 3c095939.

`const attended = meetPerson.uniqueDiscussionAttendance ?? 0` quietly dropped a fail-soft the previous implementation had: its `hasAttendedEnough` began with `uniqueDiscussionAttendance == null`, so an unknown attendance count never produced a verdict. Defaulting it to `0` meant a participant whose rollup hadn't synced could be told they'd missed every discussion held so far — the same false-accusation class this PR exists to fix, arriving through a different door. `attended` is now nullable and the ineligible branch requires `attended != null`, so a missing rollup reads as "not known yet". New test: `reports nothing while the attendance rollup has not synced` — six discussions all held, round over, null attendance, expects `action-plan-pending`. It fails against the previous commit.

On the minor point about short courses: `expectedDiscussionIds.length - heldSoFar <= 1` does flip to `true` after the first discussion of a two-discussion course, but it can't produce a wrong verdict there — the shortfall test needs `held - attended > 1`, which needs at least two held discussions, and with both held and none attended "missed 2" is accurate. The only effect is the action-plan nudge appearing after the first of two discussions, which seems reasonable for a course that short. Left as-is.

Tests: `certificates.test.ts` 40/40. eslint and `tsc --noEmit` clean.

eleniKougiou:
@claude review

PR #2272, opened 2026-04-13
Use UTC course deadlines and clarify deadline copy

somepersonnamedjosh:
## Summary
- use UTC day boundaries for course application deadline cutoffs
- keep CTA copy short while clarifying round listings with explicit UTC close times
- update tests and typed story fixtures for the new deadline shape

## Testing
- npm test -- src/server/routers/course-rounds.test.ts src/components/courses/SideBar.test.tsx
- npm run typecheck
- npm run lint -- src/server/routers/course-rounds.ts src/server/routers/course-rounds.test.ts src/lib/utils.ts src/components/lander/components/ScheduleRounds.tsx src/components/homepage/UpcomingRounds.tsx src/pages/courses/index.tsx src/components/courses/SideBar.tsx src/components/courses/MobileCourseModal.tsx src/components/courses/SideBar.test.tsx src/components/courses/SideBar.stories.tsx src/components/courses/MobileCourseModal.stories.tsx src/__tests__/pages/courses/index.test.tsx src/components/courses/DropoutModal.stories.tsx src/components/courses/InactiveCourseBanners.stories.tsx src/components/homepage/UpcomingRounds.stories.tsx src/components/lander/components/CourseInformationSection.stories.tsx src/components/lander/components/ScheduleRounds.stories.tsx src/components/pages/courses/CoursesPage.stories.tsx

PR #3002, opened 2026-09-14
Make course application deadlines anywhere-on-earth

eleniKougiou:
## Summary

The website closed course rounds at **00:00 UTC** the day after the application deadline. The Airtable Round `Status` formula in the Applications base was changed to anywhere-on-earth (AoE) — a round stays Future until **12:00 UTC** the day after its deadline. Until this ships the site hides a round 12 hours before the form stops accepting it, so someone in a behind-UTC timezone sees the round vanish while they can still apply.

Next deadline affected: **Sun 27 Sep** (TAIS, Biosecurity, FAIG Oct W41).

## Changes

- `course-rounds.ts` — `getDeadlineThresholdUtc()` derives the UTC day from `Date.now() - 12h`. A "12 Apr" deadline now stays open until 12:00 UTC on 13 Apr, i.e. 23:59 on 12 Apr in UTC−12. A doc comment pins the matching Airtable formula so the two don't drift apart again.
- `utils.ts` — `formatApplicationDeadlineUtcDetailed` → `formatApplicationDeadlineDetailed`; output `"27 Sep at 23:59 UTC"` → `"27 Sep at 23:59 anywhere on earth"`. The old name baked UTC into the identifier.
- `RoundGroup.tsx` — dropped the unreachable fallback that hard-coded `at 23:59 UTC` (Greptile flagged it on #2272).
- The remaining 8 files are test and story fixtures updated to the new string.

The hero CTA ("Apply by 27 Sep") is deliberately untouched — under AoE a bare date is true in every timezone.

## Heads-up: this reverses #2272

#2272 moved this from AoE to UTC day boundaries; the code it deleted was `now.getTime() - 12 * ONE_HOUR_MS`, which is what this restores. Neither that commit nor its PR body gives a rationale, and Airtable has since been changed to AoE, so the two systems currently disagree. @somepersonnamedjosh — flagging in case there was a reason for the UTC switch that isn't written down.

## Blast radius

`openRoundDeadlineCondition()` has three consumers, so this is wider than the course pages:

- `course-rounds` — landers, `/courses`, homepage
- `facilitator-applications` — including the check that a round is still open **at submission time** (a write path)
- `certificates`, via `hasUpcomingRoundsForCourseId` — gates the "apply to the next round" CTA

Facilitator applicants therefore also gain the 12 hours. That's consistent with the Airtable change (same Round table), but calling it out explicitly.

## Testing

```
npx vitest run src/server/routers/course-rounds.test.ts src/components/lander src/__tests__/pages/courses   # 93 passed
npx vitest run src/server/routers/facilitator-applications.test.ts src/server/routers/certificates.test.ts  # 68 passed
npm run typecheck   # clean
npm run lint        # clean
```

Threshold tests now assert a round with deadline `2026-04-12` is visible at `2026-04-13T11:59Z` and hidden at `2026-04-13T12:00Z`.

This is a copy-and-logic change with no layout impact — the only visual difference is the schedule-list subtitle reading "Applications close 27 Sep at 23:59 anywhere on earth". No screenshots attached; a human may want to eyeball the line length on mobile, since the string is ~14 characters longer than before.

## Note

Course pages use `revalidate: 5 min`, so rollover lands up to 5 minutes after 12:00 UTC.


eleniKougiou:
@claude review

eleniKougiou:
Thanks all three — no change requests, so nothing to push. Two notes for the record.

**Correcting a detail both @claude and @coderabbitai stated.** Both describe `buildDeadlineSubtitle` as now falling back to `'Applications open'` "when there's no detailed deadline." That branch is actually unreachable, and was before this PR too: the router sets `applicationDeadlineDetailed` to the string `'TBD'` when a round has no deadline (`course-rounds.ts:132-134`, `234-236`), and `'TBD'` is truthy. So a deadline-less round renders **"Applications close TBD"**, not "Applications open".

This doesn't change the verdict on the removal — the line I deleted was dead for the same reason, which is why Greptile flagged it on #2272. But I'd rather the next person reading this thread didn't inherit the wrong mental model. I've deliberately left the second dead branch alone to keep this PR to one behavioural change; worth a separate look at whether "Applications close TBD" is the copy we actually want there.

**Still open: the #2272 question.** @somepersonnamedjosh — @claude independently landed on this as the thing to resolve before merge. If the UTC boundary in #2272 was a deliberate call rather than a tidy-up, it'd be good to know before this flips back, since Airtable has already moved to AoE and the two currently disagree.

Verification on my side, for the record: `course-rounds`, `lander`, `pages/courses` (93 tests), plus `facilitator-applications` and `certificates` (68 tests) — the two suites that depend on `openRoundDeadlineCondition()` and that my first scoped run missed. Typecheck and lint clean locally, and CI agrees.

eleniKougiou:
pretty confident AoE timezone is what we want, so moving with the merge

PR #3051, opened 2026-09-18
[feat] Configure the team portal Ashby server secret

dewierwan:
Saving an Ashby key in GitHub does not make it available to the team portal at runtime. This connects the repository's `ASHBY_API_KEY` Actions secret to the infrastructure deployment, stores it as a Pulumi-managed Kubernetes Secret, and exposes it only to the `bluedot-team-apps` server container.

The repository secret has been added and its presence verified through GitHub's metadata API. The value is never committed, passed to the app build, or exposed as a public environment variable. Local infrastructure runs can use a private `ASHBY_API_KEY` environment variable or the encrypted Pulumi `ashbyApiKey` setting. Missing both fails configuration.

This provisions the credential for talent sourcing. Hosting the sourcing app in the authenticated portal, including durable data and background jobs, remains a separate change. Rotation and deployment instructions are in the portal README.

Validation:

- Infrastructure ESLint and TypeScript checks passed.
- Pulumi mocks verified the GitHub environment input, encrypted-config fallback, secret marking, Kubernetes reference, and missing-key failure.
- Workflow/service checks verified that only the infra deployment receives the GitHub credential and only the portal container receives its runtime reference.
- `git diff --check` passed. The infrastructure package has no existing test suite; no production Pulumi preview or deployment was run.


dewierwan:
@claude review

dewierwan:
Thanks for the review. Added a link from the infrastructure guide to the portal's Ashby provisioning and rotation instructions, explicitly noting that the GitHub Actions secret takes precedence over local Pulumi config. The existing comment beside the conditional in `secrets.ts` already explains that GitHub supplies the production key and Pulumi config supports local previews.

Kept the single conditional rather than introducing an override abstraction for one key. No runtime changes were needed. GitHub CI passed typecheck, lint, and tests on the implementation commit; local Pulumi mock checks also passed for the environment input, config fallback, and missing-key failure.

